PR #2, opened 2020-04-18
Improve typesafety and use yarn workspaces

hammadj:
Improves type safety by enabling strict mode and enables yarn workspaces to make managing monorepo easier


packages/app-react/@types/web3-provider.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@walletconnect/web3-provider';
```
pakokrew, line 1: This is strange, as the package repository seems to have the file but not the npm package :/

hammadj: Yea they don’t have proper types in there either, so it’s not really typesafe either way, this is just to silence the warning in strict mode

PR #3, opened 2020-04-18
Fix GraphQL codegen

hammadj:
The problem was that the introspection query didnt have the admin-secret in the auth header, so it wasn't able to get the mutation types.

Currently this is still as a separate package, but we should update the app-react queries to use the generated types now instead.


pakokrew:
Well done.
Those generated files will be used by app-react indeed, but I think better that it stays in a different package as other packages may be also requiring them.

hammadj:
@pakokrew yea normally it generates types based on the query defined in the component, but the fragment first approach with codegen-Hasura might make it more reusable.

PR #10, opened 2020-05-10
[WIP] Codegen

pakokrew:
Trying to update codgen but there's still a problem when generating code from tables that doesn't have an id field


pakokrew:
Filed an issue https://github.com/ahrnee/graphql-codegen-hasura/issues/13

krzysu:
I believe we can close this one

PR #15, opened 2020-07-06
Feature/deployment

hammadj:
Need to merge #14 first since this builds off that branch

Setups up config to deploy all our infrastructure to render.com


hasura/Dockerfile
```diff
@@ -5,21 +5,24 @@ RUN chmod +x /wait
...
 ## Default setup
 
+ARG BACKEND_HOST
+ARG AUTH_HOOK_PATH=auth-webhook
+ARG ACTION_BASE_PATH=actions
+
+ENV HASURA_GRAPHQL_DEV_MODE false
 ENV HASURA_GRAPHQL_ENABLE_TELEMETRY false
 ENV HASURA_GRAPHQL_ENABLED_LOG_TYPES startup, http-log, webhook-log, websocket-log, query-log
+ENV HASURA_GRAPHQL_AUTH_HOOK http://$BACKEND_HOST/$AUTH_HOOK_PATH
```
hammadj, line 15: @pakokrew The reason we have to do it like this is Render not allowing for string interpolation when defining ENV vars based on other services:
image

packages/app-react/src/config.ts
```diff
@@ -1,4 +1,4 @@
 export default {
-  graphqlURL: process.env.GRAPHQL_URL || 'http://localhost:8080/v1/graphql',
-  infuraId: process.env.INFURA_ID || '781d8466252d47508e177b8637b1c2fd',
+  graphqlURL: process.env.REACT_APP_GRAPHQL_URL || 'http://localhost:8080/v1/graphql',
```
hammadj, line 2: CRA strips out ENV vars that dont start with REACT_APP

render.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+databases:
+  - name: metagame-db
+    plan: starter
+
+envVarGroups:
+  - name: hasura-secrets
+    envVars:
+      - key: HASURA_GRAPHQL_ADMIN_SECRET
+        generateValue: true
```
hammadj, line 66: Randomly generated secret

Issue #25, opened 2020-07-12
Update Database Schema for new profiles setup


hammadj:
Simplify the DB schema to only the essentials we need for profiles

pakokrew:
Proposed modifications in schema :

**Player**
- id
- username
- totalXp
- rank (remove, calculated from totalXP)
- links (remove, get it from linked accounts)
- sentences (remove)
- guilds_member > (?)
- quests_completed > (?)
- XPIntervals > (?)

From the wireframes, we should add:
- role
- timezone
- enneagram profile
- skills

I didn't add availability since I don't think we'll be able to ask to hire right now.
Name, profile picture and about me are inside 3Box

**Accounts**
- player_id
- identifier
- linkToProof (remove ?)
- type


Do we keep Guilds, Quests, XPInterval ?

hammadj:
@pakokrew Link to proof comes from 3box as well, they have github/twitter in there, but yea dont need to have it in our DB, can just use remote join.

For guilds, I think we should still h

PR #33, opened 2020-07-13
Update models

pakokrew:
Updating the models by removing properties we don't use anymore and added some missing ones for #25 

**Player**

Removed
- rank (calculated from totalXP)
- links (we get it from linked accounts)
- sentences 

Added

- role (string)
- timezone (int, e.g -6, 0, +1, better idea ?)
- enneagram (enum)
- skills (string[])

FYI, [here's how to mutate string array in hasura](https://hasura.io/docs/1.0/graphql/manual/mutations/insert.html#insert-an-object-with-an-array-field)
Name, profile picture and about me are inside 3Box

**Account**

Removed linkToProof (fetchable from 3box)

**Guild**

Added 
- type (enum, ARAGON, MOLOCH...)
- identifier (text, dao address for instance)
- name
- logo (nullable)

Guild_member are fetched from remote schema

**Quest, XPInterval, Quest_completed, Guild_member**
Scrapped


hasura/migrations/1575910000000_init/up.sql
```diff
@@ -8,13 +8,22 @@ CREATE TYPE account_type AS ENUM (
...
+  'ACHIEVER',
+  'INDIVIDUALIST',
+  'INVESTIGATOR',
+  'LOYALIST',
+  'ENTHUSIAST',
+  'CHALLENGER',
+  'PEACEMAKER'
+);
+
+CREATE TYPE guild_type AS ENUM (
```
hammadj, line 23: For the guild types, since we will likely be changing this and adding more, using the PG native enum type has some drawbacks, would be good to set it up as foreign-key references: https://hasura.io/docs/1.0/graphql/manual/schema/enums.html#enums-in-a-database

hasura/migrations/1575910000000_init/up.sql
```diff
@@ -23,66 +32,34 @@ CREATE TABLE "Player" (
   "id" uuid DEFAULT public.gen_random_uuid() NOT NULL,
   "username" text NOT NULL,
   "totalXp" numeric DEFAULT 0,
-  "rank" player_rank NOT NULL DEFAULT 'PLAYER',
-  "links" json,
-  "sentences" json
+  "role" text,
+  "timezone" int,
+  "enneagram" enneagram_type,
+  "skills" text[]
```
hammadj, line 38: I think we should have a table of skills: https://hasura.io/docs/1.0/graphql/manual/schema/enums.html#using-foreign-key-references-to-a-single-column-table

That way we can allow people to search/filter based on the skills of a user and have autocomplete for a standard list of skills but also allow users to define new ones and insert them into the Skills table

pakokrew: Having the skills in a table would indeed allow to search/filter more easily but letting users insert new ones can lead to fake doubles, like mispelling, or calling the same thing with different names or punctuation. 
We should then have an autocomplete option like with algolia, and at some point we we'll have to aggregate same skills with different names.

hammadj: Yea, we can prepopulate with a standard list of skills to start and them add custom skills feature later to keep it simple

pakokrew: It seems that hasura's enums are not the good solution for skills for different reasons:
- PostgreSQL can't have columns with an array of reference to another table, which would mean people would only have one skill (https://stackoverflow.com/questions/41054507/postgresql-array-of-elements-that-each-are-a-foreign-key)
- In hasura, the elements of the enums are generated in the metadata, which means we have to manually regenerate it everytime the list of skills is updated (https://hasura.io/docs/1.0/graphql/manual/schema/enums.html#current-limitations)

The only solution I see here is to have a Skill table with a many-to-many relationships between Player-Skill

pakokrew:
@hammadj 
I have switch the guild types to the hasura enums format with foreign keys. As mentioned in their docs, we have to regenerate metadata and redeploy once new guild types are added, but that will not be much frequent. 

For the skills, as they will change more frequently, I have created a many-to-many relationship, where people can add new skills themselves, and we still are able to query/filters on them.

hasura/migrations/1575910000000_init/up.sql
```diff
@@ -8,81 +8,57 @@ CREATE TYPE account_type AS ENUM (
...
+CREATE TYPE enneagram_type AS ENUM (
+  'REFORMER',
+  'HELPER',
+  'ACHIEVER',
+  'INDIVIDUALIST',
+  'INVESTIGATOR',
+  'LOYALIST',
+  'ENTHUSIAST',
+  'CHALLENGER',
+  'PEACEMAKER'
```
hammadj, line 20: Now that we know the enum types arent great to use, should we make this a table too with many to one relationship? That way we can have a description and other metadata for each type?

pakokrew: Not sure that's a good idea, this will never change and therefore we don't really need to put these metadata it in our database

hammadj: no worries. Feel free to merge after resolving conflicts

Issue #26, opened 2020-07-12
Setup Query Action + Remote Join for 3Box Data


hammadj:
We want to be able to query the 3box API for a users profile information. Should set up a [query action](https://hasura.io/docs/1.0/graphql/manual/actions/create.html#query-type-action) that uses 3Box JS to load this data

hammadj:
done

PR #34, opened 2020-07-13
getBoxProfile

pakokrew:
Added a query action to get box profile for #26 

- Added `ethereum_address` field to Player model
- Create a graphql server in the backend that expose a first query, getBoxProfile
- Setup a remote schema in hasura, with Player.ethereum_address => getBoxProfile(address)
- Updated auth webhook/frontend for new model
- Frontend now fetches 3Box profile from hasura instead of 3box


hammadj:
> Added a query action to get box profile for #26
> 
> I couldn't setup a relation with internal models, needs to check how is it possible to do that

Just tried to do it, and yea realized you can do a remote join in the query action (Adding our player data into the response of the getBoxProfile query), but not the other way around (put the getBoxProfile query in the response for our Player data). 

Seems like you need to setup a remote schema to join external data into our tables :/

pakokrew:
Okay, so I got this working by creating a remote schema into our backend and removed the action.

Because in the **Account** table, we have ethereum addresses but also other types of accounts, it was not ideal to fetch profiles from that (can't only have the remote schema for only rows with `type=ETHEREUM`)
So I added the `ethereum_address` in the **Player** model, that will serve as a central identifier for players, and will be much more easy to process for identification and auth than having to fetch the **Account** table with `type=ETHEREUM`

I kept the `type=ETHEREUM` into accounts, in case people would like to link additional addresses, be we can remove it, as you prefer @hammadj

packages/@types/express-graphql/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'express-graphql';
```
hammadj, line 1: Dont need this, express graphql has ts definitions included

packages/backend/src/config.ts
```diff
@@ -1,5 +1,34 @@
...
+  adminKey: parseEnv<string>(
+    process.env.HASURA_GRAPHQL_ADMIN_SECRET,
+    'string',
+    'metagame_secret',
+  ),
+  ipfsEndpoint: parseEnv<string>(
+    process.env.IPFS_ENDPOINT,
+    'string',
+    'https://ipfs.infura.io',
+  ),
```
hammadj, line 33: Good shit!

PR #36, opened 2020-07-14
setup design system as a new package

krzysu:
I added a new package where I'm using Chakra as a base for our Meta design system. I set up storybook there and added fonts for now. Let me know what do you think. 
I will continue with colors and custom components but I would like to merge this already.

image


yarn.lock
```diff
@@ -130,13 +130,36 @@
   dependencies:
     "@apollo/react-common" "^3.2.0-beta.1"
 
+"@babel/code-frame@7.5.5":
```
krzysu, line 133: I don't know why there are some many changes in yarn.lock here, but that is the result of running yarn in root folder of this repo on my machine.

package.json
```diff
@@ -20,8 +20,8 @@
     "hasura:migrate:apply": "npm run hasura migrate apply",
     "test": "lerna run test --parallel",
     "clean": "lerna clean",
-    "format": "prettier --ignore-path .gitignore --write \"{*,**/*}.{ts,tsx,js,json,yml,yaml,md}\"",
-    "lint": "eslint --ignore-path .gitignore \"./packages/**/*.{ts,tsx,js}\"",
+    "format": "prettier --ignore-path .gitignore --write \"{*,**/*}.{ts,tsx,js,jsx,json,yml,yaml,md}\"",
+    "lint": "eslint --ignore-path .gitignore \"./packages/**/*.{ts,tsx,js,jsx}\"",
```
krzysu, line 24: I don't want to write storybook stories in typescript, doesn't bring any value

packages/design-system/src/theme.js
```diff
@@ -0,0 +1,14 @@
+import { theme } from '@chakra-ui/core';
+
+export const metaTheme = {
```
hammadj, line 3: Source files should be in TS, even if storybooks arent

packages/design-system/stories/0-Text.stories.jsx
```diff
@@ -0,0 +1,49 @@
+import { Stack, Text } from '@chakra-ui/core';
+import React from 'react';
+
+// eslint-disable-next-line import/no-default-export
```
hammadj, line 4: Should add a ESLint override for all stories so you dont need to do this in every file. 

Can make a .eslintrc file in the design-system/stories folder with this: 
```json
{
    "rules": {
      "import/no-default-export": "off",
      "@typescript-eslint/no-explicit-any": "off"
    }
}
```

PR #40, opened 2020-07-26
Setup Design System with NextJS Web App

hammadj:
Also replaced the local fonts with Google fonts since they are more performant and easier to package. Created a custom
Link component to render the ChakraUI link inside the NextJS Link properly.

PR #42, opened 2020-07-26
DevOps / Tooling Improvements

hammadj:
Must merge #40 first since this builds on top of that

Cleans up a bunch of the config / package.json stuff and improves speed of CI


package.json
```diff
@@ -20,7 +20,7 @@
     "hasura:console": "npm run hasura console -- --no-browser",
     "hasura:migrate:init": "npm run hasura migrate create \"init\" -- --from-server",
     "hasura:migrate:apply": "npm run hasura migrate apply",
-    "test": "lerna run test --parallel",
+    "test": "lerna run test --parallel --",
```
pakokrew, line 23: what's this for ?

hammadj: So we can forward args to the test command in each package for CI / GH Actions (e.g. --ci and --coverage). Without it the command gets consumed by lerna instead of forwarding to the test command.

PR #45, opened 2020-07-29
add fonts to storybook

krzysu:
that we can also see correct fonts when working with storybook


hammadj:
@krzysu Looks good, feel free to merge after resolving merge conflict. Also, ignore the vercel deployment failure

PR #50, opened 2020-07-31
Metamaps base prototype

TheLoneRonin:
### Description

Relating to issue #5. This pull request provides the base working prototype for the `MetaMaps` web application.

Please note that the application *is standalone* since it connects to MetaMask and 3Box... 
Thus it does not need the backend. 

So you can just do the following:

```bash
cd packages/metamaps
npm install
npm run dev
```

To test and experiment with the project.

@peth-yursick and I actually discussed how the menu layouts should be visual only... However, I've kept them text based (temporarily, for now, still) due to the fact that we need a much-needed discussion on how the user interface will work.

For example, how are maps going to be shared? So that people can collaborate on each others maps?
How are permissions delegated?
Is there a centralized ETH address to store maps?


TheLoneRonin:
Just noticed the package name (in the CI tests) is called `@metafam/web` and should be changed to `@metafam/metamaps` instead. I'll change that in the morning.

TheLoneRonin:
Oh, there's linting issues too!
I'll make sure to fix that shortly in a couple hours.

packages/metamaps/.eslintrc.json
```diff
@@ -0,0 +1,15 @@
...
+  "overrides": [
+    {
+      "files": ["./pages/**/*.{ts,tsx}"],
+      "rules": {
+        // NextJS requires default exports for pages
+        "import/no-default-export": "off",
+        "import/prefer-default-export": "error"
+      }
+    }
+  ]
```
hammadj, line 14: This is already being set in the root ESLint now, no need to have it here :) 
```suggestion
```

packages/metamaps/.eslintrc.json
```diff
@@ -0,0 +1,15 @@
...
+}
```
hammadj, line 15: This is already being set in the root ESLint now, can delete this safely

hammadj:
@TheLoneRonin Re: storage of maps. How do you feel about putting it in our backend for now, and then once we figure out the right UX and data model etc, we can then transition to putting it in more decentralized storage.

Each map should probably have a unique ID and people should be able to load into a map from the URL if we put the ID into it

TheLoneRonin:
@hammadj, for sure! I'm going to work on an update later today.

Is there anything else you'd like me to keep in mind?

hammadj:
> @hammadj, for sure! I'm going to work on an update later today.
> 
> Is there anything else you'd like me to keep in mind?

Nothing off the top of my head!

TheLoneRonin:
Hey @hammadj,

So all those issues you were experiencing were due to updating to make sure it passed lint check... It's my fault for not double-checking functionality to make sure it worked properly considering I made a lot of changes to the code to make it lint compliant. I've updated and tested the project to make sure all features work properly. Apologies for that.

So a few things.

1. RecoilJS is a great idea. Atomic computations are always superior. There is one problem with RecoilJS. It does not have full-featured support asynchronous state. See: https://recoiljs.org/docs/guides/asynchronous-state-sync. I definitely think we should keep our eyes on RecoilJS... For the obvious benefits of Atomic computations.

Furthermore, all the errors you were experiencing were part of my negligence for not testing the code after updating it to pass lint. :) 

2. The purpose of the project is to store everything in 3Box. Thus, I'm not sure if putting stuff in your backend is a good idea.

3. I've added a "loading" screen just to make the project more fluid.

4. There are still several eslint "warnings" that need to be resolved... Once the functionality is approved. I'm more than happy to go through those issues and get rid of all the warnings.

Thanks for your patience!

TheLoneRonin:
@hammadj, great! Let me just update the lint issues and I'll be adding in some changes.

I'll download the extensions required to make sure it works on mobile and other platforms etc!
I did not factor in mobile/cross wallet compatibility.

I'll also need to do an update to this PR to update the `yarn.lock`. Plus get rid of linting issues. So you may need to approve it again.

Let's continue this convo after the 2nd PR is in.

TheLoneRonin:
@hammadj, it's ready to merge. There are still some lint warnings. But it's basically just `any` types... And that's basically unavoidable.

PR #51, opened 2020-08-01
login landing page + fixed fonts

dan13ram:


packages/web/package.json
```diff
@@ -15,8 +15,10 @@
     "isomorphic-unfetch": "^3.0.0",
     "next": "latest",
     "next-google-fonts": "^1.2.1",
+    "next-images": "^1.4.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-google-font-loader": "^1.1.0",
```
krzysu, line 21: is this package in use in `web`?

dan13ram: Nope. I had added it to make fonts work. Will remove it

krzysu:
I've just run your branch locally and I confirm that fonts are not correctly displayed. I see correct definitions in CSS though. I assume something with `next-google-fonts` but maybe @hammadj has a better idea.

dan13ram:
Thanks @krzysu . 
As you mentioned the CSS has correct definitions, but I'm not sure if `next-google-fonts` is the issue since I tried to declare the fonts using simple `<link>` but to no avail.
Also I'm facing another weird issue with typecheck. I'm not too experienced with typescript, can you guys help me out here?

```
pages/index.tsx(21,9): error TS2322: Type 'string' is not assignable to type '"symbol" | "object" | "data" | "small" | "map" | "filter" | "sub" | "a" | "abbr" | "address" | "area" | "article" | "aside" | "audio" | "b" | "base" | "bdi" | "bdo" | "big" | "blockquote" | ... 334 more ... | undefined'.
```

packages/web/images.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.png';
```
hammadj, line 1: Should just add this to next-env.d.ts

packages/web/components/Link.tsx
```diff
@@ -1,13 +1,18 @@
-import { Link } from '@metafam/ds';
+import { Link, LinkProps as DSLinkProps } from '@metafam/ds';
 import NextLink, { LinkProps } from 'next/link';
 import React from 'react';
 
-type Props = LinkProps;
+type MetaLinkProps = LinkProps & DSLinkProps;
```
hammadj, line 5: The reason you are having type errors is that DSLinkProps is overwriting the "as" prop in LinkProps to something else. There's other conflicting types too in those objects, and if you just join them like this it thinks you want to accept either of these props. 

Instead, we should only use the DSLinkProps that aren't already defined in LinkProps. We can do that like this:
```suggestion
type MetaLinkProps = Omit<DSLinkProps, keyof LinkProps> & LinkProps;
```

This "omits" any of the props in DSLinkProps that match the name of any props in LinkProps

hammadj:
> Thanks @krzysu .
> As you mentioned the CSS has correct definitions, but I'm not sure if `next-google-fonts` is the issue since I tried to declare the fonts using a simple `<link>` tag but to no avail.
> Also I'm facing another weird issue with typecheck. I'm not too experienced with typescript, can you guys help me out here?
> 
> ```
> pages/index.tsx(21,9): error TS2322: Type 'string' is not assignable to type '"symbol" | "object" | "data" | "small" | "map" | "filter" | "sub" | "a" | "abbr" | "address" | "area" | "article" | "aside" | "audio" | "b" | "base" | "bdi" | "bdo" | "big" | "blockquote" | ... 334 more ... | undefined'.
> ```

Fixed the fonts issue in this PR: https://github.com/MetaFam/TheGame/pull/52

Also commented about the type error

packages/web/components/Link.tsx
```diff
@@ -1,13 +1,18 @@
...
 
-export const MetaLink: React.FC<Props> = ({ children, ...props }) => (
-  <NextLink {...props} passHref>
+export const MetaLink: React.FC<MetaLinkProps> = ({
+  children,
+  href,
+  as,
+  ...props
+}) => (
+  <NextLink href={href} as={as} passHref>
```
hammadj, line 13: Should make sure we pass in any of the other Next link props to it as well: https://nextjs.org/docs/api-reference/next/link

packages/web/components/Link.tsx
dan13ram: Thanks you!

dan13ram:
@hammadj Thanks for your feedback :) I've made the changes you've requested.

hammadj:
@dan13ram Seems like there's some merge conflicts, if you address those then I can merge it in.

dan13ram:
@hammadj I don't see any merge conflicts, and I've taken the latest pull of master branch

hammadj:
@dan13ram See here:
image


Can you try rebasing instead of pulling the master branch?

PR #55, opened 2020-08-04
define a few meta components in storybook

krzysu:
I'm trying to use Chakra components to style/theme them to fit MyMeta designs.

image
image
image
image


packages/design-system/src/MetaBox.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+
+export type MetaBoxProps = {
+  title: string;
+  children: React.ReactNode;
+};
+
+export const MetaBox: React.FC<MetaBoxProps> = ({ children, title }) => (
+  <Box>
+    <Box bg="#392373" p={4}>
```
hammadj, line 11: Can we define these values in the theme and refer to them by name instead of hardcoding them?

packages/design-system/src/theme.ts
```diff
@@ -4,10 +4,34 @@ export const theme: typeof chakraTheme = {
...
+    xs: '12px',
+    sm: '14px',
+    md: '16px',
+    lg: '18px',
+    xl: '20px',
+    '2xl': '24px',
+    '3xl': '28px',
+    '4xl': '36px',
+    '5xl': '48px',
+    '6xl': '64px',
```
hammadj, line 35: Are these the chakra defaults? Should we use rem instead of px?

packages/design-system/stories/0-Text.stories.tsx
```diff
@@ -1,44 +1,125 @@
...
+          color="#79F8FB"
+          fontWeight="normal"
+          verticalAlign="inherit"
+          onClick={() => {}}
+        >
+          Read more
+        </Button>
+      </Text>
+
+      <Link fontSize="sm" color="#79F8FB" href="https://metagame.wtf">
```
hammadj, line 78: Should always use values from theme instead of hardcoding

packages/design-system/src/theme.ts
krzysu: yes, these are chakra defaults, I pasted it here for a quick reference, I'll remove it for now and stay with defaults

packages/design-system/src/MetaBox.tsx
krzysu: sure, thanks for the feedback

packages/design-system/src/theme.ts
hammadj: Happy to keep them here actually, even if they are defaults. Makes it easy to know what we are using

packages/design-system/src/theme.ts
```diff
@@ -1,13 +1,47 @@
...
+      400: '#5a32e6',
+      500: '#4d19cd',
+      600: '#4412a0',
+      700: '#380d74',
+      800: '#260747',
+      900: '#11021d',
+    },
+    dark: '#1B0D2A',
+    purpleBoxDark: '#261943',
+    purpleBoxLight: '#392373',
```
hammadj, line 37: IMO these should be using values from the purple color palette. e.g. purpleBoxDark should be defined as purple.800. Or, we should just update the purple palette with these colors and use it directly.

packages/design-system/src/theme.ts
```diff
@@ -1,13 +1,47 @@
...
+    purpleTag: '#40347C',
+    blueLight: '#A5B9F6',
+    cyanText: '#79F8FB',
```
hammadj, line 40: Similarly with these, we should use colors out of the color palette. If the color palette doesn't have the colors we want, we should add them there. In fact, we can just generate the whole purple swatch based on the shade of purple we have by using something like this: https://smart-swatch.netlify.app/

packages/design-system/stories/0-Text.stories.tsx
```diff
@@ -1,44 +1,125 @@
...
-    <Text fontSize="lg" fontFamily="body" fontWeight="bold">
+    <Text fontSize="lg" fontFamily="body" fontWeight="bold" color="white">
       In love with MetaGame - Body Bold
     </Text>
-    <Text fontSize="lg" fontFamily="mono">
+    <Text fontSize="lg" fontFamily="mono" color="white">
       In love with MetaGame - Mono
     </Text>
-    <Text fontSize="lg" fontFamily="mono" fontWeight="bold">
+    <Text fontSize="lg" fontFamily="mono" fontWeight="bold" color="white">
```
hammadj, line 122: We shouldn't need to set these all as white if we just change the theme to Dark Mode. Can we wrap storybooks in the DarkMode theme? Feel free to do it in a separate PR and clean these up later if you prefer.

krzysu:
I will wait for #63 to be merged first

hammadj:
@krzysu #63 is merged, this needs a rebase to fix merge conflicts then its G2G

PR #63, opened 2020-08-07
Layouts for RegisterPlayer and SuccessPlayer 

dan13ram:
- registerPlayer & successPlayer layout done
- moved to chakra-ui@next
- added meta components and icons
- refactored code based on review comments


dan13ram:
@krzysu I've created meta components MetaButton and MetaHeading. 
They are slightly different from the ones on the other PR you've raised, I have only refactored my code to reuse components used in multiple places. Maybe we could modify these to be more generic.

dan13ram:
@hammadj I've incorporated the comments you had given in the other PR. 
Instead of using a darkMode, I've used global styles provided by chakra-ui@next. 
So I went ahead and removed the <Global /> component provided by `@emotion/core`, since it is no longer required.

packages/design-system/stories/1-Button.stories.tsx
```diff
@@ -8,58 +9,62 @@ export default {
...
+    <Button colorScheme="teal" variant="link">
       Button
     </Button>
   </ButtonGroup>
 );
 
 export const Icons = () => (
   <ButtonGroup spacing={4}>
-    <Button leftIcon="email" variantColor="teal" variant="solid">
+    <Button leftIcon={<EmailIcon />} colorScheme="teal" variant="solid">
```
dan13ram, line 46: This is the new way we have to use icons in chakra-ui@next

packages/web/components/RegisterPlayer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { FlexContainer } from './Container';
+
+export const RegisterPlayer: React.FC = () => {
+  return (
+    <FlexContainer flex={1}>
+      <MetaHeading m={5}>Register your Player account</MetaHeading>
+      <ButtonGroup spacing={5} mt={20}>
+        <RegisterButton textTransform="uppercase">
+          <Icon3box mr={2} /> box
```
dan13ram, line 18: As mentioned above, they have changed how icons work in chakra-ui@next. 
And when I use them as `leftIcon`, the padding they had set doesn't work and the icons are not center aligned horizontally. This can be seen by running story-book where icons are used as `leftIcon` and `rightIcon`.
Ref: https://next.chakra-ui.com/docs/migration#button & https://github.com/chakra-ui/chakra-ui/issues/1024
Hence, I've used the icons as children with margin set on them.

packages/web/pages/_document.tsx
```diff
@@ -24,10 +24,6 @@ class MetaDocument extends Document {
             rel="stylesheet"
           />
           <meta charSet="UTF-8" />
-          <meta
-            name="viewport"
-            content="width=device-width, initial-scale=1.0"
-          />
```
dan13ram, line 30: This removes the following warning:
Warning: viewport meta tags should not be used in _document.js's <Head>. 
https://err.sh/next.js/no-document-viewport-meta

packages/web/components/SuccessPlayer.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const SuccessPlayer: React.FC = () => {
+  return (
+    <FlexContainer h="100%">
+      <MetaHeading m={5}>Success!</MetaHeading>
+      <Image m={10} src={AvatarImage} />
+      <MetaButton mt={5} mb={8}>
+        Set up your profile
+      </MetaButton>
+      <Text fontFamily="heading" color="offwhite">
```
dan13ram, line 16: I'm using `<Text />` here instead of `<Heading />` since the default fontSize of `<Heading />` is too large, and I think this is cleaner than reducing the fontSize of `<Heading />`.

dan13ram: @krzysu @hammadj Do you think we should switch the names of fontFamily of `heading` and `mono`? 
Currently its like so
```
fonts: {
    body: '"IBM Plex Sans", sans-serif',
    heading: '"IBM Plex Mono", monospace',
    mono: '"Press Start 2P"',
  }
```
I suggest we change it to
```
fonts: {
    body: '"IBM Plex Sans", sans-serif',
    mono: '"IBM Plex Mono", monospace',
    heading: '"Press Start 2P"',
  }
```

hammadj: @dan13ram You can customize font size with the Heading component: https://next.chakra-ui.com/docs/components/heading#changing-font-size

Also, I think we want to keep Press Start 2P as the heading font

packages/web/components/RegisterPlayer.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <RegisterButton>
+          <EmailIcon mr={2} />
+          Email
+        </RegisterButton>
+      </ButtonGroup>
+    </FlexContainer>
+  );
+};
+
+const RegisterButton: React.FC<typeof Button> = ({ children, ...props }) => (
```
hammadj, line 30: To get the correct prop types, you will have to do:
```suggestion
const RegisterButton: React.FC<React.ComponentProps<typeof Button>> = ({ children, ...props }) => (
```

But, easier way to do it is to import/export the `ButtonProps` type from chakra

packages/web/next-env.d.ts
```diff
@@ -1,3 +1,4 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 declare module '*.png';
+declare module '@metafam/ds';
```
hammadj, line 4: Shouldnt do this, we will lose type-safety. What was the reason you added this?

packages/web/pages/login.tsx
```diff
@@ -1,62 +1,43 @@
...
-        fontSize="3xl"
-        fontFamily="mono"
-        fontWeight="bold"
-        color="#FFFFFF"
-        m={5}
+const Login: React.FC = () => {
+  const [step, setStep] = useState(0);
+  return (
+    <PageContainer backgroundImage={`url(${BackgroundImage})`}>
+      <SimpleGrid
```
hammadj, line 15: Nice, much cleaner!

packages/web/components/SuccessPlayer.tsx
krzysu: I also find it confusing to have 'mono' set to Press Start and 'heading' set to IBM Plex Mono so I'm for the suggested change, one more thing, let's set fallback for Press Start to sans-serif too, so

```
fonts: {
    body: '"IBM Plex Sans", sans-serif',
    mono: '"IBM Plex Mono", monospace',
    heading: '"Press Start 2P", sans-serif',
  }
```

packages/web/pages/login.tsx
```diff
@@ -1,62 +1,43 @@
 import { Flex, Image, SimpleGrid, Text } from '@metafam/ds';
-import React from 'react';
+import React, { useState } from 'react';
 
+import { PageContainer } from '../components/Container';
 import { MetaLink } from '../components/Link';
-import BackgroundImage from '../public/images/background.png';
+import { RegisterPlayer } from '../components/RegisterPlayer';
+import { WelcomePlayer } from '../components/WelcomePlayer';
+import BackgroundImage from '../public/images/login-background.png';
```
krzysu, line 8: for the future, `jpg` is better optimized for large background images than png, we could also think about webP

packages/web/next-env.d.ts
dan13ram: @hammadj I'm actually getting type errors when I run `yarn typecheck` without this line:
```
components/RegisterPlayer.tsx(14,20): error TS2322: Type '{ children: Element[]; flex: number; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
  Property 'flex' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
components/RegisterPlayer.tsx(17,25): error TS2322: Type '{ children: (string | Element)[]; textTransform: string; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"button", ButtonProps> & { children?: ReactNode; }'.
  Property 'textTransform' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"button", ButtonProps> & { children?: ReactNode; }'.
components/RegisterPlayer.tsx(20,10): error TS2322: Type '{ children: string; }' is not assignable to type 'ComponentWithAs<"button", ButtonProps>'.
  Type '{ children: string; }' provides no match for the signature '<TT extends ElementType<any>>(props: PropsWithAs<TT, ButtonProps>): ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | null'.
components/RegisterPlayer.tsx(21,10): error TS2322: Type '{ children: (string | Element)[]; }' is not assignable to type 'ComponentWithAs<"button", ButtonProps>'.
  Type '{ children: (string | Element)[]; }' provides no match for the signature '<TT extends ElementType<any>>(props: PropsWithAs<TT, ButtonProps>): ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<...>)> | null'.
components/SuccessPlayer.tsx(10,20): error TS2322: Type '{ children: Element[]; h: string; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
  Property 'h' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
components/WelcomePlayer.tsx(13,20): error TS2322: Type '{ children: Element[]; flex: number; maxWidth: string; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
  Property 'flex' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
pages/login.tsx(14,20): error TS2322: Type '{ children: (false | Element)[]; backgroundImage: string; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
  Property 'backgroundImage' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
pages/profile.tsx(9,20): error TS2322: Type '{ children: Element; backgroundImage: string; }' is not assignable to type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
  Property 'backgroundImage' does not exist on type 'IntrinsicAttributes & ComponentWithAs<"div", FlexProps> & { children?: ReactNode; }'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/web/pages/login.tsx
dan13ram: Sure @krzysu  I did not know this.. I can update these images with jpg (exported as jpg from figma) in the future 
=> I tried converting the jpeg to webp, and the image size increase! So I think jpeg is the better option here.
Example sizes: 
background.png 842.9KiB
background.jpg 367.9KiB
background.webp 478.9KiB

packages/web/components/SuccessPlayer.tsx
dan13ram: Great. So I will change `Press Start 2P` to be the heading font.

packages/web/components/RegisterPlayer.tsx
dan13ram: In `chakra-ui@next` they are not exporting `ButtonProps` or `LinkProps`, and I get the following error when using them:
```
components/RegisterPlayer.tsx(31,32): error TS2709: Cannot use namespace 'ButtonProps' as a type.
```

packages/web/next-env.d.ts
dan13ram: Actually the issue was that I was using `typeof Button` instead of `React.ComponentProps<typeof Button>`. 
That fixed this issue!

packages/web/components/RegisterPlayer.tsx
dan13ram: So I'm using `React.ComponentProps<typeof Button>` itself.

dan13ram:
@hammadj @krzysu I've incorporated all the requested changes. Please check and merge :)

PR #118, opened 2020-09-17
connect wallet instead of login button

krzysu:
I think we need to rethink the new/registered user flow, for now, I improved connecting of web3 wallet


packages/web/contexts/Web3Context.tsx
```diff
@@ -40,36 +46,26 @@ export const Web3ContextProvider: React.FC = ({ children }) => {
...
-      setEthersProvider(provider);
+      setAddress(await ethersProvider.getSigner().getAddress());
+      setProvider(ethersProvider);
+      setIsConnected(true);
     }
   }, [web3Modal]);
 
   const disconnect = useCallback(async () => {
     web3Modal?.clearCachedProvider();
+    setIsConnected(false);
```
hammadj, line 62: Should we also setAddress and provider to null here?

hammadj (approved):
looks good!

PR #119, opened 2020-09-18
restructure and reorganize components

krzysu:
I preparing to build correct new user flow (for registered and existing users) and as a first step, I restructured components to make it easier to move them around. I renamed `/login` to `/welcome` and added a separate `/register` page (before it was just a different state of `/login` page). But honestly I find it hard to figure out the correct user flow and connect all hanging pages together.


dan13ram (approved):
LGTM. I think we will understand the flow better as we start integrating the backend more.

PR #145, opened 2020-10-11
Implement User Auth on Web

hammadj:
Will allow for making mutations and updates to profile, etc

### DIFF
--- a/packages/app-react/src/contexts/Web3.tsx
+++ b/packages/app-react/src/contexts/Web3.tsx
@@ -1,8 +1,7 @@
 import { useApolloClient } from '@apollo/react-hooks';
 import { did } from '@metafam/utils';
 import WalletConnectProvider from '@walletconnect/web3-provider';
-import { ethers } from 'ethers';
-import { AsyncSendable, Web3Provider } from 'ethers/providers';
+import { providers } from 'ethers';
 import React, { createContext, useCallback, useEffect, useState } from 'react';
 import Web3 from 'web3';
 import Web3Modal from 'web3modal';
@@ -11,7 +10,7 @@ import { getTokenFromStore, login, loginLoading, logout } from '../apollo/auth';
 import { CONFIG } from '../config';
 
 type Web3ContextType = {
-  ethersProvider: Web3Provider | null;
+  ethersProvider: providers.Web3Provider | null;
   connectWeb3: () => Promise<void>;
   disconnect: () => void;
 };
@@ -40,12 +39,13 @@ const web3Modal = new Web3Modal({
 export const Web3ContextProvider: React.FC = ({ children }) => {
   const apolloClient = useApolloClient();
 
-  const [ethersProvider, setEthersProvider] = useState<Web3Provider | null>(
-    null,
-  );
+  const [
+    ethersProvider,
+    setEthersProvider,
+  ] = useState<providers.Web3Provider | null>(null);
 
   const connectDID = useCallback(
-    async (provider: Web3Provider) => {
+    async (provider: providers.Web3Provider) => {
       let token = getTokenFromStore();
 
       if (!token) {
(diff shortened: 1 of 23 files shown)


packages/web/pages/_app.tsx
```diff
@@ -1,9 +1,13 @@
 import { ChakraProvider, CSSReset, MetaTheme } from '@metafam/ds';
 import { PageHeader } from 'components/PageHeader';
+import { CONFIG } from 'config';
 import { Web3ContextProvider } from 'contexts/Web3Context';
+import { withUrqlClient } from 'next-urql';
 import { AppProps } from 'next/app';
 import Head from 'next/head';
 
+import { getTokenFromStore } from '../lib/auth';
```
dan13ram, line 9: can we ensure that in future commits we remove any relative path imports?
imo makes it cleaner.

packages/web/lib/hooks.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
dan13ram, line 1: can we remove this and set correct types?

packages/web/lib/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useRouter } from 'next/router';
+import { useContext, useEffect } from 'react';
+
+import { Web3Context } from '../contexts/Web3Context';
```
dan13ram, line 5:
```suggestion
import { Web3Context, Web3ContextType } from '../contexts/Web3Context';
```

packages/web/lib/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useGetMeQuery } from '../graphql/autogen/types';
```
dan13ram, line 6:
```suggestion
import { useGetMeQuery, Me, Maybe, PlayerFragmentFragment } from '../graphql/autogen/types';
```

packages/web/lib/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const useWeb3 = () => useContext(Web3Context);
```
dan13ram, line 8:
```suggestion
export const useWeb3 = (): Web3ContextType => useContext(Web3Context);
```

packages/web/lib/hooks.ts
```diff
@@ -0,0 +1,39 @@
...
+
+type UseUserOpts = {
+  redirectTo?: string;
+  redirectIfFound?: boolean;
+};
+
+export const useUser = ({ redirectTo, redirectIfFound }: UseUserOpts = {}) => {
```
dan13ram, line 15:
```suggestion
export type User = Pick<Me, 'id' | 'ethereum_address' | 'username'> & {
   player?: Maybe<PlayerFragmentFragment>;
};

export const useUser = ({ redirectTo, redirectIfFound }: UseUserOpts = {}): User => {
```

dan13ram (approved):
make some suggestions, but looks great overall

packages/app-react/src/contexts/Web3.tsx
```diff
@@ -11,7 +10,7 @@ import { getTokenFromStore, login, loginLoading, logout } from '../apollo/auth';
 import { CONFIG } from '../config';
 
 type Web3ContextType = {
```
dan13ram, line 12:
```suggestion
export type Web3ContextType = {
```

packages/web/lib/hooks.ts
hammadj: Added ESLint rule instead of ignoring in file. I think for hooks its fine to not annotate the return type since its usually based on things that are well typed / external to our code (e.g. useContext return type handled by React, useGetMeQuery return type handled by graphql-codegen).

packages/web/lib/hooks.ts
hammadj: IMO this is just adding unnecesarry boilerplate and duplicating what we already have in the generated types. Makes the code harder to understand.

PR #167, opened 2020-10-23
profile setup flow clean up

dan13ram:
Screenshot from 2020-10-23 21:50:09
Screenshot from 2020-10-23 20:41:35
Screenshot from 2020-10-23 20:41:45


hammadj (approved):
Good stuff!

PR #190, opened 2020-11-18
Locked bcrypto version to 5.2.0 to avoid compilation errors on macos

alalonde:
JayWelsh and I both ran into issues initially building the repo on macos. 

This overrides the `5.1.0` version pulled in through our dependencies.

This may not be an ideal fix, but it's here at least for reference.


hammadj (approved):
approved, feel free to merge once CI passes :) Congrats on your first PR!

PR #194, opened 2020-11-24
Tooling improvements

alalonde:
Two changes for better developer ergonomics:

- Add `lerna` key to `lerna.json` to avoid [this error](https://github.com/lerna/lerna/issues/1067#issuecomment-357023774
- Add `get-graphql-schema` to codegen repo as the scripts depend on it


hammadj (approved):
Awesome, thanks!

PR #195, opened 2020-11-26
Backend Guide

TheLoneRonin:
### Description

- A new folder `guides` has been added to the root repository.

- A guide going through how to configure the database and insert data has been created

- Updated `README` pointing to `guides`.

Also, we can work on organizing the `README` more later.

### DIFF
--- a/guides/BACKEND.md
+++ b/guides/BACKEND.md
@@ -1,0 +1,276 @@
+## MetaGame's BackEnd API Guide
+
+### Requirements
+
+To run the MetaGame backend you will the following installed.
+
+- Node.js
+
+- Docker
+
+### Configuration
+
+Before you can start the Docker containers, you must run the following.
+
+```bash
+# Copy the example environment
+cp .env.sample .env
+
+# Add the BACKEND_HOST environment variable
+echo 'BACKEND_HOST=host.docker.internal:4000' >> .env
+
+# Remove potential stale containers
+yarn docker:clean
+```
+
+### Starting the BackEnd
+
+After the above is configured, you can start the Docker containers.
+
+```bash
+yarn start:docker
+```
+
+### Getting Started
+
+After running `yarn start:docker` make sure to confirm the backend is running.
+
+```bash
+# Confirm the backend works
+$ curl localhost:4000
+
+# This should output
+ok
+```
+
+You can debug further if there are problems by running `docker ps`.
+
+```bash
+$ docker ps
+
+# Check to see if there are any issues with the docker containers.
+CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                    NAMES
+6355ef5641a0        the-game_backend    "docker-entrypoint.s…"   51 minutes ago      Up 2 minutes        0.0.0.0:4000->4000/tcp   the-game_backend_1
+970f1ffb79c8        the-game_hasura     "/bin/sh -c '/wait &…"   51 minutes ago      Up 2 minutes        0.0.0.0:8080->8080/tcp   the-game_hasura_1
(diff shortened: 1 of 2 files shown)


.env.sample
```diff
@@ -10,3 +10,5 @@ NEXT_TELEMETRY_DISABLED=1
 HASURA_PORT=8080
 NEXT_PUBLIC_INFURA_ID=781d8466252d47508e177b8637b1c2fd
 NEXT_PUBLIC_GRAPHQL_URL=http://localhost:8080/v1/graphql
+
+BACKEND_HOST=host.docker.internal:4000
```
hammadj, line 14: Note that we only want to do this when running the backend outside of docker. Normally we just want to use the Docker version. I'm thinking we get rid of this since it's kinda hacky and make a simpler way to trigger that function in the backend. One good way would just be to check if the DB is empty, and if so call that function to populate it, that way there's no need for people to do it manually

hammadj (approved):
Looks good! Just one note about the backend host and initial data population steps, feel free to merge after reverting that env variable and you can make another PR for automatically populating it instead at startup if there's an empty DB

.env.sample
TheLoneRonin: I like this idea a lot!
You're right in regards to docker and how the ports will be different.

TheLoneRonin:
@hammadj, I've also improved the formatting a bit more on those changes.

We can improve the `docker-compose` setup in the next PR.

https://github.com/MetaFam/TheGame/pull/195/files?short_path=6e7fcc6#diff-6e7fcc626ed3596d3b0a098360ac789f2e6a26b33e54ea7aedd1877726d8bbc7

TheLoneRonin:
So there are a couple things I need to go over.

1. Permissions and confirming changing to `tables.yml`.

2. A frontend guide. Which essentially goes over how to connect queries with `urql`.

The frontend guide I will send in a new PR after I wrap up the new revamped Metamaps.
I will ping you once I add in the permissions section and reorganize the document a bit more.

guides/BACKEND.md
```diff
@@ -0,0 +1,226 @@
...
+            id: {_eq: "[UUID]"}
+        },
+        # The columns you want to update
+        _set: {
+            data: "..."
+        }) {
+        # you can either supply `returning` or `affected_rows` for the response
+        affected_rows
+    }
+}
```
hammadj, line 179: In the case where you're updating single items by their id / primary key, a better way to do it is use `update_Item_by_pk`.

guides/BACKEND.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+5. Furthermore, when selecting or updating data. You can add permissions for specific columns. You select which ones should be allowed via the provided checkboxes.
+
+6. Finally, make sure that the changes for permissions are updated in `hasura/metadata/tables.yaml`.
+`Pre-update check`nsert and read data immediately. There are pre-generated functions that come with Hasura. For creating new entries. The following are example queries you could send immediately to `http://localhost:8080/v1/graphql`.
+
+
+1. Insert a new item
```
hammadj, line 140: Should add a title for this section, right now it just goes from `6.` to `1.` . Should also link to Hasura's docs on making queries and mutations for people who want to dive deeper: 
https://hasura.io/docs/1.0/graphql/core/queries/index.html#queries
https://hasura.io/docs/1.0/graphql/core/mutations/index.html

hammadj (approved):
Few minor changes requested, otherwise looks good, feel free to merge after addressing !

TheLoneRonin:
@hammadj, great! I don't have merge permissions. But I have updated it.

PR #196, opened 2020-11-26
Metamaps Table

TheLoneRonin:
### Description

- Adds a new `Map` table to Hasura

- The initial data types are `name` and `data`. All metadata for the map is stored as a JSON blob in `data`

- Directly connected to the `Player` table

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -92,6 +92,9 @@
       filter: {}
 - table:
     schema: public
+    name: Map
+- table:
+    schema: public
     name: Player
   object_relationships:
   - name: EnneagramType
--- a/hasura/migrations/1606686200858_create_table_public_Map/down.sql
+++ b/hasura/migrations/1606686200858_create_table_public_Map/down.sql
@@ -1,0 +1,1 @@
+DROP TABLE "public"."Map";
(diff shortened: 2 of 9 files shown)


hasura/migrations/1606365023960_create_table_public_Map/up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+CREATE TABLE "public"."Map"("id" uuid NOT NULL DEFAULT gen_random_uuid(), "player_id" uuid NOT NULL, "name" text NOT NULL, "data" text NOT NULL, PRIMARY KEY ("id") , FOREIGN KEY ("player_id") REFERENCES "public"."Player"("id") ON UPDATE restrict ON DELETE restrict, UNIQUE ("id"));
```
hammadj, line 2: I would call it something like "owner_id" or "author_id" instead so it's clear who this player is in relation to the map

hammadj (changes requested):

hasura/migrations/1606365023960_create_table_public_Map/up.sql
TheLoneRonin: Agreed! Right now, I'm storing everything using `data` as a text blob.
I'm unsure if we want to integrate it as `MapItem`. I don't think there are any real benefits since everything is rendered interactively client side anyway.

In my opinion, if this is good as is. Let's change it to author_id. I will be merging the MetaMaps core codebase soon.

TheLoneRonin:
@hammadj, just updated that. I will open a new PR once I successfully migrate the new MetaMaps code in.

TheLoneRonin:
Hmmm for some weird reason it made redundant changes. One second.
Going to close this PR and make a new one sorry about that @hammadj!

Issue #191, opened 2020-11-19
Allow setting and reading of player timezones


alalonde:
- [x] Add IANA timezone to player table + migration
- [x] Add new screen to profile setup flow (after "availability")
- [x] Read from new column on player page

alalonde:
Another thing I'd like to add is a link to update your timezone. This could just jump to the appropriate step in the setup flow. However it'd require a bit of refactoring for the setup flow navigation to support inner links (which should probably happen anyway)

alalonde:
Fixed with #230 and #231

PR #203, opened 2020-12-03
Add timezone to setup flow and display it in your profile

alalonde:
Addresses #191 

Outstanding questions:

- I see that there's already an `int timezone` column in the database. The best practice is to store the IANA time zone name for best precision. I didn't remove it as I'm not sure if there are any ramifications.

Fixed along the way:

- Skills weren't being saved until you saved your availability
- Cleaned up graphql mutations


hammadj:
image
Looks like build is failing because schema file wasn't updated for the latest stuff in Hasura. Will need to run yarn update-schema first so it knows about the new tz field.

dan13ram (approved):
LGTM except for the build error. 
Cleaning my docker and restarting worked (didn't have to run `yarn update-schema`)

alalonde:
Yeah the schema was definitely updated; it seems there's some lingering cruft during the builds. Can we update the build to regenerate each time?

alalonde:
I would debug the build myself but I think I need to be authorized to even view it..

hammadj:
Ok so the issue with this build is that you're trying to update the backend and the frontend at the same time. Since the backend doesn't have these new fields, the front end build fails because its still pointing to the current version of the backend running.

To address this, you should split up the PR into one that updates the backend first and then another PR after that to implement the front end changes.

hammadj (commented):
@alalonde Ok so the issue with this build is that you're trying to update the backend and the frontend at the same time. Since the backend doesn't have these new fields, the front end build fails because its still pointing to the current version of the backend running.

To address this, you should split up the PR into one that updates the backend first and then another PR after that to implement the front end changes. It will be much more atomic and easier to review that way as well (plus more XP)

pakokrew:
> @alalonde Ok so the issue with this build is that you're trying to update the backend and the frontend at the same time. Since the backend doesn't have these new fields, the front end build fails because its still pointing to the current version of the backend running.
> 
> To address this, you should split up the PR into one that updates the backend first and then another PR after that to implement the front end changes. It will be much more atomic and easier to review that way as well (plus more XP)

Having a PR that update just the backend but not the frontend will make the frontend break once merged, no ?

PR #206, opened 2020-12-11
Update guide to include typescript compile

pakokrew:
I cloned the repo to a fresh install to verify that everything was working out of the box, and here's my findings:

- The `BACKEND_HOST=host.docker.internal:4000` line is only needed when one wants to start the backend without docker. Not needed in most cases.

- Before we can run either the frontend or the backend, every apps must be compiled with typescript with `yarn typecheck`.
If we don't do that, the backend will appear running with `docker ps` but will not be started as the typescript watcher found errors in the code and won't start the node server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,17 +121,3 @@ yarn docker:start
 ```shell script
 yarn typecheck
 ```
-
-### Run backend without docker for debugging
-
-If you want to run the NodeJS backend service out of docker to be able to debug with your IDE:
-
-**Add environment variable** to tell hasura where to find the backend (may only work on MacOS)
-```shell script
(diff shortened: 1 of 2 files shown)


dan13ram (approved):
LGTM. 
```
echo 'BACKEND_HOST=host.docker.internal:4000' >> .env
```
Maybe we can keep this with a comment telling us we can use it for backend dev perhaps?

pakokrew:
> ```
> echo 'BACKEND_HOST=host.docker.internal:4000' >> .env
> ```
> 
> Maybe we can keep this with a comment telling us we can use it for backend dev perhaps?

That was in the readme before. I'll put this back in the guide.

PR #213, opened 2020-12-14
Metamaps Alpha

TheLoneRonin:
### Description

A complete revamp of Metamaps that includes:

- Integration with Hasura and GraphQL

- Improved visual design and interfaces

- Improved contextual menu items and actions

- Customizable backgrounds

- New shapes and objects

- Integrations for images and videos

- Customizable colors for objects

- Assignable objects integrated with 3Box profiles

### Notes

This version is ready for people to use... But, might be a bit too early to consider as a replacement for existing tools.


TheLoneRonin:
Migrated to new PR

pakokrew (changes requested):
Looks interesting, but I couldn't start the app on my machine. How do you run it ?

The way you use Redux is very uncommon - that's not problematic for a first iteration, but it may lead to problems for maintenance and performance. 

The most urgent to me is to be sure that the store only contains literals - so being serializable. I see you store react component inside of it, that's a no-go.

I will try to make it easy to share a component between our apps so you don't have to handle wallet & auth yourself.

PR #214, opened 2020-12-14
nft gallery working

dan13ram:


hammadj:
Ok so the issue with this build is that you're trying to update the backend and the frontend at the same time. Since the backend doesn't have these new fields, the front end build fails because its still pointing to the current version of the backend running.

To address this, you should split up the PR into one that updates the backend first and then another PR after that to implement the front end changes.

hammadj (changes requested):
Ok so the issue with this build is that you're trying to update the backend and the frontend at the same time. Since the backend doesn't have these new fields, the front end build fails because its still pointing to the current version of the backend running.

To address this, you should split up the PR into one that updates the backend first and then another PR after that to implement the front end changes.

dan13ram:
Thanks @hammadj. will raise new PRs for the same

PR #216, opened 2020-12-14
Setup New Guilds Tables and PlayerGuild relationship

hammadj:


hasura/migrations/1607542601049_alter_table_public_Guild_add_column_join_button_url/down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."Guild" DROP COLUMN "join_button_url";
```
pakokrew, line 1: Could you squash these ADD/DROP columns migrations?

pakokrew (changes requested):

hasura/migrations/1607542601049_alter_table_public_Guild_add_column_join_button_url/down.sql
hammadj: So I recently was on a call with Hasura and they said they're starting to recommend people to not squash migrations since it can start to become problematic and you lose the ability to see where things went wrong. For these few it's not a problem, but yea in general I want to start avoiding squashing unless we are sure it's not going to cause problems like these new columns we are adding

pakokrew: fair enough

pakokrew: I do not see clearly where that can be problematic.
Here we do a pull request, that will affect the production only once deployed, and should be able to be rolled back 
Maybe they meant squashing migration that are already deployed ?

I find it very confusing like this, hard to review, without saying that we will endup with an enourmous list of migration over time

hasura/migrations/1607540645353_update_guild_tyles/up.sql
```diff
@@ -0,0 +1,9 @@
+DELETE FROM "Guild";
```
pakokrew, line 1: That one doesn't have down migration

hasura/migrations/1607539959503_alter_table_public_Guild_alter_column_type/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE ONLY "public"."Guild" ALTER COLUMN "type" SET DEFAULT 'The area of focus for the guild (e.g. funding, project, etc)';
```
pakokrew, line 1: looks like this one is dropped on a next one

pakokrew (changes requested):

pakokrew:
I added some relationships in hasura

pakokrew:
It is missing a lot of permissions in hasura, is it meant to be done on a later PR ?

dan13ram (approved):
LGTM, needs rebase but

PR #217, opened 2020-12-14
Add build workflow to CI so errors can be shown in GitHub instead of Vercel

hammadj:
Should help contributors debug their builds easier


dan13ram (approved):
lgtm

PR #218, opened 2020-12-14
Frontend Guide

TheLoneRonin:
### Description

- Goes over how to create GraphQL queries.

- Goes over how to create GraphQL mutations.

- Goes over how to integrate GraphQL into React functional components.

### DIFF
--- a/guides/FRONTEND.md
+++ b/guides/FRONTEND.md
@@ -1,14 +1,32 @@
-## MetaGame's FrontEnd API Guide
+# MetaGame's FrontEnd API Guide
 
-### Requirements
+## Requirements
 
 In order to develop MetaGame FrontEnd components, you need.
 
 - Node.js
 
-- Docker
+- Yarn
 
-You will need to be able to start the backend with:
+You can either choose to only develop the frontend and use our deployed backend services, or run the services on your own machines. If you're about to only work on UI, it's much quicker to connect to the remote servers.
+
+### Install dependencies
+
+```bash
+yarn
+```
+
+### Connecting to remote servers
+
+To connect to production servers, you'll need to create a file `.env.local` under `packages/web` and write this inside:
+
+```
+NEXT_PUBLIC_GRAPHQL_URL=https://hasura-7s6e.onrender.com/v1/graphql
+```
+
+
+### Connecting to local server
+If you want to connect to locally running backend services, you will need to be able to start the backend with:
 
 ```bash
 yarn docker:start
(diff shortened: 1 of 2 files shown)


guides/FRONTEND.md
```diff
@@ -0,0 +1,199 @@
...
+export const Component: FC = () => {
+  const [state, executeMutation] = useMutation(CreateMap);
+  const { fetching, error, data } = state;
+
+  async function insertMap(author, name) {
+    await executeMutation({ author, name });
+  }
+
+  return(
+    <button type="button" onClick={async () => await insertMap('Hello World', 'Test 1234')}>
```
pakokrew, line 184: Avoid putting asynchronous function in synchronous event handlers. Errors will be lost here. 

Better write something like 
```
function insertMap(author, name) {
   executeMutation({ author, name })
      .catch(console.error); // or stuff
  }
```

pakokrew (approved):
good one :)

guides/FRONTEND.md
TheLoneRonin: That is true!
Technically, you can write a try/catch in the event handler.

But the example provided is wrong.
Thanks, will update.

pakokrew: React event handlers are synchronous. Putting a try/catch could be a solution but the function would still be asynchronous. If by any mischance it throws, it leads to an unhandled exception, which is not good

TheLoneRonin:
@pakokrew can merge once you approve. (I do not have merge priveliges)

guides/FRONTEND.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Docker
+
+You will need to be able to start the backend with:
+
+```bash
+yarn docker:start
+```
+
+If you have difficulties running the backend, checkout the [Backend Guide](BACKEND.md).
```
hammadj, line 17: We dont need them to run the backend, they can just set the `NEXT_PUBLIC_GRAPHQL_URL` in their .env file to `https://hasura-7s6e.onrender.com/`

hammadj (changes requested):
Should update to include instructions on running only the frontend without needing to spin up the backend. Basically just need to add a .env.local file in the web package: https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables

pakokrew:
> Should update to include instructions on running only the frontend without needing to spin up the backend. Basically just need to add a .env.local file in the web package: https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables

Note to myself: The `.env.local` file must be in `packages/web` folder

pakokrew (approved):

PR #221, opened 2020-12-15
updated backend to read 3box collectiblesFavorite

dan13ram:


packages/backend/src/handlers/remote-schemas/resolvers/getBoxProfile/resolver.ts
```diff
@@ -47,3 +50,13 @@ function getImage(
   }
   return '';
 }
+
+function getCollectiblesFavourites(
+  collectiblesFavorites: Array<CollectiblesFavorite> | null | undefined,
```
pakokrew, line 55: the schema is
`collectiblesFavorites: [CollectiblesFavorites!]`

so maybe the return type should be only Array<CollectiblesFavorite> and always return an empty array if theres nothing in the profile

packages/backend/src/handlers/remote-schemas/resolvers/getBoxProfile/resolver.ts
```diff
@@ -34,6 +34,9 @@ export const getBoxProfile: QueryResolvers['getBoxProfile'] = async (
       height: 300,
     }),
     website: boxProfile.website,
+    collectiblesFavorites: getCollectiblesFavourites(
+      boxProfile?.collectiblesFavorites,
```
pakokrew, line 38: maybe that `?` is not necessary here

pakokrew (changes requested):

packages/backend/src/handlers/remote-schemas/resolvers/getBoxProfile/resolver.ts
```diff
@@ -47,3 +50,13 @@ function getImage(
   }
   return '';
 }
+
+function getCollectiblesFavourites(
+  collectiblesFavorites: Array<CollectiblesFavorite> | null | undefined,
+) {
+  if (!collectiblesFavorites) return null;
```
dan13ram, line 57: @pakokrew its handled here. The schema allows for null value to be returned.

packages/backend/src/handlers/remote-schemas/resolvers/getBoxProfile/resolver.ts
dan13ram: so do you want me to change the schema to `collectiblesFavorites: [CollectiblesFavorites!]!`

dan13ram: I wanted the schema to remain uniform with the rest of the values:
```
  type BoxProfile {
    ethereumAddress: String
    name: String
    description: String
    location: String
    job: String
    emoji: String
    imageUrl: String
    coverImageUrl: String
    website: String
    collectiblesFavorites: [CollectiblesFavorites!]
  }
```

dan13ram: also note the `type CollectiblesFavorite` in this line is different from our typeDef, and comes from 3box.

pakokrew: Ok my bad, didn't see that all the properties were nullable in the GraphQL schema. It's ok like this then !

pakokrew (approved):
LGTM

PR #215, opened 2020-12-14
MyMeta SEED Ticker

TheLoneRonin:
### Description

- Shows the price of SEEDs on the top header.

- Pulls data from CoinGecko.

- Adds `node-fetch` to mymeta. Which was already a child dependency.

### Screenshots

**Desktop**

**Mobile**

PR #223, opened 2020-12-15
MyMeta SEED Ticker

TheLoneRonin:
### Description

- Migrated from #215

- Shows the price of SEEDs on the top header.

- Pulls data from CoinGecko.

- Adds `node-fetch` to mymeta. Which was already a child dependency.

### Screenshots

**Desktop**

**Mobile**


pakokrew:
code LGTM

However writing '$31.2 SEEDS' is a bit confusing, I first thought it was how much I have in seeds. 

Maybe better to write something like 'SEED price: $31.2'

pakokrew (approved):

packages/web/components/Ticker.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <Flex alignItems="center" margin="15px">
+      <Image
+        src={MetaGameLogo}
+        alt="MetaGameLogo"
+        objectFit="contain"
+        boxSize="2rem"
+        margin="0 15px 0 0"
+      />
+      <P>${price.toFixed(2)} SEEDS</P>
```
hammadj, line 26: Try using this component instead, will look better: `<Heading size="xs" color="white">`

hammadj (approved):
Looks good, agree that we should change it to say SEED price or just "SEED - $33.22"

packages/web/components/Ticker.tsx
TheLoneRonin: Im going to update with the heading.
Will post new screenshot previews soon. Just gotta boot up the servers.

TheLoneRonin:
Thank you for the suggestion of the `Heading` component. The style looks much better!

image

image

PR #230, opened 2020-12-16
Feature/add timezone backend

alalonde:
I noticed an existing timezone (int) field already in the player table. Didn't want to delete it until I know it's not being used. I'm storing IANA timezones as they're more precise, and an hourly differential can be computed accurately from that.

Seems we should remove the timezone field if it's not being used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ out/
 .env.test.local
 .env.production.local
 
+lerna-debug.log*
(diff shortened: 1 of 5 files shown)


hammadj (approved):

pakokrew (approved):
Yeah could be interesting to check where that `timezone` prop is used and remove it in another PR

PR #232, opened 2020-12-17
Simple fix of a command

mikeki:
Just a quick guide update


pakokrew (approved):
That was the one needed !

PR #234, opened 2020-12-18
MyMeta - App Drawer

TheLoneRonin:
### Description

- Adds the app drawer for both desktop and mobile versions of the app

- Nice transitions and animations for on hover and click events

- Optimized mobile for several different screen resolutions

- Following video showcases how app drawer looks

https://user-images.githubusercontent.com/68780068/102654062-ed997600-413d-11eb-8189-a16fda7e5c84.mp4

### Notes 

- Not all links have been populated with the correct URL. Please notify me of what the links should have which URLs.


packages/web/pages/_app.tsx
```diff
@@ -17,6 +18,7 @@ const app: React.FC<AppProps> = ({ pageProps, Component }) => {
       </Head>
       <Web3ContextProvider>
         {!pageProps.hidePageHeader && <PageHeader />}
+        {!pageProps.hidePageHeader && <MobileFooter />}
```
mikeki, line 21: isn't this a bit confusing? hidePageHeader would hide the Footer

TheLoneRonin: They both have the same functionality.

TheLoneRonin: I do agree it could be confusing from variable.
Should discuss whether or not it's appropriate though!

mikeki:
LGTM in general. I don't have merge rights but overall looks great

packages/web/components/DrawerItems.tsx
```diff
@@ -0,0 +1,81 @@
...
+export interface DrawerItemType {
+  href: string;
+  isExternal?: boolean;
+  src: string;
+  alt: string;
+}
+
+export const DrawerItems = [
+  {
+    href: '/',
```
hammadj, line 21: Can we include the proper links in each of these? Shouldnt merge stuff into prod with broken links :)

packages/web/components/MobileFooter.tsx
```diff
@@ -0,0 +1,175 @@
...
+}) => {
+  return (
+    <MetaLink zIndex="2" href={href} isExternal={isExternal}>
+      <Button
+        display="flex"
+        flexDirection="column"
+        justifyContent="center"
+        alignItems="center"
+        width="20vw"
+        height="5rem"
```
hammadj, line 26: Should use standardized sizes from theme instead of hardcoded values

packages/web/components/MobileFooter.tsx
```diff
@@ -0,0 +1,175 @@
...
+  isExternal,
+}) => {
+  return (
+    <MetaLink
+      zIndex="2"
+      href={href}
+      isExternal={isExternal}
+      margin="0 !important"
+      width="7rem"
+      height="7rem"
```
hammadj, line 51: Why is the margin set like this? Also, shouldnt be using hard coded sizes, better to use responsive sizes or predefined sizes in theme

packages/web/components/MobileFooter.tsx
```diff
@@ -0,0 +1,175 @@
...
+      </MenuItem>
+
+      <Button
+        display="flex"
+        zIndex="11"
+        textDecoration="none"
+        variant="link"
+        flexDirection="column"
+        justifyContent="center"
+        alignItems="center"
```
hammadj, line 123: A lot of these props are being reused. MenuItem, SubMenuItem, and this all look like they are doing the same thing. Should consolidate the props and compose the components instead of repeating them, code will look much cleaner that way as well.

A good rule of thumb is to make things like MenuItem and SubMenuItem as generic and reusable as possible and declare them in their own files so that other people can also see it and re-use it. Defining one of specific components in files like this leads to a lot of repeated code which makes it hard to maintain and introduces inconsistencies in the UX/UI

packages/web/components/MobileFooter.tsx
```diff
@@ -0,0 +1,175 @@
...
+        top="0"
+        width="100vw"
+        height="calc(100vh - 5rem)"
+        background="linear-gradient(180deg, rgba(76, 63, 143, 0.5) 62.76%, rgba(184, 169, 255, 0.5) 100%);"
+        display="grid"
+        gridTemplateColumns="auto auto auto"
+        justify="center"
+        align="center"
+        padding="1rem 1rem 6rem 1rem"
+        transition="opacity 0.8s cubic-bezier(0.65, 0, 0.35, 1)"
```
hammadj, line 161: Seems like a lot of really specific hard coded values here too. This Drawer component should be separated out into its own component and reused in the PageHeader instead of implementing the same thing with minor tweaks twice. Otherwise if someone wants to update it, they have to know and remember to update it in two different places.

packages/web/components/PageHeader.tsx
```diff
@@ -1,28 +1,80 @@
...
+      isExternal={isExternal}
+      textDecoration="none"
+      _hover={{ textDecoration: 'none' }}
+    >
+      <Button
+        display="flex"
+        flexDirection="column"
+        justifyContent="center"
+        alignItems="center"
+        width="calc(32rem / 5)"
```
hammadj, line 35: Should use responsive styles and not hard coded ["magic numbers"](https://codeburst.io/software-anti-patterns-magic-numbers-7bc484f40544)

packages/web/components/PageHeader.tsx
```diff
@@ -1,28 +1,80 @@
...
       isExternal={isExternal}
       textDecoration="none"
       _hover={{ textDecoration: 'none' }}
     >
       <Button
+        display="flex"
+        alignItems="center"
+        justifyContent="center"
+        width="8.25rem"
+        height="8.25rem"
```
hammadj, line 68: no magic numbers, use theme styles

packages/web/components/PageHeader.tsx
```diff
@@ -1,28 +1,80 @@
...
         textDecoration="none"
         variant="link"
         p="1"
         fontFamily="mono"
         color="whiteAlpha.700"
-        _hover={{ color: 'pink.500', textDecoration: 'none' }}
+        className="filter-effect"
+        margin="1rem"
+        padding="1rem"
+        backgroundImage={`url(${MetaBoxButton})`}
```
hammadj, line 77: We shouldn't be using PNG images for this kind stuff, gets super buggy and doesnt respond well to dynamic layouts and screen sizes. Use components and styles instead.

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
       <MetaLink href="/" display="block" mr="10">
         <Image src={MetaGameImage} alt="MetaGame" mt={-2} w="9rem" />
       </MetaLink>
 
-      <Button
-        variant="link"
-        display={{ base: 'block', md: 'none' }}
-        onClick={handleToggle}
+      <Stack
+        width="48rem"
```
hammadj, line 106: No magic numbers

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+        height="100%"
+        direction="row"
+        justify="center"
+        align="center"
+        position="absolute"
+        top="0"
+        left="calc(50% - 24rem)"
+        padding="0 0 0.5rem 0"
```
hammadj, line 114: No magic numbers

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
-          fill="white"
-          width="1.5rem"
-          viewBox="0 0 20 20"
-          xmlns="http://www.w3.org/2000/svg"
+        <Image
+          src={MetaDrawer}
+          alt="MetaDrawer"
+          w="48rem"
+          position="absolute"
+          left="calc(50% - 24rem)"
```
hammadj, line 121: No magic numbers

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+          top="0"
+          zIndex="1"
+        />
+
+        <MenuItem
+          href="https://forum.metagame.wtf/c/quest/5/l/latest?board=default"
```
hammadj, line 127: This quest board isnt really used much. I think for the main items in this nav, it should have a link to the Wiki, a link to buy SEED, a link to the forums and a link to the newsletter

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
-        width={{ base: 'full', md: 'auto' }}
-        alignItems="center"
-        flexGrow={1}
-        my={{ base: 8, md: 0 }}
-      >
-        <Stack
-          direction={{ base: 'column', md: 'row' }}
-          spacing={{ base: 4, md: 6, lg: 10 }}
+        <MenuItem
+          href="https://forum.metagame.wtf/c/quest/5/l/latest?board=default"
```
hammadj, line 135: This is the same link as quests, and not something we should link to since it isnt used

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+          textDecoration="none"
+          variant="link"
+          flexDirection="column"
+          justifyContent="center"
+          alignItems="center"
+          width="calc(32rem / 5)"
+          className="filter-effect"
+          position="relative"
+          top={show ? '1rem' : '46rem'}
+          left="-0.25rem"
```
hammadj, line 154: No magic numbers

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+        direction="row"
+        flexWrap="wrap"
+        position="absolute"
+        zIndex="10"
+        top="5rem"
+        left="calc(50% - 16.5rem)"
+        padding="1rem 0"
+        transition="opacity 0.8s cubic-bezier(0.65, 0, 0.35, 1)"
+        opacity={show ? 0 : 1}
+        pointerEvents={show ? 'none' : 'inherit'}
```
hammadj, line 186: No magic numbers

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+      >
+        <Image
+          src={MetaBox}
+          alt="MetaBox"
+          w="33rem"
+          position="absolute"
+          left="calc(50% - 16.5rem)"
```
hammadj, line 193: No magic numbers, and we shouldnt be using an image for this. Doesnt animate well, doesn't work well with responsive styles, and slow to load. As a rule of thumb, we pretty much should never be using images as UI elements with very rare exceptions.

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
       justify="space-between"
       wrap="wrap"
       color="offwhite"
-      py="6"
+      py="4"
       px="8"
+      position="relative"
+      display={{ base: 'none', md: 'flex' }}
```
hammadj, line 99: If you remove this on the mobile version then there's no "Connect Wallet" button anymore. Not a fan of having two separate components for this reason, should try to make a single AppDrawer component that automatically updates its layout for mobile.

packages/web/components/MobileFooter.tsx
```diff
@@ -0,0 +1,175 @@
...
+        fontFamily="mono"
+        color="whiteAlpha.700"
+      >
+        {children}
+      </Button>
+    </MetaLink>
+  );
+};
+
+const SubMenuItem: React.FC<React.ComponentProps<typeof MetaLink>> = ({
```
hammadj, line 39: These components are being duplicated in MobileFooter and in PageHeader. Should refactor to make these generic and re-usable in separate files instead of being in the same file as the Footer (rule of thumb: only 1 component per file).

Then, consolidate the implementation of the AppDrawer to a single component instead of having the same code duplicated.

packages/web/components/PageHeader.tsx
```diff
@@ -41,63 +93,119 @@ export const PageHeader: React.FC = () => {
...
+        left="calc(50% - 24rem)"
+        padding="0 0 0.5rem 0"
       >
-        <svg
-          fill="white"
-          width="1.5rem"
-          viewBox="0 0 20 20"
-          xmlns="http://www.w3.org/2000/svg"
+        <Image
+          src={MetaDrawer}
```
hammadj, line 117: Shouldnt be using images for this. Use UI components and styles instead.

hammadj (dismissed):
The drawer is really difficult to read because the text underneath shows through. Need to add a blur to the background and change all the images you added for UI components to normal UI components. 

image

packages/web/components/DrawerItems.tsx
```diff
@@ -0,0 +1,81 @@
...
+import MetaTwitter from 'assets/drawer/twitter.png';
+
+export interface DrawerItemType {
+  href: string;
+  isExternal?: boolean;
+  src: string;
+  alt: string;
+}
+
+export const DrawerItems = [
```
pakokrew, line 19: This isn't a component, maybe move that file somewhere else
(and does not contain JSX => `.ts`)

packages/web/components/PageHeader.tsx
```diff
@@ -31,7 +83,7 @@ const MenuItem: React.FC<React.ComponentProps<typeof MetaLink>> = ({
 };
 
 export const PageHeader: React.FC = () => {
-  const [show, setShow] = React.useState(false);
+  const [show, setShow] = React.useState(true);
```
pakokrew, line 86: `useDisclosure` from chakra-UI

packages/web/pages/_document.tsx
```diff
@@ -24,6 +24,8 @@ class MetaDocument extends Document {
             rel="stylesheet"
           />
           <link rel="shortcut icon" href="/favicon.png" />
+          {/* @TODO: remove and apply filter with _hover={} when https://github.com/chakra-ui/chakra-ui/issues/2831 is resolved */}
```
pakokrew, line 27: They said this is fixed in v1.x of chakra UI, did you try updating it ?

pakokrew (changes requested):
I feel like the opacity transition comes after the translation of the metagame logo, it's a bit unnatural.

Maybe try to make these two transitions feel synchronous. Or just remove the opacity transition and open the menu like an actual drawer

packages/web/pages/_app.tsx
TheLoneRonin: @mikeki, I've changed this to `hideAppDrawer` btw! :)

packages/web/components/DrawerItems.tsx
TheLoneRonin: @pakokrew, has added updates to this in his PR :)

packages/web/components/MobileFooter.tsx
TheLoneRonin: I agree we should be using response sizing.
Will try to send in an update sometime tomorrow.

packages/web/components/MobileFooter.tsx
TheLoneRonin: @pakokrew, helped resolve this and created a `utils` folder!

packages/web/components/PageHeader.tsx
TheLoneRonin: Thanks for the link! I need to review this.

packages/web/pages/_document.tsx
TheLoneRonin: I'm going to have to roll in updates using the migration guide.

packages/web/components/PageHeader.tsx
TheLoneRonin: Also added it to mobile footer too!

pakokrew (approved):

PR #238, opened 2020-12-20
Consolidate contact UI with PlayerTile and Player page

hammadj:
No longer have two different components to display contact info


packages/web/components/Player/PlayerContacts.tsx
```diff
@@ -46,6 +48,18 @@ export const PlayerContacts: React.FC<Props> = ({ player }) => {
         }
         return null;
       })}
-    </>
+      {player.ethereum_address ? (
+        <Button
+          as="a"
+          href={`https://etherscan.com/address/${player.ethereum_address}`}
+          target="_blank"
```
dan13ram, line 55: does this also set `rel="noreferrer noopener"`?
if not we need to set it.

dan13ram (dismissed):

packages/web/components/Player/PlayerContacts.tsx
pakokrew: If it does not, could be great to create a property `external` on `Button` that sets both `target="_blank"` and `rel="noreferrer noopener"`

pakokrew (approved):
MyMeta start looking neat !

packages/web/components/Player/PlayerContacts.tsx
hammadj: Yea, the ds package needs a big refactor to use all the new chakra features that make this sort of thing much easier: https://chakra-ui.com/docs/features/chakra-factory

Added it manually for now

PR #239, opened 2020-12-20
Add guildname field to uniquely identify guilds with friendly id

hammadj:
basically guild version of username


dan13ram (approved):
lgtm

dan13ram:
@hammadj do we need to update schema as well or you'll do that with the frontend work?

hammadj:
> @hammadj do we need to update schema as well or you'll do that with the frontend work?

Will do in the frontend

PR #253, opened 2020-12-28
Drop the Player.timezone column since we have tz (string) now

alalonde:
As suggested by @pakokrew [here](https://github.com/MetaFam/TheGame/pull/230#pullrequestreview-554023859)

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -1,27 +1,5 @@
 - table:
     schema: public
-    name: Account
-  object_relationships:
-  - name: Player
-    using:
-      foreign_key_constraint_on: player_id
-  select_permissions:
-  - role: player
-    permission:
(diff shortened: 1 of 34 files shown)


hasura/metadata/tables.yaml
```diff
@@ -166,54 +166,6 @@
...
-      - role
-      - timezone
-      - tz
-      - username
-      filter:
-        id:
-          _eq: X-Hasura-User-Id
-      check: {}
-      set:
-        id: x-hasura-User-Id
```
hammadj, line 216: What are all these permission updates? Doesn't seem related to the timezone column

alalonde: Ugh.. there must be a bug in Hasura or something, as these permissions keep getting dropped when I make schema changes.

Issue #220, opened 2020-12-15
Rename database models


pakokrew:
Our model names in our database currently have different casings.

Would be better to have all of them written the same way.

Suggested casing for example Player Account: `player_account`

- [ ] Rename all the models

- [ ] Rename Account -> player_account

- [ ] Update all the frontend/backend graphql queries

alalonde:
I'll take a look at this since I almost tackled it in my timezone branch. I can see this causing great merge pain though for active contributors..

alalonde:
Fixed by #255

PR #255, opened 2020-12-30
Renamed table names to be consistent snake_case

alalonde:
Opening this for discussion. I have a pretty low tolerance for naming mess so here's my first take on fixing @pakokrew's suggestions in #220. 

I kept the enum tables as PascalCase in order to differentiate them.

I suspect there's still some broken stuff in here as my local instance is missing a bunch of types of data still.

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -1,27 +1,5 @@
 - table:
     schema: public
-    name: Account
-  object_relationships:
-  - name: Player
-    using:
-      foreign_key_constraint_on: player_id
-  select_permissions:
-  - role: player
-    permission:
-      columns:
-      - player_id
-      - identifier
-      - type
-      filter: {}
-  - role: public
-    permission:
-      columns:
-      - player_id
-      - identifier
-      - type
-      filter: {}
-- table:
-    schema: public
     name: AccountType
   is_enum: true
 - table:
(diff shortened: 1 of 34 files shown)


alalonde:
Also, how can we update the build such that the types are generated at the right time? Do we have to explicitly run the `autogen` task or can we just switch the backend to build before web ?

hammadj (approved):

hammadj:
> Also, how can we update the build such that the types are generated at the right time? Do we have to explicitly run the `autogen` task or can we just switch the backend to build before web ?

So the problem right now is that the web build is pointing at the currently deployed backend instead of the updated backend in this PR. Not an easy problem to solve unless we provision a full backend for every PR which can get expensive. Usually the best course of action is to update the backend and then update the frontend in another PR, but in this situation its not really feasible since the backend update would break the current frontend so they need to go out together.

I'm ok to merge as it is right now since the tests pass which is the most important part (i.e. types are not broken). It might break some stuff in prod but YOLO, we can patch it up later. 

Great work on this, really appreciate it and its been something bothering me for a while now. Extra heart for extra XP :)

pakokrew:
Thank you very much @alalonde for handling this. Much needed and not much fun to do !

@hammadj we can still afford breaking the prod for a short amount of time it's not dramatic, but at some point that will not be acceptable anymore. Having an environment deployed for each branch is complex to setup and very expensive, but a suboptimal solution would be to have two environments, one production and one staging. So we could "break" the staging for some time (merging backend changes, then merging frontend changes, then pulling this on production).

I also would like to start a discution about migrations. I don't feel like creating one file for each change is a good idea, we will endup with a lot of files. And I believe that one migration per pull request could be great, but I'd like to hear your point of view on that.

alalonde:
> Thank you very much @alalonde for handling this. Much needed and not much fun to do !

I heart refactoring, so it's my pleasure :D 

> @hammadj we can still afford breaking the prod for a short amount of time it's not dramatic, but at some point that will not be acceptable anymore. Having an environment deployed for each branch is complex to setup and very expensive, but a suboptimal solution would be to have two environments, one production and one staging. So we could "break" the staging for some time (merging backend changes, then merging frontend changes, then pulling this on production).

+1 for staging env. Could we provision this environment for builds? So build the backend, deploy it, then build the frontend, then revert it all back to `development` branch or the like. 

> I also would like to start a discution about migrations. I don't feel like creating one file for each change is a good idea, we will endup with a lot of files. And I believe that one migration per pull request could be great, but I'd like to hear your point of view on that.

I am all for squashing migrations when it makes sense. For this PR I think it would have made sense ;p

PR #262, opened 2021-01-06
Renamed player columns to be consistent snake_case

alalonde:
Also deleted old `timezone` column and renamed `tz` to `timezone`

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -161,7 +161,7 @@
       foreign_key_constraint_on: enneagram
   - name: playerType
     using:
-      foreign_key_constraint_on: playerTypeId
+      foreign_key_constraint_on: player_type_id
   array_relationships:
   - name: Accounts
     using:
@@ -211,13 +211,12 @@
       - enneagram
       - ethereum_address
       - id
-      - playerTypeId
+      - player_type_id
       - rank
       - role
-      - scIdentityId
+      - sc_identity_id
+      - total_xp
       - timezone
-      - totalXp
-      - tz
       - username
       filter: {}
   - role: public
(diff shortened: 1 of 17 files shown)


hammadj (commented):
Noticed there's still some other columns that aren't snake_cased yet, namely:
- EnneagramType
- GuildType
- AccountType
- PlayerRank
- SkillCategory

Do you plan to refactor those in a separate PR?

alalonde:
> Noticed there's still some other columns that aren't snake_cased yet, namely:
> 
>     * EnneagramType
> 
>     * GuildType
> 
>     * AccountType
> 
>     * PlayerRank
> 
>     * SkillCategory
> 
> 
> Do you plan to refactor those in a separate PR?

Those are all enum tables so I left them as they are to distinguish them from the other data tables. I could go either way per renaming them as well.

pakokrew:
I suddenly realised: why renaming the timezone property to `tz` ? I believe you did that because there was an other one and didn't want conflict, but now that the old one is not used anymore, why not calling this one `timezone`? It would be much cleaner like this. Newcomers reading `tz` may be confused

> Those are all enum tables so I left them as they are to distinguish them from the other data tables. I could go either way per renaming them as well.

I am okay to case enums differently from tables, but open to other suggestions than PascalCase form them.

alalonde:
> I suddenly realised: why renaming the timezone property to `tz` ? I believe you did that because there was an other one and didn't want conflict, but now that the old one is not used anymore, why not calling this one `timezone`? It would be much cleaner like this. Newcomers reading `tz` may be confused

That's actually exactly what I've done, it's just a bit confusing due to the diff. `timezone` is the old `tz` column

hammadj:
> Those are all enum tables so I left them as they are to distinguish them from the other data tables. I could go either way per renaming them as well.

Ah ok, makes sense, the differentiation is nice. Approved!

hammadj (approved):

Issue #254, opened 2020-12-29
[App Drawer] Fixing The App Drawer


peth-yursick:
As promised, time to expand on the issues mentioned for the app drawer!
![image](https://user-images.githubusercontent.com/55857222/103298590-f837f300-49fa-11eb-824c-693cd8cccb44.png)

Here's a few:
- The drawer itself should be at least 30% thinner. Looks too big now + shows the noise the parts that are pasted from figma.
- The apps in the bar itself should be centered, currently aligned top
![image](https://user-images.githubusercontent.com/55857222/103298985-d5f2a500-49fb-11eb-90d8-5456828e11bc.png)
- The drawn out part should be smaller than the drawer on all sides, currently just on top left
![image](https://user-images.githubusercontent.com/55857222/103299541-14d52a80-49fd-11eb-88ba-41fb7ff19bab.png)
- The drawer should be transparent blur like the cards in the profile & NFT gallery 
- The font should be IBM Plex Sans (I think currently its just IBM Plex)
- There should be no underscore when hovering buttons
- The apps in the drawer should have equal spacing on all sides, like in the original design
![image](https://user-images.githubusercontent.com/55857222/103299164-34b81e80-49fc-11eb-8a64-2ee0faf00a37.png)
~~- There should be no blue frame around the logo~~
- There should be no blue frame around the buttons
![image](https://user-images.githubusercontent.com/55857222/106525161-d7742900-64e3-11eb-9485-3dcf4f4910a5.png)
- When the app drawer is open, the shortcuts from the navigation bar should move from the bar to the drawer itself. 
![image](https://user-images.githubusercontent.com/55857222/103299686-5d8ce380-49fd-11eb-946f-fdd0420f4b87.png)
~~- Different version of the logo should be used~~
![Pepe-ETH-finalspaceylight](https://user-images.githubusercontent.com/55857222/103299227-56190a80-49fc-11eb-9617-5eff90a20ca4.png)

- **Most importantly, the apps should be opening in a window like the NFT gallery, not in new pages.**

TheLoneRonin:
Perfect! Thanks for providing all of this info @peth-yursick, there are actually refactors we need to do to the source too. So being able to resolve all this is important.

Please let me get the ball moving on metamaps first then we can move forward on the drawer.

Cheers!

peth-yursick:
> 
> 
[...]

PR #294, opened 2021-01-20
Cross Browser Gradient for App Drawer

TheLoneRonin:
### Description

- Uses `backdrop-filter` when the Browser can support it.

- Removes the `.png` image for a CSS gradient.

- Adds the shine effect as seen on: https://wiki.metagame.wtf/docs/home when you hover over item elements.

#### Chrome

https://user-images.githubusercontent.com/68780068/105202099-9d804b80-5b0f-11eb-9a8a-e7708846f580.mp4

#### Firefox

https://user-images.githubusercontent.com/68780068/105202111-a244ff80-5b0f-11eb-9865-9e5ed788aa0a.mp4

-- EDIT

paired with @dysbulic


TheLoneRonin:
Before merging, need to make sure that the preview is the same as the dev environment.
For some weird reason the preview does not have filter gradient on Chrome.

Just want to confirm is this potentially my cache? Hopefully not!

dysbulic:
Chromium is still borked:

Screenshot from 2021-01-21 03-37-00

Chrome looks good though under Linux:

Screenshot from 2021-01-21 03-38-02

packages/web/components/PageHeader.tsx
```diff
@@ -217,9 +243,30 @@ export const PageHeader: React.FC = () => {
...
+            }
+            style={{ backdropFilter: 'blur(10px)' }}
+          />
+          <Image
+            position="absolute"
+            left="calc(50% - 16.5rem)"
+            top={678}
+            src={MetaGradientBottom}
+            alt="MetaGradientBottom"
+            width={522}
```
dysbulic, line 266: Should `left` and `width` be in the same units so there's no possibility of misaligning things.

(I'm assuming you're centering something.)

dysbulic (approved):

packages/web/components/PageHeader.tsx
```diff
@@ -217,9 +243,30 @@ export const PageHeader: React.FC = () => {
...
-          <Box position="absolute" left="calc(50% - 16.5rem)" top="0">
-            <NextImage src={MetaBox} alt="MetaBox" width={528} height={695} />
-          </Box>
+
+          <Box
+            position="absolute"
+            left="calc(50% - 16.5rem)"
+            top="0"
+            width={522}
+            height={680}
```
hammadj, line 252: Can we use responsive sizing instead of hard coded width/heights? Hardcoded sizes is why the drawer is breaking / looking weird on a lot of devices

packages/web/components/PageHeader.tsx
hammadj, line 266: Shouldn't use hardcoded values. Also, Chakra has a native Gradient component now, shouldn't use an image since it takes a lot more bandwidth to load vs plain CSS.

packages/web/components/PageHeader.tsx
```diff
@@ -77,11 +79,42 @@ const SubMenuItem: React.FC<React.ComponentProps<typeof MetaLink>> = ({
...
+          position: 'absolute',
+          display: 'block',
+          top: 0,
+          left: 0,
+          width: '100%',
+          height: '100%',
+          filter: 'blur(10px)',
+          transform: 'translate3d(-120%, 0, 0)',
+          background:
+            'linear-gradient(45deg, transparent, rgba(255, 255, 255, 0), rgba(255, 255, 255, 0.06), rgba(255, 255, 255, 0), transparent)',
```
hammadj, line 109: A lot of repeated code / styles, should consolidate and clean it up. You dont need 90% of these if you follow Chakra UI method of styling

hammadj (changes requested):

dysbulic:
@TheLoneRonin, @alalonde asked if I would [look at the display issues with Chromium](https://github.com/MetaFam/TheGame/issues/254#issuecomment-766457797). I was going to work off this branch since @hammadj suggested it is [the use of absolute units that is the root of the problem](https://github.com/MetaFam/TheGame/pull/294#discussion_r562357561).

The tasks are:

- [ ] leverage Chakra UI to reduce the amount of repeated styling
- [ ] change absolute units to relative ones
- [ ] use the native Chakra UI gradient component

It's entirely possible you've already done this stuff. If so, just drop me a line.

packages/web/components/PageHeader.tsx
dysbulic: Things are now in terms of `rem`s. String interpolation is used along with `calc()` to create comprehensible sizing instructions.

It's feasible that some more of those statements could be derived from `flex` or `grid` layouts, but the current implementation works.

packages/web/components/PageHeader.tsx
dysbulic: The image is the bottom cap on the drawer. The background is a CSS gradient.

packages/web/components/PageHeader.tsx
dysbulic: The bulk of the repetition was a `_before` with a `display` of `none`. Removing it didn't affect the display, so I removed it.

pakokrew:
- Drop blur doesn't work on my browser (Brave 1.19) and even making weird flickering.

- Button logos are still starting to load with a wrong placeholder image. That was fixed by replacing `Image` component by [the one of nextjs ](https://nextjs.org/docs/api-reference/next/image), but it look like what I've done have been reverted.


Uploading Screen Recording 2021-01-31 at 13.22.41.mov…


Uploading Screen Recording 2021-01-31 at 13.21.07.mov…

dysbulic:
> - Button logos are still starting to load with a wrong placeholder image. That was fixed by replacing Image component by the one of nextjs , but it look like what I've done have been reverted.

The issue with the Next.js `Image` is that the dimensions have to be specified in pixels and switching to a responsive layout meant that they weren't.

I can hopefully simply put a sized `div` around it and then use `Image`'s `layout='fill'`.

hammadj (changes requested):
Blur not working in Safari, CI failing due to linting errors

image

packages/web/components/MobileFooter.tsx
```diff
@@ -93,66 +89,65 @@ export const MobileFooter: React.FC = () => {
...
+            display="flex"
+            textDecoration="none"
+            variant="link"
+            flexDirection="column"
+            justifyContent="center"
+            alignItems="center"
+            onClick={onToggle}
+            _focus={{ boxShadow: 'none' }}
+          >
+            <Image // TODO use NextImage component once images are without text
```
pakokrew, line 134: Use next/image

packages/web/components/PageHeader.tsx
```diff
@@ -90,6 +116,27 @@ const SubMenuItem: React.FC<React.ComponentProps<typeof MetaLink>> = ({
   );
 };
 
+const BoxedNextImage: React.FC<
```
pakokrew, line 119: Don't know if this is good practice, but that's a good workaround for next/image !

Maybe we could put this in the `design-system` package instead ? Or at least in a component of its own in `web`, so it could be reusable (🔔 MobileFooter)

packages/web/components/MobileFooter.tsx
```diff
@@ -93,66 +89,65 @@ export const MobileFooter: React.FC = () => {
...
-          variant="link"
-          flexDirection="column"
-          justifyContent="center"
-          alignItems="center"
-          width="20vw"
-          height="5rem"
-          onClick={onToggle}
+        {DrawerItemsLeft.map((item) => (
+          <MenuItem href={item.href} isExternal={item.isExternal} key={item.href}>
+            <NextImage src={item.src} alt={item.alt} width={35} height={35} />
```
pakokrew, line 107: these are still with fixed size maybe use new `BoxedNextImage` component

packages/web/components/MobileFooter.tsx
```diff
@@ -93,66 +89,65 @@ export const MobileFooter: React.FC = () => {
...
 
-      {DrawerItemsRight.map((item) => (
-        <MenuItem href={item.href} isExternal={item.isExternal} key={item.href}>
-          <NextImage src={item.src} alt={item.alt} width={35} height={35} />
-          {item.text}
-        </MenuItem>
-      ))}
+        {DrawerItemsRight.map((item) => (
+          <MenuItem href={item.href} isExternal={item.isExternal} key={item.href}>
+            <NextImage src={item.src} alt={item.alt} width={35} height={35} />
```
pakokrew, line 146: same as l.107

packages/web/components/MobileFooter.tsx
```diff
@@ -162,28 +157,30 @@ export const MobileFooter: React.FC = () => {
...
-            return (
-              <SubMenuItem href={item.href} key={item.alt}>
-                <SubImage src={item.src} alt={item.alt} />
-                {item.text}
-              </SubMenuItem>
-            );
-          })}
+          {DrawerSubItems.map((item) => (
+            <SubMenuItem href={item.href} key={item.alt}>
+              <SubImage src={item.src} alt={item.alt} />
```
pakokrew, line 180: BoxedNextImage

packages/web/components/PageHeader.tsx
```diff
@@ -217,17 +272,46 @@ export const PageHeader: React.FC = () => {
...
+                `linear-gradient(
+                  180deg, rgba(76, 63, 143, ${opacity}) 62.76%,
+                  rgba(184, 169, 255, ${opacity}) 100%
+                )`
+              )
+            })()}
+            style={{ backdropFilter: 'blur(10px)' }}
+          />
+          {/* ToDo: switch to BoxedNextImage */}
+          <Image
```
pakokrew, line 294: BoxedNextImage

packages/web/components/PageHeader.tsx
```diff
@@ -100,37 +147,36 @@ export const PageHeader: React.FC = () => {
...
-        <Box position="absolute" left="calc(50% - 23.5rem)" top="0" zIndex="1">
-          <NextImage
+        {/* margin-inline-start = 0.5rem is set and I don't know how to remove it */}
+        <Box
+          position="absolute"
+          left={`calc(50% - (${drawer.width} / 2) + 0.5rem)`}
+          top="0"
+          zIndex="1"
+        >
+          <Image
```
pakokrew, line 175: BoxedNextImage

pakokrew (changes requested):
Few images components were forgotten.
Also some linting errors.

On brave 1.19, drop blur filter works for in MobileFooter but not in PageHeader

Good luck, we're almost good !

packages/web/components/PageHeader.tsx
dysbulic: My understanding is that `NextImage`'s primary function is to load the lowest resolution image possible for a given display size.

It is using [`scrset` and `sizes`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_different_sizes), but since it doesn't enlarge an image past its original resolution that isn't really affecting these images because they're so small.

I'll:
- [ ] move the `BoxedNextImage` to the design system
- [ ] change the references in the page header
- [ ] add usage in the page footer

dysbulic:
> Blur not working in Safari, CI failing due to linting errors

I'm pretty sure the problem is that Next.js is rendering the pages server-side where `background-filter` isn't supported. This disables the opacity of the drawer.

I [posted to StackOverflow](https://stackoverflow.com/questions/66017821/how-to-handle-unsupported-css-properties-in-next-js-or-use-supports-directives) on the issue.

packages/web/components/PageHeader.tsx
pakokrew: I didn't dig into the details of the next/image component, and maybe I'm wrong saying it's better to use it for that particular case.

What is sure, is that there shouldn't be a placeholder image like we see in the video I uploaded while they are loading. This is not the case in other parts of the app. So creating a custom image component is a good idea, where we will be able to switch either to next/image or to a custom solution solving that issue.

packages/web/components/LoginButton.tsx
```diff
@@ -45,16 +46,25 @@ export const LoginButton: React.FC = () => {
             {user.player ? getPlayerName(user.player) : 'Unknown'}
           </MetaLink>
           <HStack spacing={2}>
-            <MetaLink href="/profile/setup">Setup profile</MetaLink>
+            <MetaLink
```
pakokrew, line 49: Why these changes in this branch ?

pakokrew:
I think I solved the issue. There was an error in the web console:
`next js Prop `className` did not match. Server: css-azed Client:css-12J21`

Problem is well explain here: https://www.joshwcomeau.com/react/the-perils-of-rehydration/
The page rendered by the server and the first render of react MUST be the same, otherwise react is lost and can't rehydrate correctly.

packages/web/components/LoginButton.tsx
dysbulic: When you narrow the screen when the profile options are written out they are wide enough they push the profile picture behind the drawer. The switch to icons was to narrow everything.

hammadj:
> I think I solved the issue. There was an error in the web console:
> `next js Prop 'className' did not match. Server: css-azed Client:css-12J21`
> 
> Problem is well explain here: https://www.joshwcomeau.com/react/the-perils-of-rehydration/
> The page rendered by the server and the first render of react MUST be the same, otherwise react is lost and can't rehydrate correctly.
> 
> It is left to replace all the images component that doesn't use the new one.

Works in Chrome but still doesn't work in Safari. Here's the code I usually use for background blur that works in Safari as well: https://github.com/expo/expo/blob/sdk-40/packages/expo-blur/src/BlurView.web.tsx. We could literally just import that expo/blur library and use it instead of doing all our own implementation too.

Screenshot from Safari of that BlurView component: 
image

Screenshot of AppDrawer in Safari:
image

dysbulic:
@hammadj, Safari needs the `-webkit-` prefix for `backdropFilter`. I added it, so it should be sorted.

packages/web/components/MobileFooter.tsx
dysbulic: Done.

packages/web/components/MobileFooter.tsx
dysbulic: Done.

packages/web/components/MobileFooter.tsx
dysbulic: Done.

packages/web/components/PageHeader.tsx
dysbulic: This one is the only one that's tricky. It has `max-width`  and `max-height` set to `100%`. It will expand beyond it's original dimensions on the y-axis, but not on the x.

I went into the debugger and added random CSS props & tried lots of different values, but it just wouldn't line up when using `BoxedNextImage`.

packages/web/components/PageHeader.tsx
dysbulic: Done.

packages/web/components/MobileFooter.tsx
dysbulic: Done.

packages/web/pages/guild/[guildname].tsx
```diff
@@ -81,7 +81,7 @@ export const getStaticProps = async (
 
   return {
     props: {
-      guild,
+      guild: guild || null,
```
Comment (dysbulic) [inline packages/web/pages/guild/[guildname].tsx]: This will give the incorrect output when `guid` is `false` (or anything other than undefined).

dysbulic (approved):
I'm a little bleary eyed, but I see nothing glaringly wrong with this code.

I've tested a reasonable amount on the browsers available to me.

alalonde:
Just checked the latest deployment, backdrop blur effect looks fine on safari/FF mac. lgtm

pakokrew (approved):

dysbulic:
@TheLoneRonin, I'd like to split the credit with this with you. ¿Is that acceptable?

@other-author: @dysbulic

pakokrew:
> @TheLoneRonin, I'd like to [split the credit](https://sourcecred.io/docs/beta/plugins/github/#edges:~:text=%40other%2Dauthor) with this with you. ¿Is that acceptable?
> 
> @other-author: @dysbulic

I've already allowed myself to edit the main post to credit you 😇

pakokrew:
Let's merge this.

peth-yursick:
> 
> 
> Let's merge this.

still shows gradient on Firefox on Windows

PR #296, opened 2021-01-22
Add case in SC migration script where user doesnt already have a SC identityID

hammadj:


hammadj:
instamerging for quick bug fix, feel free to review after

PR #299, opened 2021-01-24
Fixed flashing on load with mobile footer drawer

### LINKED ISSUE #268: [App Drawer] Mobile Issues
There are currently 4 known issues on the mobile app drawer:
~~- When loading MyMeta, the appdrawer quickly flashes as if its open then closes. It should be closed by default & not appearing in a flash like that.~~
~~- The buttons have no titles~~
~~- The crystal still has the blue frame thing~~
- The light blue app drawer background appears over a part of the navigation bar
~~- There is a blue frame around the logo button~~

alalonde:
Resolves #268

### DIFF
--- a/packages/web/components/MobileFooter.tsx
+++ b/packages/web/components/MobileFooter.tsx
@@ -158,8 +158,8 @@ export const MobileFooter: React.FC = () => {
         animate={isOpen ? 'show' : 'hide'}
         transition={{ duration: 0.25 }}
         variants={{
-          show: { opacity: 1, pointerEvents: 'inherit' },
-          hide: { opacity: 0, pointerEvents: 'none' },
+          show: { display: 'block', opacity: 1, pointerEvents: 'inherit' },
+          hide: { display: 'none', opacity: 0, pointerEvents: 'none' },
         }}
         onClick={onClose}
       >


packages/web/components/MobileFooter.tsx
```diff
@@ -158,8 +158,8 @@ export const MobileFooter: React.FC = () => {
         animate={isOpen ? 'show' : 'hide'}
         transition={{ duration: 0.25 }}
         variants={{
-          show: { opacity: 1, pointerEvents: 'inherit' },
-          hide: { opacity: 0, pointerEvents: 'none' },
+          show: { display: 'block', opacity: 1, pointerEvents: 'inherit' },
+          hide: { display: 'none', opacity: 0, pointerEvents: 'none' },
```
dysbulic, line 162: If you're using `display`, do you need `opacity` and `pointerEvents`?

dysbulic, line 162: The advantage to using `opacity` is that it can be animated.

alalonde: pointerEvents probably not. To be clear this PR doesn't attempt to do any refactoring / cleanup which is sorely needed.

dysbulic:
The right way to fix this is to add `style={{ opacity: 0 }}` to the `motion.div`. This causes it to render initially as transparent and avoids a rerender by the animating component.

`display` is a non-animatable property, so you won't get the fade in/out using it.

hammadj (approved):
Approving since it fixes the base problem but agree that the drawer needs a major refactor / rework

dysbulic:
I'm working on switching everything to responsive units currently in `@mymeta/responsive-drawer` which is based off @TheLoneRonin's `@mymeta/drawer-gradient`.

I've included the opacity fix outlined above and the mobile layout looks good except for on a tablet where things are too big.

For the time being, I was considering using the header version on a tablet.

PR #300, opened 2021-01-24
Use the footer on tablets (< lg, > md)

alalonde:

PR #305, opened 2021-01-25
[MyMeta] Use responsive units on the header and footer

dysbulic:
This is a refactor of the app drawer to use responsive units. It includes changes from pull requests #294, #299, & #300.


dysbulic:
The only other thing I would suggest is not to use a monospace font for the drawer component names. They are too close together as it is.

dysbulic:
I am in the process of removing the units from the buttons to allow them to flow more naturally. I should have an update later today…

hammadj (commented):
Feels much better now, however the spacing is a bit weird on mobile still where the footer height feels too high and the padding of the app icons container is not right (too much padding on the bottom)

dysbulic:
I switched the base branch to `@mymeta/drawer-gradient` so that @TheLoneRonin would still get XP from the merging of #294.

dysbulic:
I removed most of the explicit sizes from the footer menu elements so that they get their dimensions from the layout flow.

It had the unintended side effect of making the drawer work in landscape mode on mobile by shrinking the icons to nothing.


https://user-images.githubusercontent.com/181523/106229327-5eab5f00-61bb-11eb-84f4-a1ae26d11815.mp4

dysbulic:
I have the convenient function `function t() { git commit -S -am "$*"; }` defined in my `.bashrc` so I can do `t this is my commit message` to commit. If I don't pay attention though, the `#` character is interpreted as the start of a comment.

My last message should have read:

> removed box from focused menu items ala. #268

There was a blue box shadow around menu items when they were focused. That is removed.

peth-yursick:
> 
> 
> I have the convenient function `function t() { git commit -S -am "$*"; }` defined in my `.bashrc` so I can do `t this is my commit message` to commit. If I don't pay attention though, the `#` character is interpreted as the start of a comment.
> 
> My last commit message should have read:
> 
> > removed box from focused menu items ala. #268
> 
> There was a blue box shadow around menu items when they were focused. That is removed.

Does this fix everything from #268 then?

dysbulic:
I'm pretty sure everything in #268 is fixed.

- [x] Hide on load
- [x] Remove blue boxes around menu items when focused

I didn't fix "The light blue app drawer background appears over a part of the navigation bar", but I'm not seeing it.

hammadj (approved):

PR #314, opened 2021-01-29
Remove unused codegen stuff and update Hasura to latest version

hammadj:


dan13ram (approved):
love it when i see so much red

PR #315, opened 2021-01-29
Use exact versions in package.json files

hammadj:
Since Dependabot is updating packages for us, we should only use exact version numbers to be explicit
for what version we are on and reduce chances of having incorrect versions installed


dysbulic (approved):
I follow your logic for using explicit version numbers, and these changes all seem to be working in the preview version.

dan13ram (approved):
lgtm

pakokrew:
yarn.lock is here to guarantee we have the same package versions

PR #318, opened 2021-01-29
Codegen wasn't completely removed

dysbulic:
#316 removes the `codegen` task, but there were still references remaining that prevented building a clean docker distribution.


hammadj (approved):
good catch, thanks

PR #319, opened 2021-01-30
Discord Bot MVP

hammadj:
Setup of discord bot architecture

### DIFF
--- a/packages/discord-bot/codegen.yml
+++ b/packages/discord-bot/codegen.yml
@@ -1,0 +1,18 @@
+overwrite: true
+require:
+  - ts-node/register
+generates:
+  ./src/__generated__/hasura-sdk.ts:
+    schema: '../../schema.graphql'
+    documents:
+      - ./src/graphql/**/(!(*.d)).ts
+    plugins:
+      - typescript
+      - typescript-operations
+      - typescript-graphql-request
+      - add:
+          content: '/* eslint-disable */'
+    config:
+      immutableTypes: true
+      scalars:
+        account_type: "'ETHEREUM' | 'DISCORD' | 'GITHUB' | 'TWITTER' | 'DISCOURSE'"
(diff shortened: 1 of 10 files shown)


dysbulic (approved):
This seems pretty straightforward.

packages/discord-bot/src/index.ts
```diff
@@ -0,0 +1,19 @@
+// Use the Client that are provided by @typeit/discord NOT discord.js
+import { Client } from '@typeit/discord';
+
+import { CONFIG } from './config';
+
+async function start() {
+  const client = new Client({
+    classes: [
+      `${__dirname}/*Discord.ts`, // glob string to load the classes
```
dysbulic, line 9: This is causing the tests to fail.

dysbulic:
The fixes to the build check (elements returning `undefined` which can't be serialized as JSON) are in #294.

packages/discord-bot/codegen.yml
```diff
@@ -0,0 +1,18 @@
+overwrite: true
+require:
+  - ts-node/register
+generates:
+  ./src/__generated__/hasura-sdk.ts:
+    schema: '../../schema.graphql'
+    documents:
+      - ./src/graphql/**/(!(*.d)).ts
```
dysbulic, line 8: This is causing the tests to fail when it doesn't match anything.

pakokrew:
Maybe add dockerfile + compose configurations ?

hammadj:
> The fixes to the build check (elements returning `undefined` which can't be serialized as JSON) are in #294.

Would be better to make changes directly to master next time. When PRs get too big and are trying to do too many things at once it doesn't work out well. If you want to split XP with another player you can write "paired with @username" in the PR description

hammadj:
> Maybe add dockerfile + compose configurations ?

Im intending this to be used as a package / imported by the backend instead of running as its own service, so didn't make a Dockerfile for that reason. If you think its worth it just for dev purposes I can add it in

pakokrew:
> > Maybe add dockerfile + compose configurations ?
> 
> Im intending this to be used as a package / imported by the backend instead of running as its own service, so didn't make a Dockerfile for that reason. If you think its worth it just for dev purposes I can add it in

Thought it would become a service of its own. Then I guess we don't need docker configs.

pakokrew (approved):

Issue #295, opened 2021-01-21
[MetaMaps] Only authenticate when necessary


dysbulic:
Currently if a user has not connected their $ETH wallet they are not able to do anything other than look at a message asking them to authenticate.

The application should only ask the user to log in when they are performing an interaction, such as saving a map, that needs to be tied to their account.

If they are not logged in when they start, a new ephemeral map should instantiated that they can edit anonymously.

There should also be a `Connect` button in the same place as on [my.metagame.wtf](//my.metagame.wtf). Ideally the component from `packages/web/components/LoginButton.tsx` should be abstracted to `packages/design-system` and then included in MetaMaps via the import of `@metafam/ds`.

ace-smart:
I would like to work on this issue?

ace-smart:
@dysbulic @TheLoneRonin In the MetaMaps, the `Setup Profile` link is needed still?
I think `Setup Profile` feature should be available on only `MyMeta`.
What do you think about this?
![image](https://user-images.githubusercontent.com/76216456/106624306-f4196b00-658e-11eb-878e-8f1ca3382825.png)

dysbulic:
> [Is] the `Setup Profile` link is needed still?

I think we need the ability to insert content by the profile icon. I [suggested an 'options` parameter](https://github.com/MetaFam/TheGame/pull/328/files/0a9ac1532c5ef95dc37ea806d0d6b7e6d5ffbdd6#r568305756) in #328.

IxaBrjnko:
If we can pull the merge from MetaMaps Alpha, this would be a lot easier to work on in the main repo...

dysbulic:
MetaMaps has been removed from active development.

PR #328, opened 2021-02-02
Made abstracted Login Button component into the design-system. 

ace-smart:
As mentioned in https://github.com/MetaFam/TheGame/issues/295#issue-790857701:
Connect button has been abstracted to packages/design-system.


packages/web/components/PageHeader.tsx
```diff
@@ -93,6 +103,15 @@ const SubMenuItem: React.FC<React.ComponentProps<typeof MetaLink>> = ({
 export const PageHeader: React.FC = () => {
   const { isOpen, onToggle, onClose } = useDisclosure();
 
+  const { connectWeb3, disconnect, isConnected } = useContext(Web3Context);
```
hammadj, line 106: There's a `useWeb3` hook in web/lib/hooks.js. you can use that instead of this

packages/design-system/src/LoginButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Box>
+          <Link
+            href={`/player/${user.username}`}
+            fontFamily="body"
+            color="whiteAlpha.700"
+          >
+            {user.player ? user.playerName : 'Unknown'}
+          </Link>
+          <HStack spacing={2}>
+            <Link color="cyan.400" href="/profile/setup">
```
hammadj, line 62: This has implementation details that don't belong in the design system (profile setup route has to do with NextJS in the `web` package) since they cant be universally used. Would be better to just use the generic button component or abstract the implementation details of the NextJS app.

hammadj (approved):

packages/design-system/src/LoginButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { MetaButton } from './MetaButton';
+
+type UserInfo = {
+  playerImage?: string;
+  playerName?: string;
+  username?: string;
+  player?: any;
+};
+
+type LoginButtonPros = {
```
hammadj, line 21: Typo

packages/design-system/src/LoginButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+      return <Spinner color="purple.500" size="sm" />;
+    }
+    if (!user?.player) return null;
+    return (
+      <HStack>
+        <Avatar
+          src={user.playerImage}
+          name={user.playerName}
+          as={Link}
+          href={`/player/${user.username}`}
```
dysbulic, line 51: This URL may not exist on all sites where this button is used.

Maybe you could have a parameter `hrefFor` with the default value ``(user) => `/player/${user.username}`​``, then for the `href` parameter call `hrefFor(user)`.

packages/design-system/src/LoginButton.tsx
dysbulic: The tricky part is whether to always include the `Disconnect` button. It is arguably always warranted, but, for example, I recently pushed a change to MyMeta to replace the `Disconnect` text with a `⨯`.

Maybe `mutations` could be a function with a default value of the `Disconnect` button, but which receives the `disconnect` function as a parameter allowing the generation of whatever UI is desired.

Is that too convoluted?

pakokrew:
It's not a bad idea to put the LoginButton in the design system, but I think the most important part is the web3 hooks that should be in a shared package. This has a complex logic that must be the same between apps and shouldn't be reimplemented multiple times. The LoginButton is a really simple component.

For the problems you have about "Setup profile" considering you want to share the LoginButton, then it should only display "connect" or "profile info + disconnect". "Setup profile" should be moved somewhere else. 
When connected, the profile info could redirect to a page url set as a param, where the user could have a way to edit their profile. 

The "Setup profile" will be changed anyway. Once the user is onboarded, they will not have to go through the all wizard to edit their profile and they should be able to atomically edit their profile on their page.

pakokrew (changes requested):
see my previous comment

ace-smart:
> see my previous comment

I am also thinking of building shared web3 context.
However, I am curious if apps has its own dependencies and auth mechanism or something else...

pakokrew:
> > see my previous comment
> 
> I am also thinking of building shared web3 context.
> However, I am curious if apps has its own dependencies and auth mechanism or something else...

Metagame apps will definitely share the same auth mechanism, and changes to it should be reflected on ever apps. In one app really need a different auth system, it can implement its own. At least now for Metamaps it should be the same.

ace-smart:
> > > see my previous comment
> > 
> > 
> > I am also thinking of building shared web3 context.
> > However, I am curious if apps has its own dependencies and auth mechanism or something else...
> 
> Metagame apps will definitely share the same auth mechanism, and changes to it should be reflected on ever apps. In one app really need a different auth system, it can implement its own. At least now for Metamaps it should be the same.

Ok, you are right.
May I make another PR of the update including web3 context?
If you ok, I want you to approve this PR and merge.

pakokrew:
> Ok, you are right.
> May I make another PR of the update including web3 context?
> If you ok, I want you to approve this PR and merge.

Actually I don't really see the point of this PR. If this is just about sharing a "LoginButton component", which includes very few logic and that will be anyway different between apps, I am not sure it is necessary. 

I would highly prefer to work first on sharing the auth mechanism with the Web3 context and hooks, integrate them into MyMeta and MetaMaps, see how that fits, and then maybe work on a shared Login component.

pakokrew:
> Ok, you are right.
> May I make another PR of the update including web3 context?
> If you ok, I want you to approve this PR and merge.

Actually I don't really see the point of this PR. If this is just about sharing a "LoginButton component", which includes very few logic and that will be anyway different between apps, I am not sure it is necessary. 

I would highly prefer to work first on sharing the auth mechanism with the Web3 context and hooks, integrate them into MyMeta and MetaMaps, see how that fits, and then maybe work on a shared Login component.

ace-smart:
> I would highly prefer to work first on sharing the auth mechanism with the Web3 context and hooks, integrate them into MyMeta and MetaMaps, see how that fits, and then maybe work on a shared Login component.

I see

hammadj:
Agree with @pakokrew. I think making a specific package to extract the web3 context stuff would be the way to go, could even use the `@metafam/utils` package we have already. That way they can use those same top level components and logic.

ace-smart:
@pakokrew @hammadj @dysbulic I moved `Web3Context` stuff to the @metafam/utils packages to be shared from other packages.
Please review commits. :)

packages/utils/package.json
```diff
@@ -13,9 +13,12 @@
     "precommit": "yarn lint-staged"
   },
   "dependencies": {
+    "@walletconnect/web3-provider": "^1.2.1",
```
pakokrew, line 16: Remove the packages that are not needed anymore from `web`

pakokrew (approved):

pakokrew:
@hammadj could you authorise the deployment please :)

TheLoneRonin:
Hey @ace-smart!
Are there any other changes you'd like to this pull request?

If not, we can merge it now.

pakokrew:
LGTM we can merge

packages/web/components/PageHeader.tsx
ace-smart: It isn't necessary from now because `Web3Context` had been moved to `@metafam/utils`

ace-smart:
> Hey @ace-smart!
> Are there any other changes you'd like to this pull request?
> 
> If not, we can merge it now.

not, please merge this PR

Issue #278, opened 2021-01-15
[MyMeta] Profile Edit Page Design


META-DREAMER:
Currently editing a profile takes you through the whole onboarding flow. This isn't ideal since we want people to be able to edit individual parts of the profile. Also updating the username is the first thing in the flow and that's not something we want people to be doing often.

Would be nice to have a simpler "edit profile" page that let people update their profile info after creating it. It would make sense for each "card" in the profile to have its own edit button so when you click edit on your Skills card it shows a modal with just the skills editor. Likewise with the personality stuff, and for the profile image it should redirect to 3box.

Reusing the same components in the onboarding flow would be a requirement as well so we don't have to maintain multiple components for editing stuff.

alalonde:
Yes, I was thinking something similar. What I planned on implementing was just converting the onboarding flow such that each step had its own link (which it should regardless), and then linking to a step directly. Then you could just change the button to be "Save and Exit" or something of the like.

Seems a bit more lightweight implementation-wise, albeit probably not as good a UX.

dysbulic:
Profile storage is migrating to IDX (3Box profiles are getting retired), so this information will need to go to IDX after #309.

alalonde:
#333 gets us most of the way towards an implementation of this.

alalonde:
Once that's merged it's just a matter of adding links

peth-yursick:
Is this still missing the design @alalonde?

alalonde:
No, I don't think so. Perhaps you were referring to another issue?

luxumbra:
Is anything needing doing on this to get it pushed out the door?

alalonde:
No, I don't think there are any blockers at this point.

luxumbra:
> Once that's merged it's just a matter of adding links

If it needs finishing off, perhaps is a good first issue for someone? Where should the the links go?

The 'design needed' tag & 'Design' in the title suggests there's some design input required here?

davort:
Happy to help from the design side of things! Feel free to assign me to this issue. @luxumbra - will you be taking care of the code?

davort:
FYI - started designing the "Edit Profile" page: https://www.figma.com/file/zDHqledo5H2aKR3KDvbur8/Meta-Game?node-id=6952%3A68

dysbulic:
> FYI - started designing the "Edit Profile" page…

[...]

PR #333, opened 2021-02-08
Refactored profile edit flow to use browser history

alalonde:
This is purely a refactor with no new added functionality beyond back/forward browser navigation support.

This accomplishes several things:

- Makes linking to sections of the setup flow possible (#278, will need additional UI)
- Moves all step-specific logic out of the setupcontext and into a new step-specific page
- SetupOptions data structure was refactored to allow for much simpler computation of current step / section

I'm sure there's some performance improvements that could be made here, and probably some stuff I didn't refactor that I probably should, especially in the `SetupContext` component.


packages/web/components/LoginButton.tsx
```diff
@@ -45,7 +46,9 @@ export const LoginButton: React.FC = () => {
             {user.player ? getPlayerName(user.player) : 'Unknown'}
           </MetaLink>
           <HStack spacing={2}>
-            <MetaLink href="/profile/setup">Setup profile</MetaLink>
+            <MetaLink href="/profile/setup/username">
+              {user.username && user.username !== address ? 'Edit profile' : 'Setup profile'}
```
dysbulic, line 50: I'd use `user?.username` since `user` can be `null`.

packages/web/components/Setup/SetupHeader.tsx
```diff
@@ -7,20 +7,21 @@ import { useSetupFlow } from 'contexts/SetupContext';
 import React from 'react';
 
 export const SetupHeader: React.FC = () => {
-  const { step, screen, onNextPress, onBackPress, options } = useSetupFlow();
+  const { stepIndex, onNextPress, onBackPress, options } = useSetupFlow();
+
+  const {sectionIndex} = options.steps[stepIndex];
```
dysbulic, line 12: Spaces inside object…

dysbulic (commented):
I don't think your code breaks this, but if I setup a profile then go back to edit it, the first screen is populated with the username and I can click the "Next" button.

This isn't true for the next (Personality Type) screen. The type isn't selected and I can't click "Next".

packages/web/components/LoginButton.tsx
alalonde: no, there's a if `(!user?.player) return null;`above it, and a nullreference error would occur before here anyway.

alalonde:
> I don't think your code breaks this, but if I setup a profile then go back to edit it, the first screen is populated with the username and I can click the "Next" button.
> 
> This isn't true for the next (Personality Type) screen. The type isn't selected and I can't click "Next".

Let me check this and confirm there's not a regression. Thanks for looking over the PR.

alalonde:
> > I don't think your code breaks this, but if I setup a profile then go back to edit it, the first screen is populated with the username and I can click the "Next" button.
> > This isn't true for the next (Personality Type) screen. The type isn't selected and I can't click "Next".
> 
> Let me check this and confirm there's not a regression. Thanks for looking over the PR.

Nope, this doesn't currently work in prod currently. I'll create a separate PR to address it

pakokrew:
Will review that once #342 is merged and this is rebased :)

hammadj:
> Will review that once #342 is merged and this is rebased :)

Merged that, this can be rebased now @alalonde

alalonde:
Rebased and ready for review.

packages/web/components/Setup/SetupProfile.tsx
```diff
@@ -1,89 +1,16 @@
...
-    setPlayerType,
-    availability,
-    setAvailability,
-    timeZone,
-    setTimeZone,
-    skills,
-    setSkills,
-  ]);
+export const SetupProfile: React.FC = ({children}) => {
+  const {options, stepIndex} = useSetupFlow();
```
hammadj, line 7: 🙏🙏🙏🙏🙏🙏🙏🙏

packages/web/contexts/SetupContext.tsx
```diff
@@ -1,182 +1,76 @@
...
-  }>;
-};
+const urlPrefix = `/profile/setup/`;
 
 type SetupContextType = {
-  options: Array<SetupOption>;
-  step: number;
-  screen: number;
+  options: SetupOptions;
+  stepIndex: number;
```
hammadj, line 9: Alec "Refactoring God" LaLonde

hammadj (approved):
Much much cleaner / better, good stuff!

packages/web/components/LoginButton.tsx
hammadj: The type checker would've picked it up anyways if user could've been null here

pakokrew (approved):

dysbulic (approved):

dan13ram (approved):
This is awesome work @alalonde ! Thank you =)

packages/web/pages/profile/setup/timeZone.tsx
```diff
@@ -0,0 +1,38 @@
...
+export type DefaultSetupProps = InferGetStaticPropsType<typeof getStaticProps>;
+
+const TimeZoneSetup: React.FC<DefaultSetupProps> = () => {
+
+  const [timeZone, setTimeZone] = useState<string>('');
+  const { user } = useUser({ redirectTo: '/' });
+
+  if (user?.player) {
+    const {player} = user;
+    if (player.timezone && !timeZone) {
```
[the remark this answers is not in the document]
alalonde: I looked into this a bit. I think the best solution would be to refactor each `<SetupComponent>` to just take a `player` as a prop and thus figure out the value of the field, rather than the parent component having this logic. Since each setup is its own page, we shouldn't need `useState` anymore either, as the data will be retrieved from graphql on load.

Issue #290, opened 2021-01-18
[Quests] First iteration


pakokrew:
This is the spec for the first iteration of the **Quests** app, which will be split in different PR and will require work from different people.

Trying to follow https://www.notion.so/Quests-57662b138f6240d386924b5c5f076312

## Objective

We first try to create a simple system for replacing the #quests channel in discord. This will implement the Phase 0 we describe in Notion. 

- Only people holding at least 100 pSEED (seeds in liquidity pool) will be able to create quests
- Creation of a page to display available quests and see details
- Creation of a page to create a quest, only accessible by moderators
- Creation of a page to edit a quest, only accessible by quest creator
- Quests will not contain rewards. Expected work should be tracked by sourcecred and generate XP/seeds
- Quest creator can edit the quest and have the responsibility to accept submissions.

Quests can be done either:
- once (UNIQUE)
- once per player (PERSONAL)
- once per player every X days (RECURRING)

## Future work

As this is a first iteration, it will miss a lot of feature that will come in the future. 
What we may do after this:

- Create a plugin for SourceCred to distribute XP based on the quests that were completed if we want.
- Create a "role" in guild_player, to define permissions inside a guild, which would enable guild member to create quests
- Allow anyone member of a guild to create a quest, use (conviction/quadratic) voting to pass the quest from *proposal* to *open*
- Allow anyone to accept quest completed submission through voting
- Add a section in the player's profile to show completed quests
- Having Evergreen quests move on daily/weekly basis. They can be done only the day they are active
- Add skills needed to complete a quest
- Accomplishing this will result in being able to assert player's skills

### /quests

Page displaying all the existing quests.

Should display as a paginated vertical list.

Each line contains:
- quest.title
- quest.created_at
- quest.status
- quest.guild
- number of people that have submitted/completed the quest
- skills related to the quest

Filterable by:
- status
- skills
- guild
- created by me

Sortable by:
- created_at

### /quests/{quest_id}

Page containing all the details about a specific quest.

Should display every *quest* prop

Should display all player's submissions.

If the viewer is the quest creator, he can edit the quest or accept submissions.

If the viewer is a player and the quest is OPEN, he can send a submission

### /quests/{quest_id}/edit

Page to edit a quest. Only available to quest creator.

Should allow to edit quest's [title, description, external_link] props

### /quests/new

Creation of a quest. Only available to moderators.

Should ask for quest's [guild_id, title, description, external_link] props

Creator must be member of the guild

Creator must be 

[...]

PR #339, opened 2021-02-10
Quests models

pakokrew:
Implementing #290 

Creation of the base model for quests.


hasura/metadata/tables.yaml
```diff
@@ -343,6 +400,169 @@
...
+      - cooldown
+      - description
+      - external_link
+      - repetition
+      - status
+      - title
+      filter:
+        created_by_player_id:
+          _eq: X-Hasura-User-Id
+      check: null
```
alalonde, line 476: what is `check`?

alalonde (approved):
looks good as an initial pass

hasura/metadata/tables.yaml
pakokrew: I don't know, this is auto generated by hasura. Maybe it is the "post-update" check.

hammadj: Yep thats what it is

hammadj (approved):

PR #342, opened 2021-02-11
Persist personality type immediately rather than when saving player type

alalonde:


dysbulic (approved):
I have no qualms or even quibbles. I tested it and it works.

¿If you have a chance, could you look at #294? My header in Chromium and Chrome on Ubunutu for my.metagame.wtf looks like:

Screenshot from 2021-02-12 23-07-49

I did replace the words for profile configuration and disconnecting to icons b/c it was wide enough to go behind the drawer when using text:

Screenshot from 2021-02-12 23-09-36

I'll put it back if that's what's needed to get this request approved.

pakokrew (approved):

PR #346, opened 2021-02-16
Make username params lowercase to match those in the database

### LINKED ISSUE #345: [MyMeta] Profile page URLs are case-sensitive, should be case-agnostic
Currently profile page URLs are case-sensitive, which is not accessible or user-friendly at the very least, and could perhaps lead to bigger issues (e.g. duplicate usernames, just using different combinations of upper and lower case).

As an example, my profile page can be accessed on https://my.metagame.wtf/player/Davor

image

...while the all lower case variant of the same URL ( https://my.metagame.wtf/player/davor ) will return a 404 page

image

This should be fixed so that even all-caps or mixed-case variants of the same URL return the same result – the same profile page, provided that profile exists.

alalonde:
Fixes #345


c0mput3rxz (commented):
You may need to ensure user name is set to a lowercase value on mutation to ensure linkage.

/packages/web/components/Setup/SetupUsername.tsx
<Input
        background="dark"
        placeholder="USERNAME"
        value={username}
        onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
          setUsername(e.target.value)
        }
        isInvalid={invalid}
      />

packages/web/pages/player/[username].tsx
```diff
@@ -180,12 +180,14 @@ export const getStaticPaths: GetStaticPaths<QueryParams> = async () => {
 export const getStaticProps = async (
   context: GetStaticPropsContext<QueryParams>,
 ) => {
-  const username = context.params?.username;
+  // usernames in the DB are lowercase. This essentially 
+  // makes this page's router variable case-insensitive.
+  const username = context.params?.username?.toLowerCase();
   const player = await getPlayer(username);
```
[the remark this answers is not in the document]
Comment (pakokrew) [reply packages/web/pages/player/[username].tsx:186]: When does this generation happens ? Curious to know where nextjs is aware of the list of players.

Maybe change to `const player = await getPlayer(username) || null;`

And also,  if `username`  is nullish, it is not necessary to call `getPlayer`

pakokrew (changes requested):
Still a problem when players havent set username

packages/web/pages/player/[username].tsx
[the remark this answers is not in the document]
Comment (alalonde) [reply packages/web/pages/player/[username].tsx:186]: > When does this generation happens ? Curious to know where nextjs is aware of the list of players.

This is part of next.js's [static rendering](https://nextjs.org/docs/basic-features/pages#static-generation-with-data). At build time, the list of players is fetched. 

> 
> I don't really understand the error, as by reading the code, the player props could not be `undefined` but `null`.

Yeah, I force-pushed a fix for that particular problem.

I changed the player username lookup to be case-insensitive. I believe this will fix the problem.  

Confusingly, in my local DB all the usernames are lowercase as they were all imported via `migrateSourceCredAccounts`, but on prod they are mixed-case.

alalonde:
Thinking about this some more, I believe the proper solution is to ensure usernames are lowercase. Currently the `unique` column constraint on username is case-sensitive, so you can end up with two players with e.g. usernames `Alec` and `alec` which is certainly not what we want. Solution will involve:

- Write a data migration to lower-case all existing usernames
- Automatically lower-casing new usernames entered by a user
- lowercase the url param for username on the `[username]` route.

hammadj:
> Thinking about this some more, I believe the proper solution is to ensure usernames are lowercase. Currently the `unique` column constraint on username is case-sensitive, so you can end up with two players with e.g. usernames `Alec` and `alec` which is certainly not what we want. Solution will involve:
> 
> * Write a data migration to lower-case all existing usernames
> * Automatically lower-casing new usernames entered by a user
> * lowercase the url param for username on the `[username]` route.

In this case, should we mark this PR as blocked until we do the data migration for lower-casing existing names? The proper way to do it would be to create a SQL migration file to do it so that everyones local DBs get updated when they pull the latest changes. Otherwise I can just do it manually directly on the prod instance. Lmk either way

pakokrew:
> In this case, should we mark this PR as blocked until we do the data migration for lower-casing existing names? The proper way to do it would be to create a SQL migration file to do it so that everyones local DBs get updated when they pull the latest changes. Otherwise I can just do it manually directly on the prod instance. Lmk either way

Let's write a migration :)

packages/web/pages/player/[username].tsx
```diff
@@ -180,12 +180,17 @@ export const getStaticPaths: GetStaticPaths<QueryParams> = async () => {
 export const getStaticProps = async (
   context: GetStaticPropsContext<QueryParams>,
 ) => {
-  const username = context.params?.username;
-  const player = await getPlayer(username);
-
+  const username = context.params?.username?.toLowerCase();
```
Comment (pakokrew) [inline packages/web/pages/player/[username].tsx:183]: Maybe it would be great to redirect to the page with the username correctly lowercase, so that if the user share the URL it is the good one.

https://maxschmitt.me/posts/next-js-redirects/

`if (context.params?.username !== context.params?.username?.toLowerCase();) `

pakokrew (changes requested):
The regex check is only present on the frontend, so if we manually call the graphql endpoint with a bad formated username, it will be accepted. 

The check should be done on the backend side. I'm not sure hasura can natively do that, so maybe we should have an handler for that.

packages/web/pages/player/[username].tsx
Comment (alalonde) [reply packages/web/pages/player/[username].tsx:183]: > The regex check is only present on the frontend, so if we manually call the graphql endpoint with a bad formated username, it will be accepted.
> 
> The check should be done on the backend side. I'm not sure hasura can natively do that, so maybe we should have an handler for that.

Yup, agreed; I spent about an hour playing around with regex / custom checks in postgres / hasura before throwing up my hands. I have another idea that should be cleaner.

Also agree about the redirect; it seems we should be somehow communicating to the user what the correct link is.

Comment (alalonde) [reply packages/web/pages/player/[username].tsx]: I implemented both changes; have a look @pakokrew

pakokrew (approved):
Well done ;)

dan13ram (approved):
LGTM. Lets get this merged! 

Not sure why build is failing on Vercel. But might have something to do with backend and frontend changes being in the same PR. 

Can you raise a separate PR with just backend changes, which we can merge first before rebasing and merging this PR?

alalonde:
I believe the deployment is failing because I added a regex check in postgres on the username field. We likely have existing usernames that violate the regex.

PR #354, opened 2021-02-18
Switch MyMeta over to 7/7/7/14/21 rank distributions

alalonde:
… with bonus tests! Resolves #324

Issue #341, opened 2021-02-10
[Bot] Automate role assignments using discord bot


hammadj:
- setup trigger in Hasura on the role field in the Player table so that whenever it changes a function is called that uses the discord bot client to update that Players role in discord (diamond, platinum, etc)

alalonde:
[Started working](https://github.com/MetaFam/TheGame/tree/feature/auto-update-discord-roles) on this today. I created the hasura trigger, however, we will need to at least partially implement #184 as we need a mapping between our ranks and a discord Role ID. I'll jump over to that one then. Probably quite a bit of refactoring involved there.

Issue #184, opened 2020-11-18
[MEGA ISSUE] Discord as source of truth for guilds


META-DREAMER:
### Context

Currently we use Moloch / Aragon as a means for discovering DAOs and other external guilds.  There is a Daohaus API for exploring these DAOs, however, any given guild / DAO must use Moloch / Aragon for this API to know about them.  We have observed that almost every organization who has expressed interest in becoming a Metagame guild has their own discord server, however.  And that discord provides a robust API that can be leveraged to easily onboard guilds.

### Design

We will create a Discord Bot that can be used to both onboard guilds and provide a way of interacting with MetaGame.  

- MyMeta "Join as guild" page: If a guild expresses interest in joining MetaGame, they can navigate to a "Join as a guild" page within MyMeta. This will provide some introductory text about MetaGame as well as a [Discord auth URL](https://discordjs.guide/preparations/adding-your-bot-to-servers.html#bot-invite-links) that can be clicked to add the bot to their Discord server.  
- Once accepting the various permissions required by the bot, the user will be redirected back to MyMeta, at which time we will attempt to fetch the necessary metadata to create the Guild in MetaGame. 
  - We can match existing guilds by manually add a `discord_id` to the existing guilds in our DB
  - Initial data to collect and allow editing: 
    - `guildname` (basically a username for a guild)
    - `description`
    - `discord_invite_url` (just the code, we will prepend "https://discord.gg/" to the input field
    - `join_button_url`
    - `logo` (fetch the image and display, allow editing the URL. Editing will display the new image)
    - `moloch_address` ?
    - `name`
    - `type` enum: `FUNDING` | `SOCIAL` | `SERVICE` | `PROJECT` | `RESEARCH`
    - `website_url`
  - We will also fetch all the roles for the discord server and populate a multi-select widget. This will allow the user to select the roles that indicate membership in the guild. We will also ask for roles that indicate some sort of administrator / owner status.
- On submission, add the guild + metadata. Save the discord server ID on the guild table and discord membership role IDs in a new JSONB column. Also, query the list of members in the discord server, filter by whatever roles count as members, match the member's discord ID with our `player.discord_id` and insert the match into the GuildPlayer table.  Only add discord guild members that already have player accounts in MetaGame.

### Prerequisites

- Refactor the DB to better match this deeper Discord integration. Move discord_id from player_account / discord to main player table, as well as adding a discord_id to the guild table. Migrate all the data along the way.
[...]

PR #355, opened 2021-02-19
Auto-update roles in discord when they're changed in the DB

alalonde:
Takes care of #341. Note that this piggy-backs off of #354, so that should go first.

### DIFF
--- a/packages/backend/src/handlers/triggers/updateDiscordRole.ts
+++ b/packages/backend/src/handlers/triggers/updateDiscordRole.ts
@@ -1,6 +1,8 @@
-import fetch from 'node-fetch';
+import Discord from 'discord.js';
 
+import { CONFIG } from '../../config';
 import { AccountType_Enum, Player } from '../../lib/autogen/hasura-sdk';
+import { client } from '../../lib/hasuraClient';
 import { TriggerPayload } from './types';
 
 export interface UpdateRole {
@@ -27,10 +29,16 @@ export const updateDiscordRole = async (
     previousRole: oldPlayer?.rank?.toString(),
     newRole: newRank
   }
-  // look up guild by guildname = 'metagame' (for now), 
-  // then look up guild_account for DISCORD type
-  // will need to create new graphql queries for these (can they be joined into one?)
 
+  // look up guild by guildname = 'metagame' (for now), 
+  const metafam = await client.GetGuild({guildname: 'metafam'});
+
+  const discordClient = new Discord.Client();
+  discordClient.login(CONFIG.discordBotToken);
+
+  // todo add jsonb field to guild. { ranks: [] }
+  // populate also.
+  
 
   // call discord API. for each query we'll need serverId, playerId, and roleId
   // since we don't have roleIds, we'll need to look up the roles in the server 
@@ -39,5 +47,4 @@ export const updateDiscordRole = async (
 
   // if there's an oldRank, delete it.
   // if there's a new rank, add it 
-  fetch()
 };
--- a/packages/backend/src/handlers/triggers/updateDiscordRole.ts
+++ b/packages/backend/src/handlers/triggers/updateDiscordRole.ts
@@ -1,7 +1,8 @@
-import Discord from 'discord.js';
+/* eslint-disable no-console */
+import Discord, { Role } from 'discord.js';
 
 import { CONFIG } from '../../config';
-import { AccountType_Enum, Player } from '../../lib/autogen/hasura-sdk';
+import { AccountType_Enum, Player, PlayerRank_Enum } from '../../lib/autogen/hasura-sdk';
 import { client } from '../../lib/hasuraClient';
 import { TriggerPayload } from './types';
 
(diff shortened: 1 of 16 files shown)


hammadj:
>  Note that this piggy-backs off of #354, so that should go first.

If a PR is piggy backing off another, select that other PRs branch as the base for this PR. When that other PR gets merged, GitHub will automatically update the base of this to be on master. This way it makes it easy to see only the changes this PR makes

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (newRank == null) return;
+   
+    // look up guild by guildname = 'metagame' (for now), 
+    const getGuildResponse = await client.GetGuild({guildname: 'metafam'});
+    const discordAccount = getGuildResponse.guild[0]?.guild_accounts?.find(a => a.type === AccountType_Enum.Discord);
+    if (discordAccount == null) return;
+
+    // call discord API. We'll need serverId, playerId, and roleIds
+    const discordClient = new Discord.Client();
+    discordClient.login(CONFIG.discordBotToken);
```
hammadj, line 38: Should probably instantiate discord client in a separate file since we will likely re-use it. Also, discord.js dependency already exists in the `discord-bot` package, would be ideal to import it from there (might need to create a `createClient` export from there though, right now its just being run standalone but the idea is to be able to import it in other packages and re-use it)

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // todo add jsonb field to guild and populate ranks, e.g. { ranks: [] }
+    const guild = await discordClient.guilds.fetch(discordAccount.identifier);
```
hammadj, line 41: `discordAccount` makes it sound like it's a user account for discord, but seems like this is actually a `discordGuild`.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (guild == null) {
```
hammadj, line 42: Why not just
```suggestion
    if (!guild) {
```

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    // look up guild by guildname = 'metagame' (for now), 
+    const getGuildResponse = await client.GetGuild({guildname: 'metafam'});
+    const discordAccount = getGuildResponse.guild[0]?.guild_accounts?.find(a => a.type === AccountType_Enum.Discord);
+    if (discordAccount == null) return;
+
+    // call discord API. We'll need serverId, playerId, and roleIds
+    const discordClient = new Discord.Client();
+    discordClient.login(CONFIG.discordBotToken);
+
+    // todo add jsonb field to guild and populate ranks, e.g. { ranks: [] }
```
hammadj, line 40: I want to make "discord roles as a source of truth for guild membership" and display all the roles a player has in each guild on their profiles, so the idea of "ranks" can just be subsumed by the "roles" (with the difference being we have specific logic to update ranks automatically based on XP and update roles in discord, whereas the other roles would just be synced from discord into our DB).

However I know that trying to keep two sources of truth in sync (DB and what role is set in discord) is never a good time, so was thinking we could even fetch the roles directly from discord API and remote join them to the Player table. Happy to do syncing for now but I can imagine it causing issues in the future

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    // since we don't have roleIds, we'll need to look up the roles in the server 
+    // and just match by strings for the time being.
+    // https://discord.com/developers/docs/resources/guild#get-guild-roles
+    const rankDiscordRoles = new Map<PlayerRank_Enum, Role>();
+
+    const discordRoleManager = await guild.roles.fetch();
+    discordRoleManager.cache.forEach(discordRole => {
+      const rankMatch = Object.values(PlayerRank_Enum).find(rank => {
+        return rank.toUpperCase() === discordRole.name.toUpperCase();
+      });
```
hammadj, line 56: This seems pretty fragile, should create a TODO / issue here so we don't forget about it. We should be storing the "recognized roles" for each guild under the guild tables and moving the `rank` field to be on the `guild_player` table. TBH it shouldn't even be called rank anymore, just a list of roles on the guild_player (as mentioned in comment above)

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    // and just match by strings for the time being.
+    // https://discord.com/developers/docs/resources/guild#get-guild-roles
+    const rankDiscordRoles = new Map<PlayerRank_Enum, Role>();
+
+    const discordRoleManager = await guild.roles.fetch();
+    discordRoleManager.cache.forEach(discordRole => {
+      const rankMatch = Object.values(PlayerRank_Enum).find(rank => {
+        return rank.toUpperCase() === discordRole.name.toUpperCase();
+      }); 
+      if (rankMatch != null) {
```
hammadj, line 57: Why not
```suggestion
      if (rankMatch) {
```

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+        rankDiscordRoles.set(rankMatch, discordRole);
+      }
+    })
+
+    const discordPlayer = await guild.members.fetch(discordPlayerAccount.identifier);
+    if (discordPlayer == null) {
```
hammadj, line 63: Why not
```suggestion
    if (!discordPlayer) {
```

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  
+    // Add the new rank.
+    const discordRoleForRank = rankDiscordRoles.get(newRank);
+    if (discordRoleForRank == null) {
+      console.warn(`discord role associated with ${newRank} was not found!`);
+    } else {
+      await discordPlayer.roles.add(discordRoleForRank);
+      console.log(`Added role ${newRank} for player ${newPlayer?.username} in discord`);
+    }
```
hammadj, line 88: Would be nice to also send a message in one of the discord channels to notify the user with something like "@username has levelled up to Gold rank!" or "@username has moved down to Silver rank!"

hammadj (commented):
Few questions / suggestions

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: okay, yeah, this was my original thought too, but it was a bit challenging to get an instantiated (logged-in) version of the client exported and then imported elsewhere.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: I prefer `==` null as it is more explicit and typesafe. `!false` is true, but `false == null` is still false.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: Yes, I was going to tackle that in #184. At that point this will probably get refactored.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: Yup, I'll take care of this in #184.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: Yeah, for sure. I would probably tackle this in a separate PR

alalonde:
Just pushed a commit that imports the discord client from the `discord-bot` package. Fair amount of legwork to get that working locally and in docker. To deploy, we just need to include the bot token in the config and give that bot `manage-roles` permission in the Metafam discord.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: renamed this.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: implemented this in the last few commits.

hammadj (approved):

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,92 @@
...
+  payload: TriggerPayload<Player>,
+) => {
+  const {old: oldPlayer, new: newPlayer} = payload.event.data;
+
+  try {
+    if (newPlayer == null) return;
+   
+    const getPlayerResponse = await client.GetPlayer({playerId: newPlayer.id});
+    const discordPlayerAccount = getPlayerResponse.player_by_pk?.Accounts?.find(a => a.type === AccountType_Enum.Discord);
+    if (discordPlayerAccount?.identifier == null) return;
```
dysbulic, line 25: ¿Should this also return if `undefined`?

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const newRank = newPlayer?.rank;
+
+    if (newRank == null) return;
+   
+    // look up guild by guildname = 'metagame' (for now), 
+    const getGuildResponse = await client.GetGuild({guildname: 'metafam'});
+    const discordGuildAccount = getGuildResponse.guild[0]?.guild_accounts?.find(a => a.type === AccountType_Enum.Discord);
+    if (discordGuildAccount == null) return;
```
dysbulic, line 34: ¿Should this return if `discordGuildAccount` is `undefined`?

packages/backend/src/handlers/triggers/updateDiscordRole.ts
dysbulic: * `undefined == null` → `true`
* `undefined === null` → `false`

It's safest to use `===` as often as possible.

dysbulic (approved):
Everything seems sensical to me.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
alalonde: >     * `undefined == null` → `true`
> 
>     * `undefined === null` → `false`
> 
> 
> It's safest to use `===` as often as possible.

Yes, so if `guild` is `undefined` or `null`, return. In my experience, one usually doesn't need to worry about the distinction.

PR #360, opened 2021-02-22
Fix mobile connect

### LINKED ISSUE #338: [MyMeta] Error reconnecting with Metamask after an initial connection with WalletConnect
> [Davor]: [I'm] connecting with MetaMask and this screen pops up whenever I load a new page on my.metagame.wtf
> 
> 
> 
> But can't seem to get past it. … I hit the "Sign" button, but then nothing happened after that ☹
> 
> [The error console when I attempt to sign is:]
>
> Screenshot_20210209
>
> I remembered that the first time I tried this, I used the Argent app, FWIW. But even then, I didn't get any feedback from the website whatsoever

It looks to me like Web3Modal has cached WalletConnect and is attempting to use it. Why the Metamask dialog is coming up, I don't know.

Adding `modal.clearCachedProvider()` to the `Disconnect` action should be done.

This doesn't fix Davor's issue since he can't connect.

pakokrew:
Trying to fix #338 

Need someone to try it out

### DIFF
--- a/packages/utils/src/did/index.ts
+++ b/packages/utils/src/did/index.ts
@@ -2,6 +2,8 @@ import { providers, utils } from 'ethers';
 import { Base64 } from 'js-base64';
 import { v4 as uuidv4 } from 'uuid';
 
+import { signerHelper } from '../ethereumHelper';
+
 const tokenDuration = 1000 * 60 * 60 * 24 * 7; // 7 days
 
 type Claim = {
@@ -29,7 +31,7 @@ export async function createToken(
   };
 
   const serializedClaim = JSON.stringify(claim);
-  const proof = await signer.signMessage(serializedClaim);
+  const proof = await signerHelper(provider, serializedClaim);
 
   return Base64.encode(JSON.stringify([proof, serializedClaim]));
 }
--- a/packages/utils/src/ethereumHelper.ts
+++ b/packages/utils/src/ethereumHelper.ts
@@ -1,0 +1,21 @@
+import { providers } from 'ethers';
+
+export async function signerHelper(
+  provider: providers.Web3Provider,
+  rawMessage: string,
+) {
+  const ethereum = provider.provider;
+  const signer = provider.getSigner();
+  const address = await signer.getAddress();
+  if (!ethereum.request) throw new Error('invalid ethereum provider');
+
+  let params = [rawMessage, address.toLowerCase()];
+  if (ethereum.isMetaMask) {
+    params = [params[1], params[0]];
+  }
+  const signature = await ethereum.request({
+    method: 'personal_sign',
+    params,
+  });
+  return signature;
+}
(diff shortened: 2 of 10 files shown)


dysbulic:
The profile isn't showing up when using WalletConnect some of the time.


https://user-images.githubusercontent.com/181523/108678628-52b07580-74b9-11eb-9abe-0e36f7a2efa5.mp4

pakokrew:
> The profile isn't showing up when using WalletConnect some of the time.> 

Thanks for trying it.
I've improved it a little bit, what do you see now ?

dysbulic:
I'm seeing quite a few different things that I've got no good explanation for (other than caching somewhere):

https://user-images.githubusercontent.com/181523/109064055-e6e82b80-76b7-11eb-8eef-a88cf8034680.mp4

pakokrew:
> I'm seeing quite a few different things that I've got no good explanation for (other than caching somewhere):
> 
>  2021-02-24.15-45-30.mp4

The first problem you mention here was already present, and due to the fact that we do not handle account change on the provider. This PR was about fixing connecting on mobile phones browsers. If you don't mind I would prefer open another issue&pr.

For the second one, I'll investigate...

dysbulic:
> For the second one, […] I can't reproduce the error, even when trying with account not registered yet. Nothing in the console?

I wish I had something productive to report. Under Ubuntu in both Chrome and Firefox, I open a new incognito window – which should have no saved settings – and the profile disappears after I authenticate and sign the wallet connect message.

The only thing in the console is "`wallet connect`".

dan13ram:
Hey @pakokrew I did a round of testing and still find issues when disconnecting and connecting with a new account.
https://www.loom.com/share/d90b0f7885ba4ffca365f9ec8e4b09cf

dysbulic:
I ran upon this issue trying to test out some mobile fixes to the site. I spent a few hours debugging and my guess is that one of the main issues has to do with signature verification via WalletConnect. [All of the wallets seems to be broken in one way or another.](https://github.com/WalletConnect/walletconnect-monorepo/issues/347#issuecomment-798782047)

I couldn't get a profile to display using Argent which can't generate valid signatures, but things worked with Trust Wallet easily enough.

I suspect the problem is not being able to validate the `authToken`.

pakokrew:
Just seeing your comments now.

There are two major distincts issues here:
- Handling when user changes account in wallet
- Signing with contract-based wallets such as Argent

I'll try to answer these two issues separately

pakokrew:
Just found that the issue of connect/disconnect/reconnect with another account is not coming from this Web3Context, but from the urql setup we have. 
We need to rewrite how we manage the authentication token inside `withUrql` in _app

If you don't mind, I'd prefer merge this @dan13ram @dysbulic then work separately on handling account change from provider and better setting up urql. There's already an issue for the first on, I'll create another one for the send.

pakokrew:
Ok I found a quick fix that resolves the issue with URQL when explicitely disconnect and reconnect with another account.

dan13ram (approved):
As you said, lets merge this so we can fix the address issue in a separate PR.

Looks like there is a lint issue, though. Please fix that before merging!

PR #361, opened 2021-02-22
setup brightId verified labels

dan13ram:


pakokrew (changes requested):
I don't understand how you plan to store brightid verification in the backend.
I believe what you did in the frontend trustlessly works, but I don't see what will change the information in the backend.

What we did before for github and twitter, is that the backend will check for the verifications (from 3box for these, in `updateBoxProfile` action), and then add PlayerAccount roles with the corresponding informations. Then the frontend just reads the player=>accounts values to know the verified accounts. 

Would you mind using the same logic, make the backend actions verify brightid in addition to 3box, and add a row in Player_Account ?

packages/web/graphql/mutations/updateBrightIDStatus.ts
```diff
@@ -0,0 +1,12 @@
+import gql from 'fake-tag';
+
+export const UpdateBrightIDVerifiedMutation = gql`
+  mutation UpdatePlayerBrightIDVerified($verified: Boolean!) {
```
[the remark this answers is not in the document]
dan13ram: this is changing completely and won't be necessary

dan13ram:
> I don't understand how you plan to store brightid verification in the backend.
> I believe what you did in the frontend trustlessly works, but I don't see what will change the information in the backend.
> 
> What we did before for github and twitter, is that the backend will check for the verifications (from 3box for these, in `updateBoxProfile` action), and then add PlayerAccount roles with the corresponding informations. Then the frontend just reads the player=>accounts values to know the verified accounts.
> 
> Would you mind using the same logic, make the backend actions verify brightid in addition to 3box, and add a row in Player_Account ?

please check https://github.com/MetaFam/TheGame/pull/362. I've updated the logic completelt

packages/web/graphql/mutations/updateBrightIDStatus.ts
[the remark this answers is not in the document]
dan13ram: as mentioned in the labels this is a WIP. I'll change the logic and update this PR today. Thanks for taking a look! :)

pakokrew:
> please check #362. I've updated the logic completely

My bad, didn't see the WIP label 🤦‍♂️ ...
So you will not store anything in the backend anymore ?

dan13ram:
> > please check #362. I've updated the logic completely
> 
> My bad, didn't see the WIP label  ...
> So you will not store anything in the backend anymore ?

Yeah as you can see in the other PR #362 I've added a remote schema relationship which queries BrightID apis to check for status, which is stored against the player => `brightid_status`.
This will be used by the frontend to display verified status.

In case a player is not verified, then we shall show a QR for the user to scan using their brightID app and verify.

Then the user must **refresh** the page for the status to get updated. 

Please let me know if there can be a better implementation here!

pakokrew:
> > > please check #362. I've updated the logic completely
> > 
> > 
> > My bad, didn't see the WIP label  ...
> > So you will not store anything in the backend anymore ?
> 
> Yeah as you can see in the other PR #362 I've added a remote schema relationship which queries BrightID apis to check for status, which is stored against the player => `brightid_status`.
> This will be used by the frontend to display verified status.
> 
> In case a player is not verified, then we shall show a QR for the user to scan using their brightID app and verify.
> 
> Then the user must **refresh** the page for the status to get updated.
> 
> Please let me know if there can be a better implementation here!

I believe that's good, the only issue I see here is that hasura will have to fetch brightid endpont each time we fetch user data, which can be slow. 
We were thinking about having some caching for remote schema but that's not for today.
About having to refresh, can't nextjs allow us to invalidate the initial props and fetch them again?

dan13ram:
> I believe that's good, the only issue I see here is that hasura will have to fetch brightid endpont each time we fetch user data, which can be slow.
> We were thinking about having some caching for remote schema but that's not for today.

Sounds good. We can then look at that as a separate issue

> About having to refresh, can't nextjs allow us to invalidate the initial props and fetch them again?

Yes its possible to do that. But will require an action from the user. Do you think we can have a refresh button somewhere on the profile for re-fetching user's data?

pakokrew:
> > I believe that's good, the only issue I see here is that hasura will have to fetch brightid endpont each time we fetch user data, which can be slow.
> > We were thinking about having some caching for remote schema but that's not for today.
> 
> Sounds good. We can then look at that as a separate issue
> 
> > About having to refresh, can't nextjs allow us to invalidate the initial props and fetch them again?
> 
> Yes its possible to do that. But will require an action from the user. Do you think we can have a refresh button somewhere on the profile for re-fetching user's data?

Maybe while the modal with the QRCode is open we can poll brightID to see if the profile have been validated, if yes, re-fetch.

dan13ram:
> Maybe while the modal with the QRCode is open we can poll brightID to see if the profile have been validated, if yes, re-fetch.
My bad. It is possible to refetch only for queries made on the frontend. The /player/[username] pages are statically generated on the backend. And there isn't a way currently to re-generate them from the frontend. It just automatically gets updated every 1 sec. 
See https://github.com/MetaFam/TheGame/blob/b0ab6f1e2d072f9a4746ca2124e8cee1a67edca0/packages/web/pages/player/%5Busername%5D.tsx#L190

packages/web/components/Player/PlayerTile.tsx
```diff
@@ -28,145 +30,153 @@ type Props = {
...
+            <Avatar
+              size="xl"
+              src={getPlayerImage(player)}
+              name={getPlayerName(player)}
+            />
+            <Flex>
+              <Heading size="xs" color="white">
+                {getPlayerName(player)}
+              </Heading>
+              {verified && (
```
dan13ram, line 76: this line is the only change in this file along with useBrightIdStatus hook.

packages/web/lib/hooks/brightId.ts
```diff
@@ -0,0 +1,82 @@
...
+export const useBrightIdStatus = ({
+  player,
+}: {
+  player: PlayerFragmentFragment;
+}): {
+  verified: boolean;
+  deeplink: string;
+  universalLink: string;
+} => {
+  const verified =
```
pakokrew, line 20: Maybe you can wrap these 3 variables in an object that you compute with `useMemo` then return.

packages/web/lib/hooks/brightId.ts
```diff
@@ -0,0 +1,82 @@
...
+  const contextId = player.id;
+
+  useEffect(() => {
+    if (!contextId || !poll) return () => undefined;
+
+    let isSubscribed = true;
+
+    const update = () => {
+      fetchVerificationData(contextId).then((status) => {
+        const isVerified =
```
pakokrew, line 66: Maybe factorize this logic of computing `verified` that is used two times in this file (l.20)

packages/web/config.ts
```diff
@@ -4,4 +4,8 @@ export const CONFIG = {
   infuraId:
     process.env.NEXT_PUBLIC_INFURA_ID || '781d8466252d47508e177b8637b1c2fd',
   openseaApiKey: process.env.NEXT_OPENSEA_API_KEY || undefined,
+  brightIdAppUrl:
+    process.env.NEXT_BRIGHTID_APP_URL || 'https://app.brightid.org',
```
pakokrew, line 8: add to .env.sample

packages/backend/src/config.ts
```diff
@@ -46,8 +46,8 @@ export const CONFIG: IConfig = {
     process.env.NEXT_PUBLIC_INFURA_ID,
     '781d8466252d47508e177b8637b1c2fd',
   ),
-  brightIdNodeUrl: parseEnv(
-    process.env.BRIGHTID_NODE_URL,
+  brightIdAppUrl: parseEnv(
+    process.env.NEXT_BRIGHTID_APP_URL,
```
pakokrew, line 50: add to .env.sample

pakokrew (approved):
Minor changes (especially `useMemo` !) but LGTM

PR #362, opened 2021-02-23
updated backend for brightId

dan13ram:


packages/backend/src/handlers/remote-schemas/typeDefs.ts
```diff
@@ -1,9 +1,19 @@
 import { gql } from 'graphql-request/dist';
 
 export const typeDefs = gql`
+  scalar uuid
+
   type Query {
     getBoxProfile(address: String): BoxProfile
     getDaoHausMemberships(memberAddress: String): [Member!]!
+    getBrightIdStatus(contextId: uuid): BrightIdStatus
```
dan13ram, line 9: Without this Hasura was giving me an error when creating a remote relationship.
See https://github.com/hasura/graphql-engine/issues/5482

packages/backend/src/handlers/remote-schemas/typeDefs.ts
```diff
@@ -1,9 +1,19 @@
 import { gql } from 'graphql-request/dist';
 
 export const typeDefs = gql`
+  scalar uuid
```
dan13ram, line 4: Without this Hasura was giving me an error when creating a remote relationship.
See hasura/graphql-engine#5482

packages/backend/src/config.ts
```diff
@@ -45,4 +47,9 @@ export const CONFIG: IConfig = {
     process.env.NEXT_PUBLIC_INFURA_ID,
     '781d8466252d47508e177b8637b1c2fd',
   ),
+  brightIdContext: parseEnv(process.env.BRIGHTID_CONTEXT, 'MetaGame'),
```
alalonde, line 50: Does this need to be in the config? Meaning, is it likely to change?

packages/backend/src/handlers/remote-schemas/types/uuid.ts
```diff
@@ -0,0 +1,38 @@
+import { GraphQLScalarType } from 'graphql';
+import { Kind } from 'graphql/language';
+
+const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+const nilUUID = '00000000-0000-0000-0000-000000000000';
```
alalonde, line 5: When would this be the case? that this is the value

alalonde (approved):
Nice! Looks pretty straightforward

packages/backend/src/config.ts
dan13ram: it is not likely to change. I suppose it could be set as a constant.

packages/backend/src/handlers/remote-schemas/types/uuid.ts
dan13ram: this is just the nilUUID which is also part of the UUID space but not covered by the regex.

PR #365, opened 2021-02-25
Fixed username updating

### LINKED ISSUE #364: [MyMeta] Can't update username

alalonde:
Resolves #364. Also confirmed that you can't update somebody else's username (at least using that graphql mutation)

### DIFF
--- a/packages/web/graphql/mutations/updateUsername.ts
+++ b/packages/web/graphql/mutations/updateUsername.ts
@@ -1,12 +1,10 @@
 import gql from 'fake-tag';
 
 export const UpdateUsernameMutation = gql`
-  mutation UpdatePlayerUsername($username: String!) {
-    update_player(_set: { username: $username }, where: {}) {
-      returning {
-        id
-        username
-      }
+  mutation UpdatePlayerUsername($playerId: uuid!, $username: String!) {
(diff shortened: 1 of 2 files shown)


packages/web/graphql/mutations/updateUsername.ts
```diff
@@ -1,12 +1,10 @@
 import gql from 'fake-tag';
 
 export const UpdateUsernameMutation = gql`
-  mutation UpdatePlayerUsername($username: String!) {
-    update_player(_set: { username: $username }, where: {}) {
-      returning {
-        id
-        username
-      }
+  mutation UpdatePlayerUsername($playerId: uuid!, $username: String!) {
```
dan13ram, line 4: what error was the previous mutation causing ?

The backend is setup that only the player can update the player data information, which is checked from the header `X-Hasura-User-Id`

In case the backend is insecure => that is, if one player is able to update another players' data, then we need to update the permissions on the backend.

dan13ram (approved):
LGTM. Perhaps this needs to be solved on the backend though

packages/web/graphql/mutations/updateUsername.ts
alalonde: It didn't work at all and just failed silently. I suspect it's because there is that hasura permission you mentioned. I tried to update another player's username in my testing and its not possible.

PR #380, opened 2021-03-04
added token balance from seed-graph

dan13ram:


pakokrew (approved):
Neat. Need to use this in the backend handler for creating quests.

dan13ram:
> Neat. Need to use this in the backend handler for creating quests.

Sure. Can do that in a separate PR

PR #387, opened 2021-03-06
[MyMeta] Profile completion using the keyboard

dysbulic:
The input elements (and the accompanying changes) are:

1. Username Entry
  * enter keys saves the response and progresses to the next screen
2. Color Combination Dispositions
  * added focus and blur listeners to the "buttons"
  * added `tabIndex={0}` to the "buttons"
  * added a space and enter key listener to toggle the selection of the currently focused element
3. Player Type
  * the same basic feel as for Color Choices
4. Skills
  * enabled selection color in drop down so you can see how the keys move the selection
  * expanded the size of the dropdown where possible
5. Availability
  * fixed a bug where you couldn't completely clear the input of data
  * submits on enter press
6. Time Zone
  * enabled current selection coloring
  * enabled focused element coloring
7. Memberships
  * I don't have any, and didn't mock up any data.

I also put soft hyphens in several of the titles. They won't display normally, but if the screen is small enough the word won't fit, that's where it will hyphenate.


dysbulic:
This work is now a part of #401, so I'm closing this request.

PR #395, opened 2021-03-09
[MyMeta] Patrons leaderboard

dan13ram:
@pakokrew @peth-yursick I've pushed the latest version here. 

You can find the patrons leaderboard at https://the-game-n2zid85zf-metafam.vercel.app/patrons
We should probably add a link to the Patron Leaderboard on the NavBar. in a separate PR. 

Please let me know if I need to add/remove/edit the PatronTile.


packages/web/graphql/getPatrons.ts
```diff
@@ -0,0 +1,78 @@
...
+  // eslint-disable-next-line no-await-in-loop
+    const offsettedPatrons = await getPatronsWithOffset(page * 50);
+    patrons = offsettedPatrons.concat(patrons);
+    if (offsettedPatrons.length < 50) break;
+    page += 1;
+  }
+
+  return patrons
+    .filter((p) => BigNumber.from(p.token_balance?.pSeedBalance || 0).gt(0))
+    .sort(pSeedDescSort)
```
dan13ram, line 76: I could not do all of this on the backend. Do let me know if there is a way we can achieve this in hasura itself. 

Having said that these calls will be made only during static site generation so I think it should be fine.

packages/web/components/Patron/PatronTile.tsx
```diff
@@ -0,0 +1,117 @@
...
+  getPlayerName,
+} from 'utils/playerHelpers';
+
+type Props = {
+  player: PlayerFragmentFragment;
+};
+
+export const PatronTile: React.FC<Props> = ({ player }) => {
+  return (
+    <MetaTile key={player.id}>
```
pakokrew, line 33: I don't think you need to put `key` again here, the one on PatronList l.13 is enough for React, if I'm not wrong

packages/web/graphql/getPatrons.ts
pakokrew: As the token balance is a remote schema, hasura has to query TheGraph for every user and get its balances. One of this query will be super long to process (here 50 calls to TheGraph each loop)

Yes, it will be done only once at static generation, but maybe we should find another way of doing that.

Maybe we can query directly the Seed Graph to get a sorted list of patrons addresses, then get the corresponding players ?

dan13ram: Valid points. Firstly I was just checking, the `ethereum_address` stored in the backend is not in lowercase, and theGraph always gives us lowercase addresses. And we can't query our database using that. 

So first I'm creating a separate issue to address that before we can go ahead with this. 

https://github.com/MetaFam/TheGame/issues/404

peth-yursick:
hey man, what's the latest with this? @dan13ram

dan13ram:
We needed the lowercase ethereum_address working which we just merged a
couple days back.
Now I'm looking at the best way to query the top patrons and if that could
be done in the backend itself.

pakokrew (approved):
Looks good.

I would just say that the patron tiles are too big, for most of them there is an empty space in the middle. I believe it's because of the fact we don't display skills in patrons.

Maybe reduce the size of the cards for patrons, if you're keen to work a little bit more on this, but it's not a major issue. 

Screenshot 2021-03-16 at 09 20 34

Screenshot 2021-03-16 at 09 20 29

PR #398, opened 2021-03-09
Guild Onboarding with Discord: Backend

### LINKED ISSUE #396: add jsonb field to guild and populate ranks, e.g. { ranks: [] }
https://github.com/MetaFam/TheGame/blob/1050b6ef2547dd8fa470632904fc7cc41544c9af/packages/backend/src/handlers/triggers/updateDiscordRole.ts#L39-L44

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `todo` comment in 1050b6ef2547dd8fa470632904fc7cc41544c9af when #355 was merged. cc @MetaFam.

alalonde:
Makes some postgres schema changes to ease the forthcoming changes for #184 

Also used the new schema to simplify the logic in the `updateDiscordRole` action. 

Resolves #396

### DIFF
--- a/hasura/migrations/1615594547538_guild_add_discord_metadata_jsonb/up.sql
+++ b/hasura/migrations/1615594547538_guild_add_discord_metadata_jsonb/up.sql
@@ -1,0 +1,5 @@
+ALTER TABLE "public"."guild" ADD COLUMN "discord_metadata" jsonb NULL;
+
+UPDATE "public"."guild" SET "discord_metadata" = ('{"inviteUrl": "' || discord_invite_url || '"}')::jsonb;
+
+ALTER TABLE "public"."guild" DROP COLUMN "discord_invite_url" CASCADE;
--- a/packages/backend/src/handlers/triggers/updateDiscordRole.ts
+++ b/packages/backend/src/handlers/triggers/updateDiscordRole.ts
@@ -1,10 +1,11 @@
 /* eslint-disable no-console */
 import { createDiscordClient } from '@metafam/discord-bot';
-import { Role } from 'discord.js';
 
 import { AccountType_Enum, Player, PlayerRank_Enum } from '../../lib/autogen/hasura-sdk';
 import { client } from '../../lib/hasuraClient';
 import { TriggerPayload } from './types';
+
+type RankRoleIds = {[rank in PlayerRank_Enum]: string};
 
 export interface UpdateRole {
   playerId: string;
@@ -36,27 +37,13 @@ export const updateDiscordRole = async (
     // instantiate discord client. We'll need serverId, playerId, and roleIds
     const discordClient = await createDiscordClient();
 
-    // todo add jsonb field to guild and populate ranks, e.g. { ranks: [] }
     const guild = await discordClient.guilds.fetch(discordGuildAccount.identifier);
     if (guild == null) {
       console.warn(`No discord server found matching ${discordGuildAccount.identifier}!`);
       return;
     }
 
-    // since we don't have roleIds, we'll need to look up the roles in the server 
-    // and just match by strings for the time being.
-    // https://discord.com/developers/docs/resources/guild#get-guild-roles
-    const rankDiscordRoles = new Map<PlayerRank_Enum, Role>();
-
-    const discordRoleManager = await guild.roles.fetch();
-    discordRoleManager.cache.forEach(discordRole => {
-      const rankMatch = Object.values(PlayerRank_Enum).find(rank => {
-        return rank.toUpperCase() === discordRole.name.toUpperCase();
-      }); 
-      if (rankMatch != null) {
-        rankDiscordRoles.set(rankMatch, discordRole);
-      }
-    })
+    const rankDiscordRoleIds = getGuildResponse.guild[0]?.discord_metadata?.rankRoleIds as RankRoleIds;
 
     const discordPlayer = await guild.members.fetch(discordPlayerAccount.identifier);
     if (discordPlayer == null) {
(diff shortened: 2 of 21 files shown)


pakokrew:
What is the choice behind removing Discord from PlayerAccount, and putting it into player ?
Seems like we are having user identifiers on third party platforms splitted between Player model and PlayerAccount relation.

Do we need to keep both ? If yes, which one are on Player, which one on PlayerAccount ? If no, should we move everything ?

alalonde:
> What is the choice behind removing Discord from PlayerAccount, and putting it into player ?
> Seems like we are having user identifiers on third party platforms splitted between Player model and PlayerAccount relation.

Yes, good question. Essentially we are making Discord a first-class integration with these changes. The discord bot will be moving into our monorepo and there will be quite a bit of discord-related code moving forward, so this change will make joins, etc. cleaner and simpler. This was originally recommended by @hammadj as part of this effort.

> Do we need to keep both ? If yes, which one are on Player, which one on PlayerAccount ? If no, should we move everything ?

`PlayerAccount` will remain as there are plenty of "referenced" third-party accounts for which we don't anticipate doing deep integrations.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -17,73 +18,77 @@ export const updateDiscordRole = async (
...
+  console.log(`updateDiscordRole action triggered for player (username=${newPlayer?.username})`);
+
   try {
     if (newPlayer == null) return;
    
     const getPlayerResponse = await client.GetPlayer({ playerId: newPlayer.id });
-    const discordPlayerAccount = getPlayerResponse.player_by_pk?.Accounts?.find(a => a.type === AccountType_Enum.Discord);
-    if (discordPlayerAccount?.identifier == null) return;
+    const playerDiscordId = getPlayerResponse.player_by_pk?.discord_id;
+    if (playerDiscordId == null) return;
```
pakokrew, line 28:
```suggestion
    if (!playerDiscordId) return;
```
in case `getPlayerResponse.player_by_pk?.discord_id` returns undefined

hasura/migrations/1615594547538_guild_add_discord_metadata_jsonb/up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "public"."guild" ADD COLUMN "discord_metadata" jsonb NULL;
+
+UPDATE "public"."guild" SET "discord_metadata" = ('{"inviteUrl": "' || discord_invite_url || '"}')::jsonb;
```
pakokrew, line 3: I don't know how works jsonb in postgres, but what means `||` ? and isn't `discord_invite_url || '"` missing `:` ?

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -17,73 +18,77 @@ export const updateDiscordRole = async (
...
 
     const newRank = newPlayer?.rank;
 
     if (newRank == null) return;
```
pakokrew, line 32: Same for undefined

pakokrew (approved):

packages/backend/src/handlers/triggers/updateDiscordRole.ts
dysbulic: He got me with that one too, `undefined == null` is `true`.

hasura/migrations/1615594547538_guild_add_discord_metadata_jsonb/up.sql
dysbulic: `||` is SQL string concatenation.

packages/backend/src/handlers/triggers/updateDiscordRole.ts
```diff
@@ -17,73 +18,77 @@ export const updateDiscordRole = async (
...
     // Add the new rank.
-    const discordRoleForRank = rankDiscordRoles.get(newRank);
+    const discordRoleForRank = rankDiscordRoleIds[newRank];
     if (discordRoleForRank == null) {
-      console.warn(`discord role associated with ${newRank} was not found!`);
+      console.warn(`Discord role associated with ${newRank} was not found!`);
     } else {
       await discordPlayer.roles.add(discordRoleForRank);
-      console.log(`Added role ${newRank} for player ${newPlayer?.username} in discord`);
+      console.log(`$  {newPlayer?.username}: added role ${newRank}`);
```
dysbulic, line 91:
```suggestion
      console.log(`${newPlayer?.username}: added role ${newRank}`);
```

dysbulic (approved):
Did you have to merge `master` when there were conflicts in `schema.graphql`?

I'm having trouble b/c I need to run `yarn generate` to satisfy typescript, but whichever `schema.graphql` I use (`HEAD` or `master`), there are missing variables.

The process I did to try to fix it is:
```bash
git checkout master
yarn docker:clean
yarn docker:start
git checkout @mymeta/pers-type-part-1 hasura/migrations/1616101057636_create-mtg-color-table
git checkout @mymeta/pers-type-part-1 hasura/migrations/1616124361730_drop-enneagram-table
yarn hasura migrate apply
yarn hasura metadata export
yarn update-schema
```

The last two commands don't generate anything new. When I connect to the Hasura console and try to track the `ColorAspect` table I get an error it is already tracked.

alalonde:
> Did you have to merge `master` when there were conflicts in `schema.graphql`?
 
I am constantly rebasing off master, and actually had two separate branches each with different migrations. That proved to be a nightmare to deal with so I combined them into this branch, which is not as coherent as I would like..

PR #399, opened 2021-03-10
[MyMeta] updated seed-balances remote schema logic

dan13ram:


packages/backend/src/handlers/remote-schemas/resolvers/getTokenBalance/queries.ts
```diff
@@ -1,8 +1,8 @@
 import { gql } from 'graphql-request/dist';
 
 export const GetTokenBalance = gql`
-  query GetTokenBalance($address: String!) {
-    userTokens(where: { address: $address }) {
+  query GetTokenBalance($address: ID!) {
```
dysbulic, line 4: ¿Since this returns SEED & pSEED balances, should it be `GetTokenBalances`?

The same basic question applies throughout where a singular "`balance`" was used, but multiple balances are returned.

dysbulic (approved):
Looks comprehensible. Just had a relatively unimportant question on plurality…

PR #400, opened 2021-03-11
fixed daohaus shares bug

dan13ram:
Props to @dysbulic who debugged this issue !!

### DIFF
--- a/packages/backend/src/handlers/remote-schemas/typeDefs.ts
+++ b/packages/backend/src/handlers/remote-schemas/typeDefs.ts
@@ -40,8 +40,8 @@ export const typeDefs = gql`
     summoner: String!
     title: String
     version: String
-    totalShares: Int!
-    totalLoot: Int!
+    totalShares: String!
+    totalLoot: String!
   }
 
   type Member {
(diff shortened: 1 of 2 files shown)


dysbulic (commented):
The parsed value is [used to compute a percentage](https://github.com/MetaFam/TheGame/blob/5534ff1a45642ee0590410487c5fa8a4bbdfc276/packages/web/components/Player/Section/PlayerMemberships.tsx#L80). I'm pretty sure you'll need a `parseInt` where it is accessed or some other accommodation.

dan13ram:
> The parsed value is [used to compute a percentage](https://github.com/MetaFam/TheGame/blob/5534ff1a45642ee0590410487c5fa8a4bbdfc276/packages/web/components/Player/Section/PlayerMemberships.tsx#L80). I'm pretty sure you'll need a `parseInt` where it is accessed or some other accommodation.

But it is not computing a percentage, but simply printing it. I double-checked this before raising the PR :)

dysbulic (approved):

dysbulic:
> But it is not computing a percentage, but simply printing it. I double-checked this before raising the PR :)

I don't know why I was thinking I had seen a percentage printed somewhere. I don't even have any DAOs. 🤷

PR #401, opened 2021-03-11
[MyMeta] Use the Five-Color Mana system in place of the Enneagram

dysbulic:
This set of changes accomplishes three things:

* Replaces the Enneagram with something akin to Magic's Five-Color system. It wasn't permissible to use any of the Magic iconography or naming for intellectual property issues.
* Allows navigation of the profile editing flow using a keyboard.
* Enables navigation of the profile setup on mobile.


hasura/metadata/tables.yaml
```diff
@@ -4,18 +4,27 @@
   is_enum: true
 - table:
     schema: public
-    name: EnneagramType
-  is_enum: true
+    name: ColorAspect
```
alalonde, line 7: This should be an enum table, no? we use lowercase for naming normal tables and PascalCase for enum tables

hasura/migrations/1614544655501_create-mtg-colors-table/up.sql
```diff
@@ -0,0 +1,77 @@
...
+  SET description = 'Perfection through knowledge…'
+  WHERE name = 'Wisdom'
+;
+UPDATE "ColorAspect"
+  SET description = 'Satisfaction through ruthlessness…'
+  WHERE name = 'Ambition'
+;
+UPDATE "ColorAspect"
+  SET description = 'Freedom through action…'
+  WHERE name = 'Chaos'
```
alalonde, line 61: Where did these descriptions come from? Curious why there are only a handful, too

alalonde (commented):
This is a massive PR. It would really be helpful for review purposes to split it up into the three items in the description.  If feasible, anyway..

dysbulic:
> This is a massive PR. It would really be helpful for review purposes to split it up into the three items in the description. If feasible, anyway..

I should be able to at last separate the personality type things and the other parts of the profile editing flow.

I'll go back to the drawing board.

PR #406, opened 2021-03-12
add if not null check to ethereum address

dan13ram:


dysbulic (approved):
I can't dig into the data to be 100% sure, but this looks good to me.

dan13ram:
> I can't dig into the data to be 100% sure, but this looks good to me.

Yeah I tested locally, and it worked

PR #407, opened 2021-03-12
Revert ethereum address lowercase

dan13ram:


dysbulic (approved):
Looks reasonable. I'm a little worried that the constraint has already been added to the db and removing the migrations will cause metadata integrity issues, but we can give it a shot and see.

PR #410, opened 2021-03-15
added seed-graph as a new remote schema

dan13ram:
Adding this to make it easier to query (sort, filter, etc) the seed-graph directly via a single graphql url from frontend

### DIFF
--- a/hasura/metadata/remote_schemas.yaml
+++ b/hasura/metadata/remote_schemas.yaml
@@ -2,3 +2,7 @@
   definition:
     url_from_env: REMOTE_SCHEMA_ENDPOINT
     timeout_seconds: 60
+- name: seed-graph
+  definition:
+    url: https://api.thegraph.com/subgraphs/name/dan13ram/seed-graph
+    timeout_seconds: 60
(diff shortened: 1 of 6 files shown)


pakokrew (commented):
Don't know if we should do this. Why do you need it ?

Problem when adding remote-schema, is that external schema are merged onto our root schema. Here, it will add `transfers + userTokens` to our Root schema, next to our `Users, Guilds,...` etc
I understood that it's on Hasura's roadmap to add namespaces or more custom import of remote schemas, but not yet available, and we'd really need that.

But adding the remote schema really depends on why we do this. 
Will the subgraph directly called by the frontend ?
If it will be only called from a custom backend handler (that'll return a list of patron for instance) we don't need it to our public schema and keep it clean.

dan13ram:
> Don't know if we should do this. Why do you need it ?

Hey you can see how I'm using it in https://github.com/MetaFam/TheGame/pull/395
This way we are only doing a total of 2 calls to our backend.

But I agree to your point of keeping our schema clean. I'll see if we can do this directly in a backend handler and return only a list of patrons.

dan13ram:
@pakokrew I've updated the logic and no longer using a new remote-schema

pakokrew (approved):
Nice like this :)

PR #416, opened 2021-03-16
Support connection from smart wallets

### LINKED ISSUE #413: [MyMeta] Fix connection from smart wallets
Our auth system use wallet signatures to generate tokens. 

Unfortunatelly, smart contract wallet use a different system than a simple `eth_sign`, and we need to implement something different to support them.

Example of smart contract wallets: 
- Argent
- Gnosis Safe
- Pillar

Links:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md
https://github.com/WalletConnect/walletconnect-example-dapp/pull/63
https://docs.argent.xyz/wallet-connect-and-argent

pakokrew:
Trying to fix #413 

As this does change the backend, preview will not be working here.

Smart wallets use a different mechanism when signing messages. See https://docs.argent.xyz/wallet-connect-and-argent

This implementation works with Argent, but may not work with other wallets which doesn't implement the EIP in the same fashion.

The biggest problem here is that the authentication hook, which is called everytime there is a request on hasura, has to query on-chain information (contract code + verify signature), which will slow down a lot execution times. Auth hooks must be super-fast. 

The mechanism we have setup previously only required to verify a signature off-chain, which is not super cheap but neither super expensive. Now we're going through async requests on infura nodes ... But has the advantage to be trust-less, with the user self-signing tokens. 

We have to lookup for better solutions to be able to authenticate in a DID (decentralized-identity) way, without lags on requests. Maybe IDX will provide something better for that. At the moment I don't see any faster way of doing it apart from the traditional centralized JWS mechanism.

### DIFF
--- a/packages/utils/src/ethereumHelper.ts
+++ b/packages/utils/src/ethereumHelper.ts
@@ -1,9 +1,9 @@
-import { providers } from 'ethers';
+import { Contract, providers, utils } from 'ethers';
 
 export async function signerHelper(
   provider: providers.Web3Provider,
   rawMessage: string,
-) {
+): Promise<string> {
   const ethereum = provider.provider;
   const signer = provider.getSigner();
   const address = await signer.getAddress();
@@ -18,4 +18,47 @@ export async function signerHelper(
     params,
   });
   return signature;
+}
+
+const smartWalletABI = [
+  'function isValidSignature(bytes32 _message, bytes _signature) public view returns (bool)',
+];
+
+enum WalletType {
+  EOA,
+  SMART,
+}
+
+async function getWalletType(
+  address: string,
+  provider: providers.BaseProvider,
+): Promise<WalletType> {
+  const code = await provider.getCode(address);
+  return code === '0x' ? WalletType.EOA : WalletType.SMART;
+}
+
+export async function verifySignature(
+  address: string,
+  message: string,
+  signature: string,
+  provider: providers.BaseProvider,
+): Promise<boolean> {
+  const walletType = await getWalletType(address, provider);
+  if (walletType === WalletType.EOA) {
+    const recoveredAddress = utils.verifyMessage(message, signature);
+    return address === recoveredAddress;
+  }
+  if (walletType === WalletType.SMART) {
+    const arrayishMessage = utils.toUtf8Bytes(message);
+    const hexMessage = utils.hexlify(arrayishMessage);
+    const hexArray = utils.arrayify(hexMessage);
+    const hashMessage = utils.hashMessage(hexArray);
+
+    const contract = new Contract(address, smartWalletABI, provider);
+
+    const returnValue = await contract.isValidSignature(hashMessage, signature);
+
+    return returnValue;
+  }
+  throw new Error('unsupported wallet type');
 }
--- a/packages/utils/src/ethereumHelper.ts
+++ b/packages/utils/src/ethereumHelper.ts
@@ -44,21 +44,23 @@ export async function verifySignature(
   provider: providers.BaseProvider,
 ): Promise<boolean> {
   const walletType = await getWalletType(address, provider);
+
   if (walletType === WalletType.EOA) {
     const recoveredAddress = utils.verifyMessage(message, signature);
     return address === recoveredAddress;
   }
-  if (walletType === WalletType.SMART) {
-    const arrayishMessage = utils.toUtf8Bytes(message);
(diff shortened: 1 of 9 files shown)


packages/utils/src/ethereumHelper.ts
```diff
@@ -19,3 +19,46 @@ export async function signerHelper(
...
+    const hexArray = utils.arrayify(hexMessage);
+    const hashMessage = utils.hashMessage(hexArray);
+
+    const contract = new Contract(address, smartWalletABI, provider);
+
+    const returnValue = await contract.isValidSignature(hashMessage, signature);
+
+    return returnValue;
+  }
+  throw new Error('unsupported wallet type');
```
dan13ram, line 63: This code is unreachable. 
Instead do you think we should wrap the previous logic in a try catch and throw 'unsupported wallet type' if something throws inside? 
I say this because (as you have mentioned) other smart wallets will have non-`0x` code but may not always support `isValidSignature`.

pakokrew: You're right, this condition is here because I started to use the argent-sdk which returns a specific type for argent. 

I'll correct this. But I was more asking for help on the bigger picture, see the PR main post ;)

packages/utils/src/ethereumHelper.ts
```diff
@@ -19,3 +19,48 @@ export async function signerHelper(
...
+    return address === recoveredAddress;
+  }
+
+  // Smart wallet
+  const arrayishMessage = utils.toUtf8Bytes(message);
+  const hexMessage = utils.hexlify(arrayishMessage);
+  const hexArray = utils.arrayify(hexMessage);
+  const hashMessage = utils.hashMessage(hexArray);
+
+  const contract = new Contract(address, smartWalletABI, provider);
```
dan13ram, line 59: This also throws sometimes if the contract doesn't support the ABI, and should be inside the try-catch

dan13ram (changes requested):
Just a small change, otherwise looks good!

In general, its great that we are supporting smart-wallets! 🤓️

Regarding supporting other wallets: It will be tedious to support every kind of wallet out there. We can strive to support standards only such as eip-1271 in this case.

Regarding the issue of speed: The speed issue appears only in the case of smart-wallets, and imo this is a trade-off the user is bearing vs the added security that they gain by the use of smart-contract wallets.

packages/utils/src/ethereumHelper.ts
dan13ram: Yes I've shared my thoughts below :)

packages/utils/src/ethereumHelper.ts
pakokrew: Are you sure about this ? It is a constructor, so it's sync, so I don't think it can go check the contract code and throw here

pakokrew:
> Regarding supporting other wallets: It will be tedious to support every kind of wallet out there. We can strive to support standards only such as eip-1271 in this case.
We'd have to list the wallet we want to support and track their implementation. For now, the only ones I know supporting EIP 1271 are Pillar and Argent. Gnosis Safe talks about it in their doc, but their code throws 'unsupported'.

> Regarding the issue of speed: The speed issue appears only in the case of smart-wallets, and imo this is a trade-off the user is bearing vs the added security that they gain by the use of smart-contract wallets.

If you feel it's not a problem to have long query time, lets go for it, but we may have to find some trick later ...

dan13ram:
> If you feel it's not a problem to have long query time, lets go for it, but we may have to find some trick later ...

Correct me if I'm wrong but this long query time is _only_ in case of smart-wallets, which were rendered useless earlier? 
Now with these changes, atleast the users can login albeit with a delay; which imo is better than not being able to login. 

We can create a separate issue to research this and find a better solution if-need-be in the future. Along with that maybe we could do some user testing or a survey to find out how many of our users really use smart-wallets and how many of them have an issue with the delay.

dan13ram (approved):

packages/utils/src/ethereumHelper.ts
dan13ram: True. I've had some issues with ethersJS sometimes though. But never mind. Lets go ahead with this

pakokrew:
> Correct me if I'm wrong but this long query time is _only_ in case of smart-wallets, which were rendered useless earlier?

Yes the delay will only happen for users logged in with smart wallet, but will apply to every request they will do.

Lets go for this and see how we can do better later

PR #417, opened 2021-03-16
Spin up a backend in Github Actions for the web build to point to 

### LINKED ISSUE #389: Set up a testing environment for the backend
I have a pull request, #384, that is blocked b/c no one can test it b/c it makes changes to Hasura's structure and there's no testing environment for Hasura.

There should be some workflow for testing changes that alter the backend structure.

alalonde:
To resolve #389

### DIFF
--- a/.github/workflows/PR-CI.yml
+++ b/.github/workflows/PR-CI.yml
@@ -57,7 +57,9 @@ jobs:
         uses: styfle/cancel-workflow-action@0.4.0
         with:
           access_token: ${{ github.token }}
+
       - uses: actions/checkout@v2
+
       - uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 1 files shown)


dysbulic (approved):
This looks pretty direct.

It makes an instance available during the build, but it shuts it down after the build runs?

So, we would still need a service to keep instances running to access for review?

alalonde:
> This looks pretty direct.
> 
> It makes an instance available during the build, but it shuts it down after the build runs?

Yes, that's correct. I assume github actions shuts it down when the build completes; i didn't put anything in there explicitly.

> 
> So, we would still need a service to keep instances running to access for review?

Yes, this solely solves the frontend <-> backend build issues.

Issue #403, opened 2021-03-12
[MyMeta] Next button in the profile editing flow doesn't stay disabled


dysbulic:
In the profile editing flow there is a consistent button at the bottom of the screen to accept the current input and switch to the next screen.

This button is disabled once it is pressed. The disabling though is only for the period of time that the app is talking to Hasura, so it becomes reenabled before the next page loads.

I went through the editing process, and you can see the issue repeatedly.

https://user-images.githubusercontent.com/181523/110888525-5bc38400-82ba-11eb-9289-6cabefa78683.mp4

The button should stay disabled until the next page loads. The flashing isn't a real danger for corrupted data or anything, but it looks sloppy.

PR #418, opened 2021-03-18
set next button as loading until next page loads

dan13ram:
Closing #403

### DIFF
--- a/packages/web/components/Setup/SetupDone.tsx
+++ b/packages/web/components/Setup/SetupDone.tsx
@@ -3,11 +3,12 @@ import { FlexContainer } from 'components/Container';
 import { PlayerTile } from 'components/Player/PlayerTile';
 import { useUser } from 'lib/hooks';
 import { useRouter } from 'next/router';
-import React from 'react';
+import React, { useState } from 'react';
 
 export const SetupDone: React.FC = () => {
   const router = useRouter();
   const { user } = useUser({ redirectTo: '/' });
+  const [loading, setLoading] = useState(false);
   return (
     <FlexContainer flex={1}>
       <MetaHeading mb={10}>Game on!</MetaHeading>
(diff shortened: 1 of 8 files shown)


packages/web/components/Setup/SetupDone.tsx
```diff
@@ -20,10 +21,14 @@ export const SetupDone: React.FC = () => {
       >
         {user?.player && <PlayerTile player={user.player} />}
         <MetaButton
-          onClick={() => router.push('/')}
+          onClick={() => {
+            setLoading(true);
```
dysbulic, line 25: ¿Have you seen a delay on this component updating state?

I'm pretty sure this update is fast enough that we don't need the `loading` check, but it shouldn't hurt anything to have this.

packages/web/components/Setup/SetupUsername.tsx
```diff
@@ -33,14 +35,16 @@ export const SetupUsername: React.FC<SetupUsernameProps> = ({ username, setUsern
       if (error.message.includes('Uniqueness violation')) {
         errorDetail = 'This username is already taken 😢';
       } else if (error.message.includes('username_is_valid')) {
-        errorDetail = 'A username can only contain letters, numbers, and dashes.';
+        errorDetail =
```
dysbulic, line 38: I try to always have a character that marks the end of an indentation block.

```suggestion
        errorDetail = (
          'A username can only contain letters, numbers, and dashes.'
        );
```

dysbulic (approved):
This looks pretty straightforward. You added a `loading` state variable that maintains the `isLoading` property as `true` until the page load finishes.

I have no significant suggestions.

packages/web/components/Setup/SetupDone.tsx
dan13ram: Yeah. The delay is noticeable when testing in `dev` mode, but it won't be in prod.
Like you said though, it won't hurt to keep it.

PR #420, opened 2021-03-19
[MyMeta] Add color picker; remove Enneagram

dysbulic:
At the request of @alalonde, I'm breaking #401 into smaller chunks. This chunk adds the color information, replaces the Enneagram in the profile editing section, and removes the Enneagram table.


dysbulic:
render.com doesn't like the # in this branch name. I'm going to have to reopen this.

PR #421, opened 2021-03-19
Deploy frontend through render.yaml

dysbulic:
In order to use render.com to spin up a complete edition of the site for #389, a copy of the frontend needs to be generated as well.

This request adds a frontend target to the `render.yaml` configuration, and adds a couple variables to make Hasura work better.

There is [no variable interpolation](https://render.com/docs/yaml-spec#properties-available-to-environment-variables:~:text=render.yaml%20does%20not%20support%20variable%20interpolation) in the `render.yaml` config, so it is necessary to do the interpolation in code. This adds a check to `config.ts` allowing for either the entire URL to be set, or just the hostname in which case a render.com URL is formed using it.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -40,6 +40,21 @@ services:
           type: web
           property: host
 
+  - type: web
+    name: frontend
+    env: node
+    plan: starter
+    buildCommand: yarn
+    startCommand: yarn web:dev
+    envVars:
+      - key: GRAPHQL_HOST
+        fromService:
+          name: hasura
+          type: web
+          property: host
+      - key: NEXT_PUBLIC_GRAPHQL_URL
+        value: https://$GRAPHQL_HOST/v1/graphql
+
 databases:
   - name: metagame-db
     plan: starter
--- a/render.yaml
+++ b/render.yaml
@@ -1,3 +1,4 @@
+previewsExpireAfterDays: 3 # automatically remove preview instances
 services:
   # A Docker web service
   - type: web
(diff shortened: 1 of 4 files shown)


alalonde:
Is the intent to use render.com instead of vercel?

render.yaml
```diff
@@ -20,6 +21,10 @@ services:
           property: hostport
       - key: HASURA_GRAPHQL_SERVER_PORT
         value: 10000
+      - key: HASURA_GRAPHQL_ENABLE_CONSOLE
+        value: true
+      - key: PORT
```
alalonde, line 26: Do we still need the other port config also?

render.yaml
```diff
@@ -40,6 +45,21 @@ services:
...
 
+  - type: web
+    name: mymeta
+    env: node
+    plan: starter
+    buildCommand: yarn && yarn web:build
+    startCommand: yarn --cwd packages/web/ start
+    envVars:
+      - key: NODE_VERSION
+        value: 12.18.3
```
alalonde, line 56: I think we just have 12.x defined in the build. Would be good to keep that the same if you can just specify the major version

alalonde (approved):

render.yaml
dysbulic: Yeah, `HASURA_GRAPHQL_SERVER_PORT` has no effect on the setting and without it the instance isn't reachable.

render.yaml
dysbulic: I'll see if I can set it that way.

dysbulic:
> Is the intent to use render.com instead of vercel?

No, this is for spinning up test instances with a modified backend w/o having to also create a new Vercel project.

The cost for Vercel is $7/service/month. Running postgres, hasura, express, & next.js is $28/month times the number of open requests. I think it could get expensiveish.

I don't want to monkey with things too much, just sometimes quickly bring up a test environment.

PR #438, opened 2021-03-25
[Quests] Backend updates

pakokrew:
Only backend code from #435

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,6 +48,7 @@ services:
       - ./packages/discord-bot:/usr/src/app/packages/discord-bot
       - ./package.json:/usr/src/app/package.json
       - ./tsconfig.base.json:/usr/src/app/tsconfig.base.json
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -48,6 +48,7 @@ services:
       - ./packages/discord-bot:/usr/src/app/packages/discord-bot
       - ./package.json:/usr/src/app/package.json
       - ./tsconfig.base.json:/usr/src/app/tsconfig.base.json
+      - ./schema.graphql:/usr/src/app/schema.graphql
```
alalonde, line 51: Is this so the container doesn't have to generate this itself?

alalonde (approved):

docker-compose.yml
pakokrew: The problem was that when the schema was updated, we had to rebuild the backend docker image. The watchers that runs inside can't know about changes in hasura so doesn't regenerate it (as opposed to TS files generated from this schema and queries in our code)

PR #439, opened 2021-03-27
Move players page to /players

pakokrew:


packages/web/pages/index.tsx
```diff
@@ -1,24 +1,18 @@
...
 
-type Props = InferGetStaticPropsType<typeof getStaticProps>;
-
-export const getStaticProps = async () => {
-  const players = await getPlayers();
+export const getStaticProps = () => {
   return {
-    props: {
-      players,
+    redirect: {
```
dysbulic, line 7: Returning a redirect during page generation raises an error. That was killing your Vercel build, now it is dying b/c Hasura has changed on `master`.

dysbulic (changes requested):

packages/web/pages/index.tsx
pakokrew: Can you show me the log ?
This is on the documentation https://nextjs.org/docs/basic-features/data-fetching

pakokrew: Ok just saw this in the nextjs doc:

> Note: Redirecting at build-time is currently not allowed and if the redirects are known at build-time they should be added in next.config.js.

So I don't see when this feature can be supported 😵 

So we have to remove all of redirects ? what should we replace it with then ?

dan13ram: If this is raising issues, we could do this like so?
```
import React, { useEffect } from "react";
import Router from 'next/router'

...
useEffect(() => {
   const {pathname} = Router
   if(pathname == '/' ){
       Router.push('/hello-nextjs')
   }
 });
 ```
 ref: https://stackoverflow.com/a/58182678/5513532

Comment (pakokrew) [reply packages/web/pages/index.tsx:7]: @dysbulic will rebase, would love to see vercel's logs on this build.
It looks like the errors appear only when you specifically ask a dynamic route to render with getStaticPath, but it redirects. Should not happen here, not dynamic page. 

https://nextjs.org/docs/messages/gsp-redirect-during-prerender

Comment (pakokrew) [reply packages/web/pages/index.tsx]: using redirects in next.config now.

File: packages/web/next.config.js [hunk 2]
```diff
@@ -1,4 +1,14 @@
 // eslint-disable-next-line
 const withImages = require('next-images');
 
-module.exports = withImages();
+module.exports = withImages({
+  async redirects() {
```
dysbulic, line 5: This looks like the solution.

dysbulic (approved):

packages/web/pages/index.tsx
```diff
@@ -1,24 +1,10 @@
+import { LoadingState } from '@metafam/ds';
```
dan13ram, line 1: is this page necessary now that we are adding the redirect in the config itself?

dan13ram (approved):
we can remove the unnecessary `index.tsx` file and then merge :)

PR #446, opened 2021-03-29
Guild Onboarding with Discord: OAuth implementation

alalonde:
This implements the Discord OAuth guild exchange and persistence of the fetched guild information. Next will be the frontend page for the guild owner to edit the fetched information.


packages/web/config.ts
```diff
@@ -18,4 +18,8 @@ export const CONFIG = {
     process.env.NEXT_BRIGHTID_APP_URL || 'https://app.brightid.org',
   brightIdNodeUrl:
     process.env.NEXT_BRIGHTID_NODE_URL || 'http:%2f%2fnode.brightid.org',
+  discordBotClientId: 
+    process.env.NEXT_DISCORD_BOT_CLIENT_ID || '808843072980647986',
+  backendUrl:
```
pakokrew, line 23: I'm afraid the backend service is not exposed in production. On **render**, it is defined as a private service, and I believe it should stay like this.

The `redirect_uri` should point to a public path, like a frontend page that'll call the backend handler via an hasura action, or even better a NextJS API route for keeping secrets server-side if needed.

pakokrew (changes requested):

packages/web/config.ts
alalonde: Oh man, good catch.  Apparently hasura doesn't support exposing plain ol' express REST endpoints, so I'll look into going the next.js API route.

alalonde: Hmm, looked into next.js API routes and it doesn't seem like a good fit. Routing server-only traffic through our frontend framework, translating the requests into graphql, etc.  I don't think this will be the last time we need to expose a REST endpoint, either.   What if we just stood up an nginx reverse proxy with a whitelist of exposed endpoints on the backend instances?

alalonde: It appears that we should be able to just use [rewrite/redirect rules](https://render.com/docs/redirects-rewrites) to do this in *render*. We can just explicitly rewrite the desired public route and create a 401 page or something as a catch-all for all other routes.

pakokrew: Why would we need to expose REST endpoints, when we already have a graphql endpoint which can handle any actions ? Here you don't really need a REST endpoint but a redirect URL. 

IMHO I think is better to prefer stick with an oauth Next API route, and see later if we really need other REST endpoints. Not sure rewrites will be able to reach our backend. And not sure having to setup and manage a reverse proxy with a whitelist is cheaper than having Next APIs

pakokrew (approved):

Issue #325, opened 2021-02-01
[MyMeta] - Add actual guild pages


peth-yursick:
Right now guilds only have cards:
![image](https://user-images.githubusercontent.com/55857222/106524523-fb833a80-64e2-11eb-9fa4-d90b73f2f7ee.png)
They should have profiles similar to player profiles

alalonde:
Can you provide some sort of design / layout for what a guild profile might look like? I assume it would be its own page much like player profiles currently

peth-yursick:
> 
> 
> Can you provide some sort of design / layout for what a guild profile might look like? I assume it would be its own page much like player profiles currently

Yep! It will be pretty much the same as the player profiles except with different info.
Here's some of it: https://www.figma.com/file/n36ggiRBQEGzRUbGxWfkMX/Meta-Game-Copy?node-id=58%3A2

mquellhorst:
I've set up the guild pages but as far as I see there ain't much info in the database to fill the page with yet.
Is there anymore to find about a single guild except what's in the guild table?

mquellhorst:
#460

PR #457, opened 2021-03-29
Guild pages with data available so far

mquellhorst:
Addresses #325


mquellhorst:
Going to push from local branch into the repo instead

PR #460, opened 2021-04-02
Feature/guild pages

mquellhorst:
The following data is needed to fill in the required information

guild.twitter_url
guild.github_url
guild.reddit_url ??
guild.mastodon_url ??
guild.Players
guild.quests

Guild should be able to define their NFT gallery address

We need to define what we will place in the Stats block

### DIFF
--- a/packages/web/components/Guild/GuildLinks.tsx
+++ b/packages/web/components/Guild/GuildLinks.tsx
@@ -1,42 +1,75 @@
-import { Button, Wrap, WrapItem } from '@metafam/ds';
+import { IconButton, Wrap, WrapItem } from '@metafam/ds';
 import { GuildFragmentFragment } from 'graphql/autogen/types';
 import React from 'react';
-import { FaDiscord, FaGlobe } from 'react-icons/fa';
+import { FaDiscord, FaGithub, FaGlobe, FaTwitter } from 'react-icons/fa';
+
+import { ProfileSection } from '../ProfileSection';
 
 type Props = {
   guild: GuildFragmentFragment;
 };
 
-export const GuildLinks: React.FC<Props> = ({ guild }) => (
-  <Wrap>
-    {guild.website_url ? (
-      <WrapItem>
-        <Button
-          as="a"
-          href={guild.website_url}
-          target="_blank"
-          size="xs"
-          colorScheme="blackAlpha"
-          leftIcon={<FaGlobe />}
-        >
-          Website
-        </Button>
-      </WrapItem>
-    ) : null}
-    {guild.discord_invite_url ? (
-      <WrapItem>
-        <Button
-          as="a"
-          href={guild.discord_invite_url}
-          target="_blank"
-          size="xs"
-          bgColor="discord"
-          _hover={{ bgColor: 'discordDark' }}
-          leftIcon={<FaDiscord />}
-        >
-          Discord
-        </Button>
-      </WrapItem>
-    ) : null}
-  </Wrap>
-);
+export const GuildLinks: React.FC<Props> = ({ guild }) => {
+  return (
+    <ProfileSection title="Links">
+      <Wrap>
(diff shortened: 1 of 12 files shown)


packages/web/components/Guild/GuildTile.tsx
```diff
@@ -46,14 +51,6 @@ export const GuildTile: React.FC<Props> = ({ guild }) => (
...
     </MetaTileHeader>
     <MetaTileBody>
-      <VStack spacing={2} align="stretch">
-        <Text fontFamily="mono" fontSize="sm" color="blueLight">
-          LINKS
-        </Text>
-        <HStack mt="2">
-          <GuildLinks guild={guild} />
-        </HStack>
-      </VStack>
```
alalonde, line 56: What is the rationale for removing the links ?

packages/web/components/Guild/GuildLinks.tsx
```diff
@@ -1,44 +1,75 @@
...
+                size="lg"
+                bgColor="discord"
+                icon={<FaDiscord />}
+              />
+            </a>
+          </WrapItem>
+        ) : null}
+        {guild.website_url ? (
+          <WrapItem>
+            <a href={guild.website_url} target="_blank" rel="noreferrer">
```
alalonde, line 48: wrong link. Let's not show this if a github URL is not available

packages/web/components/Guild/GuildLinks.tsx
```diff
@@ -1,44 +1,75 @@
...
+              <IconButton
+                variant="outline"
+                aria-label="Discord Server"
+                size="lg"
+                colorScheme="blackAlpha"
+                icon={<FaGlobe />}
+              />
+            </a>
```
alalonde, line 26: Personally, I prefer the previous styling

alalonde (changes requested):

packages/web/components/Guild/GuildTile.tsx
mquellhorst: Cause the links are in a different place on the guild details page

packages/web/components/Guild/GuildLinks.tsx
mquellhorst: yeah I know, I put this in temporarily cause the fields aren't available yet

packages/web/components/Guild/GuildLinks.tsx
mquellhorst: This is a lot closer to the styling from the designs

packages/web/components/Guild/GuildTile.tsx
alalonde: Seems useful to have them in the tiles at-a-glance as well ? There isn't a whole lot of info at the moment on a particular tile

packages/web/components/Guild/GuildLinks.tsx
alalonde: Is the design in that figma document?

mquellhorst: yeah

packages/web/components/Guild/GuildTile.tsx
mquellhorst: Yeah well we could keep them in the GuildHero than we need to update the design.
A guild could have guite a lot of links tho which makes the hero a bit too messy

packages/web/components/Guild/GuildLinks.tsx
alalonde: OK, we don't want to show the github link is website_url is set though.

alalonde:
Sorry for the delay on this. Can you rebase and make those changes, then we'll merge it in

alalonde (approved):
Rebased and added two missing fields for guilds.

PR #463, opened 2021-04-06
Fixed webhook auth for public requests

alalonde:
Found this bug while implementing the discord oauth handler. I believe it hasn't been an issue before because any graphql requests are happening at compile-time currently.

### DIFF
--- a/packages/backend/src/handlers/auth-webhook/handler.ts
+++ b/packages/backend/src/handlers/auth-webhook/handler.ts
@@ -11,7 +11,7 @@ const unauthorizedVariables = {
 function getHeaderToken(req: Request): string | null {
   const authHeader = req.headers.authorization;
   if (!authHeader) return null;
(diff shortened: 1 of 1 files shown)


packages/backend/src/handlers/auth-webhook/handler.ts
```diff
@@ -11,10 +11,10 @@ const unauthorizedVariables = {
 function getHeaderToken(req: Request): string | null {
   const authHeader = req.headers.authorization;
   if (!authHeader) return null;
-  if (authHeader.substring(0, 7) !== 'Bearer ')
+  if (authHeader.substring(0, 7) !== 'Bearer')
```
pakokrew, line 14:
```suggestion
  if (authHeader.substring(0, 6) !== 'Bearer')
```
?

alalonde: Yes, that's correct, good catch.

PR #469, opened 2021-04-20
Fix sourcecred account migrator

### LINKED ISSUE #468: XP no longer updating on MyMeta
The XP bot is also down, so the only way for players to see their XP is by checking the [https://metafam.github.io/XP/#/accounts](Sourcecred instance) directly.

alalonde:
Upgraded sourcecred to 0.8.4, read cred graph using new sc API

Resolves #468


packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
```diff
@@ -71,13 +72,32 @@ const parseAlias = (alias: SCAlias) => {
...
+const ledgerLoadedPromise = manager.reloadLedger();
+
 export const migrateSourceCredAccounts = async (
   _: Request,
   res: Response,
 ): Promise<void> => {
-  const accountsData: SCAccountsData = await (
-    await fetch(ACCOUNTS_FILE)
-  ).json();
+  const ledgerRes: ReloadResult = await ledgerLoadedPromise;
```
dysbulic, line 91:
```suggestion
  const ledgerRes: ReloadResult = await ledgerLoadedPromise();
```

dysbulic (commented):
I'm unfamiliar with the syntax at one point. Can you omit the parens when using await?

packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
alalonde: `ledgerLoadedPromise` is of type `Promise<ReloadResult>`.  It is different syntax than usual, but correct.

Usually one would write something like `const ledgerLoadedResult = await manager.reloadLedger()`; Here I'm just assigning the promise to a variable, and awaiting that. 

The benefit here is that `reloadLedger` is called when the backend boots up. We await it to ensure that it is complete before the handler proceeds with execution.

dysbulic (approved):

packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
dysbulic: Gotcha. Very cool. Thanks for adding a trick to my quiver.

PR #475, opened 2021-04-24
[MyMeta] Allow username to be completely removed

dysbulic:
The current code replaces a users name with the name from the backend if the username input is completely cleared.

These changes allow it to be removed and add soft hyphens in a couple places to allow cleaner word splitting on mobile.

Paired with: @jsusim.

### DIFF
--- a/packages/web/components/Setup/SetupAvailability.tsx
+++ b/packages/web/components/Setup/SetupAvailability.tsx
@@ -65,7 +65,7 @@ export const SetupAvailability: React.FC<SetupAvailabilityProps> = ({
   return (
     <FlexContainer>
       <MetaHeading mb={5} textAlign="center">
-        Availability
+        Avail&#xAD;ability
       </MetaHeading>
       <Text mb={10}>
         What is your weekly availability for any kind of freelance work?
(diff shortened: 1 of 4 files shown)


packages/web/components/Setup/SetupAvailability.tsx
```diff
@@ -65,7 +65,7 @@ export const SetupAvailability: React.FC<SetupAvailabilityProps> = ({
   return (
     <FlexContainer>
       <MetaHeading mb={5} textAlign="center">
-        Availability
+        Avail&#xAD;ability
```
alalonde, line 68: why?

alalonde (commented):
Do we really want to be able to remove a username?

packages/web/components/Setup/SetupAvailability.tsx
dysbulic: It's a soft hyphen that will break on displays that are too narrow.

One of the default mobile devices in Chrome will trigger it: iPhone 4, I think.

alalonde: Ah, interesting

alalonde:
> > Do we really want to be able to remove a username?
> 
> It is that the input can't be cleared to enter a completely new name. When they delete the last character, it puts their old name in the blank.

Ha! That's super annoying! Good find..

alalonde (approved):

Issue #482, opened 2021-04-27
Fix intermittent build failures in Github Actions


alalonde:
I made a change to our build a few weeks ago to stand up a docker container with our backend code for the web build to point to.  This seems to be failing now.  This may not be a best practice but it needs some investigation to see what's going on.

dan13ram:
@alalonde 
Builds are still failing.. actually its just stuck at `yarn wait-on` for 
https://github.com/MetaFam/TheGame/pull/515
But it passed another PR https://github.com/MetaFam/TheGame/pull/529

alalonde:
Does `/healthz` even exist on hasura? May need to use a different health check endpoint.

dan13ram:
@alalonde It does exist. https://hasura.io/docs/latest/graphql/core/api-reference/health.html
This code works locally for me as well as has worked for other PRs (ref #529 and #548)

I just don't why it works sometimes and then fails at other times!

alalonde:
Yeah, #515 error is due to a separate issue I believe. I think we're good here.

PR #483, opened 2021-04-27
Fix intermittent Docker build failures

alalonde:
Addresses #482

### DIFF
--- a/.github/workflows/PR-CI.yml
+++ b/.github/workflows/PR-CI.yml
@@ -89,8 +89,14 @@ jobs:
           HASURA_GRAPHQL_ADMIN_SECRET: so-secretish
           HASURA_PORT: 8080
 
+      - name: Debug docker containers
+        run: docker ps -a
+
       - name: Build Web
         run: yarn web:build
         env:
           CI: true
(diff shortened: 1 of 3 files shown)


.github/workflows/PR-CI.yml
```diff
@@ -89,8 +89,11 @@ jobs:
           HASURA_GRAPHQL_ADMIN_SECRET: so-secretish
           HASURA_PORT: 8080
 
+      - name: Debug docker containers
+        run: docker ps -a
+
       - name: Build Web
-        run: yarn web:build
+        run: yarn wait-on http://localhost:4000/healthz && yarn web:build
```
dan13ram, line 96: Do you think we should wait for backend at port 4000 or hasura at port 8080?

dan13ram (approved):
Looks straightforward to me except for a small doubt I have

.github/workflows/PR-CI.yml
alalonde: Good question. Not sure if hasura has a similar built-in health check. This is already better than it was, so I think if we are still seeing issues we can point it to hasura insteadl

PR #485, opened 2021-04-28
computing patron rank in react

dan13ram:
@alalonde I've copied that code you've used in the `@metafam/backend` onto `@metafam/web` for computing the rank.

Definitely think we should have this code in a single place. Perhaps in `@metafam/utils` but that would require moving the hasura graphql types as well, which seems unnecessary right now.

anyways let me know what you think.

### DIFF
--- a/packages/web/utils/rankHelpers.ts
+++ b/packages/web/utils/rankHelpers.ts
@@ -1,0 +1,32 @@
+import { PlayerRank_Enum } from 'graphql/autogen/types';
+
+export const RANKS = [
+  PlayerRank_Enum.Diamond,
+  PlayerRank_Enum.Platinum,
+  PlayerRank_Enum.Gold,
+  PlayerRank_Enum.Silver,
+  PlayerRank_Enum.Bronze,
+];
+
+export const PLAYERS_PER_RANK = [7, 7, 7, 14, 21];
+
+// A summation of PLAYERS_PER_RANK.
+// This is the first index for which players will NOT be ranked, e.g.
+// the 56th player will be Bronze, and the 57th player will not be ranked.
+export const RANKED_CAP: number = PLAYERS_PER_RANK.reduce((sum, rankCount) => {
+  return sum + rankCount;
+}, 0);
+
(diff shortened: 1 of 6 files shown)


dan13ram:
@alalonde I've rebased this. Please merge this first.
The `fix/dependencies` branch builds off of this.

packages/web/utils/rankHelpers.ts
```diff
@@ -0,0 +1,32 @@
...
+// A summation of PLAYERS_PER_RANK.
+// This is the first index for which players will NOT be ranked, e.g.
+// the 56th player will be Bronze, and the 57th player will not be ranked.
+export const RANKED_CAP: number = PLAYERS_PER_RANK.reduce((sum, rankCount) => {
+  return sum + rankCount;
+}, 0);
+
+// Computes the rank for the given index. This would be the index corresponding
+// to all players ordered by total_xp DESC.
+export function computeRank(totalRankIndex: number): PlayerRank_Enum | null {
```
alalonde, line 22: you could refactor this by passing in PLAYERS_PER_RANK and RANKS to this function, and just have it return a string. I could see these two parameters being different anyway between players/patrons

alalonde (approved):
Optional refactor

packages/web/utils/rankHelpers.ts
```diff
@@ -1,31 +1,24 @@
...
+const computeRankCap = (usersPerRank: Array<number>) =>
+  usersPerRank.reduce((sum, rankCount) => {
+    return sum + rankCount;
+  }, 0);
 
 // Computes the rank for the given index. This would be the index corresponding
 // to all players ordered by total_xp DESC.
-export function computeRank(totalRankIndex: number): PlayerRank_Enum | null {
-  if (totalRankIndex >= RANKED_CAP) return null;
+export function computeRank(
```
alalonde, line 10: Yup, can you move this to the utils package and update the backend code to use it too?

dan13ram: Done :)

PR #511, opened 2021-05-05
Load accounts.json in migration handler again

alalonde:
Depends on https://github.com/MetaFam/XP/pull/56


dan13ram (approved):
Seems straightforward. But should we remove the github api token from config as well?
Also .env.sample and guides too.

alalonde:
> Seems straightforward. But should we remove the github api token from config as well?
> Also .env.sample and guides too.

We'll still need it in the forthcoming migrated bot functionality, so I'll leave it in for now.

PR #541, opened 2021-05-11
[MyMeta] increased player listing to 150

### LINKED ISSUE #498: [MyMeta] Increase player listing
Increase the number of players listed on https://my.metagame.wtf from 50 to 150

dan13ram:
Closes #498

### DIFF
--- a/packages/web/pages/players.tsx
+++ b/packages/web/pages/players.tsx
@@ -1,13 +1,27 @@
 import { PageContainer } from 'components/Container';
 import { PlayerList } from 'components/PlayerList';
+import { PlayerFragmentFragment } from 'graphql/autogen/types';
 import { getPlayers } from 'graphql/getPlayers';
 import { InferGetStaticPropsType } from 'next';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/web/pages/players.tsx
```diff
@@ -1,13 +1,27 @@
...
 type Props = InferGetStaticPropsType<typeof getStaticProps>;
 
+const LIMIT = 50;
+const OFFSET = 50;
+const TOTAL_PLAYERS = 150;
+
 export const getStaticProps = async () => {
-  const players = await getPlayers();
+  const promises: Promise<PlayerFragmentFragment[]>[] = new Array(
+    TOTAL_PLAYERS / OFFSET,
```
dysbulic, line 16: I don't think you need two variables here. The calculations are really based on the limit for how many to return. The offset is a variable computed value.

```suggestion
    TOTAL_PLAYERS / LIMIT,
```

dysbulic (commented):
Looks good to me other than I'd have the offset solely as a computed value and only have the limit as a variable.

dysbulic (approved):
👍

PR #556, opened 2021-05-13
Define render.com staging environment

### LINKED ISSUE #471: Set up staging environment
We need a staging environment to move more quickly with PRs

alalonde:
Resolves #471


alalonde:
We just need to figure out how to point the associated vercel instance to the instance stood up by render.

dan13ram:
> We just need to figure out how to point the associated vercel instance to the instance stood up by render.

Haha this is the imp point

dan13ram (approved):
LGTM, Lets figure out the vercel part now

dan13ram:
We discussed in the Builders call today regarding having a separate branch `develop` for staging deployments instead of having previews on render.

hammadj:
I'm cool with setting up a develop branch for the staging environment, but we should setup some devops tooling that can automatically open the PR from develop into master. However there's also a big downside where the develop branch can break after a PR is merged and then becomes an additional burden / difficulty for someone to fix develop. Often times that doesn't happen for a while and develop gets super bloated and huge and feature releases slow down.

I'm gonna look into Hasura Cloud and see if they have blue/red deployments, that would be the ideal solution IMO (assuming that Hasura changes are the main reason we want this staging deployment)

alalonde:
> I'm cool with setting up a develop branch for the staging environment, but we should setup some devops tooling that can automatically open the PR from develop into master. However there's also a big downside where the develop branch can break after a PR is merged and then becomes an additional burden / difficulty for someone to fix develop. Often times that doesn't happen for a while and develop gets super bloated and huge and feature releases slow down.

Makes sense, however, currently production will just be broken in this case! So this seems preferable.  

> 
> I'm gonna look into Hasura Cloud and see if they have blue/red deployments, that would be the ideal solution IMO (assuming that Hasura changes are the main reason we want this staging deployment)

No, it's really so we have a place to test large feature additions in a production-like environment. I am thinking that whomever does the merge to develop should be responsible for seeing that the new integration is sufficiently tested, and afterwards merging to master. 

I can certainly see issues arising where e.g. this new integration fails testing and then blocks subsequent progress.

PR #566, opened 2021-05-17
added filters for /players page

### LINKED ISSUE #503: [MyMeta] Add filters to /players page
Right now, there is only a list of players.
If we want to make the player explorer more useful, we should allow people to search & filter it.

A player should be able to search the player list, filter it skills & availability at the very least.

The design:
image
[Figma file](https://www.figma.com/file/zDHqledo5H2aKR3KDvbur8/Meta-Game?node-id=3110%3A4014)

dan13ram:
Working towards closing #503


alalonde:
Ooo nice!  Some UX things to improve: 

- Enter key should submit search form
- Timezones should display offset (e.g. -07:00) at least

dysbulic:
The search by text only appears to work if there's more than one character. I don't suppose many people will notice, but maybe don't do the search unless there's two or more characters.

hammadj:
If I search and then remove the search text, it doesn't go back to the unfiltered list.

Also feel like the UX might be better if it was a floating sidebar on desktop and a collapsible section on mobile.

Also would be nice to be able to filter by DAO membership, could replace timezone for that.

dan13ram:
Thanks for the feedback @hammadj .
> If I search and then remove the search text, it doesn't go back to the unfiltered list.
I've fixed the search bar issue. But for the other things I would like to do them in a separate PR if you don't mind. 

> Also feel like the UX might be better if it was a floating sidebar on desktop and a collapsible section on mobile.
I'd like some design feedback from @davort and others regarding the mobile/desktop ideas you have. If we have the design, I can definitely take it from there and build it out :)

> Also would be nice to be able to filter by DAO membership, could replace timezone for that.
Adding the dao membership filters is also not easily done currently since the data is fetched from a remote schema, and we lose filtering options because of that.

dan13ram:
@alalonde do you think we can merge this PR, so I can work on the other feedback in a newer PR. 
I feel this is becoming too big, and its best to split the work into multiple PRs.

hammadj (approved):

davort:
hey @hammadj – thanks for the suggestions. I tried to abstract it a bit in https://github.com/MetaFam/TheGame/issues/503#issuecomment-844853572 and go from there.

Here is a Figma link with the latest designs for the filter UI: https://www.figma.com/file/zDHqledo5H2aKR3KDvbur8/Meta-Game?node-id=5746%3A1577 --> please check it out and lemme know what you think, either here or comment there in Figma. cc @dan13ram

PR #570, opened 2021-05-20
Update migrateSourceCredAccounts script

hammadj:
Fixes flakiness and bugs preventing accounts from being updated correctly.


alalonde:
You can test this in our staging environment now by merging to develop. I just copied all player / account / skill data from prod into the new staging env

alalonde:
I can also send you the CSV dumps from prod if you want to test locally

packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
```diff
@@ -73,46 +48,57 @@ export const migrateSourceCredAccounts = async (
...
 ): Promise<void> => {
-  const accountsData: SCAccountsData = await (
-    await fetch(Constants.SC_ACCOUNTS_FILE)
-  ).json();
+  const ledgerRes = await ledgerManager.reloadLedger();
+  if (ledgerRes.error) {
+    throw new Error(`Unable to load ledger: ${ledgerRes.error}`);
+  }
+
+  const credGraph = await loadCredGraph();
```
alalonde, line 56: Have you monitored node.js during this? I implemented this way briefly for a bit on prod but it caused a large enough memory spike to have to bump up the memory allocation significantly on the instance..

alalonde (approved):
This will be a good first candidate for our test server... 

Any data migrations necessary alongside this?

packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
hammadj: just tested it and seems like memory usage spikes by ~700MB when running that command.

One way I can think of to reduce memory usage is to generate / store a separate file just containing identityIds and cred scores instead of the whole cred graph, and then load accounts using only the ledger.json. 

Another option would be to try out Renders autoscaling since we only need to have higher memory capacity for a short period of time every day when this runs.

PR #602, opened 2021-05-27
Optimize migrateSourceCredAccounts handler + require ETH address for players

hammadj:
Will delete all users without an ETH address set, effectively requiring players to create their profiles in myMeta first and greatly reducing likelihood of naming clashes

### DIFF
--- a/hasura/migrations/1622090276691_require_player_ethaddress/down.sql
+++ b/hasura/migrations/1622090276691_require_player_ethaddress/down.sql
@@ -1,0 +1,18 @@
+
+alter table "public"."player_skill" drop constraint "player_skill_player_id_fkey",
+          add constraint "Player_Skill_player_id_fkey"
+          foreign key ("player_id")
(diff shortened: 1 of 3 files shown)


hasura/migrations/1622090276691_require_player_ethaddress/down.sql
```diff
@@ -0,0 +1,18 @@
...
+          on update no action
+          on delete no action;
+
+alter table "public"."player_account" drop constraint "player_account_player_id_fkey",
+          add constraint "Account_player_id_fkey"
+          foreign key ("player_id")
+          references "public"."player"
+          ("id")
+          on update no action
+          on delete no action;
```
hammadj, line 16: Basically deletes a players related accounts and skills when a player gets deleted. Makes it simpler to do delete operations

alalonde (approved):
Let's make sure documentation is updated right around when this gets deployed to prod

PR #630, opened 2021-06-07
Fixes #357

### LINKED ISSUE #357: [MyMeta] Use text links for the profile setup and disconnect actions
Once a Player connects their wallet, there are links next to their profile picture allowing them to configure their profile or disconnect. These links were switched to icons to save space, but the look is visually unappealing.

Switch back to the text links, and only go to icons when the screen is narrow enough that the profile picture is pushed behind the app drawer (~1100px–`lg` breakpoint).

Screenshot from 2021-02-12 23-09-36

luxumbra:
- Icons show on mobile devices and we have text links for desktop.
- Also tweaked the size and spacing of the icons.
- Line 81 overrides the default button (link variant) hover style which is inconsistent with the default link style.

### DIFF
--- a/packages/web/components/LoginButton.tsx
+++ b/packages/web/components/LoginButton.tsx
@@ -52,29 +52,59 @@ export const LoginButton: React.FC = () => {
             href={`/player/${user.username}`}
             fontFamily="body"
             color="whiteAlpha.700"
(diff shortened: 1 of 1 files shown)


dan13ram (approved):
Great stuff

PR #638, opened 2021-06-09
Allow users to force-update their linked ETH address

hammadj:
Removed sc_identity_id from Player table because it enforces ETH addresses to be used as the single source of truth for users and prevents conflicts when users update their ETH address in the ledger. sc_identity_id is not needing anywhere in our backend / data model.


dan13ram (approved):
LGTM. 
The claim process is directly linked only to the eth address right? So even if a user updates their eth address in our db / sourcecred, their unclaimed seeds could still be claimed using their old address.
And in subsequent mintings they would have to claim them from their new address.

alalonde:
Needs a rebase now

alalonde (approved):

PR #581, opened 2021-05-22
Show DAO memberships across Ethereum, Polygon, and xDai mainnets

mattcan:
This Pull Request pulls DAOs from the DaoHaus subgraphs on TheGraph for Ethereum, Polygon, and xDai mainnets.

with-chain-and-contract

Some details about the change:

* when there is no title for a DAO, a formatted contract ID is shown
* when there is no known logo, a logo for the chain that the DAO is on is shown

Related: #549

PR #643, opened 2021-06-09
Backend only: add Polygon and xDai chains for DAO membership

mattcan:
This uses two new subgraphs to pull DAO memberships for Ethereum, xDai, and Polygon mainnets. The pulls are in parallel. The underlying data is structured the same way but seems to be missing some information, see https://github.com/HausDAO/daohaus-supergraph/issues/44.

These are backend only changes and so the UI is a little bit wonky. Since some DAOs will have no title, there is a default that follows the format of `Unknown ${chain} DAO`. This is pretty meaningless but more meaningful than showing a blank title. Frontend changes will be required to make this look better.

This is split from #581.

## Screenshots

*Before*:

Only Ethereum mainnet DAOs

643-before

*After*:

You can see that there are a number of DAOs without titles

643-altchain-default-title

### DIFF
--- a/packages/backend/src/lib/daoHausClient.ts
+++ b/packages/backend/src/lib/daoHausClient.ts
@@ -3,13 +3,13 @@ import { GraphQLClient } from 'graphql-request';
 import { CONFIG } from '../config';
 import { getSdk, Sdk } from './autogen/daohaus-sdk';
 
-export function clientFactory(chain: string): Sdk {
-  switch (chain) {
-    case 'polygon':
-      return getSdk(new GraphQLClient(CONFIG.daoHausPolygonGraphqlURL));
-    case 'xdai':
-      return getSdk(new GraphQLClient(CONFIG.daoHausXdaiGraphqlURL));
-    default:
-      return getSdk(new GraphQLClient(CONFIG.daoHausGraphqlURL));
-  }
+const defaultGqlClient = new GraphQLClient(CONFIG.daoHausGraphqlURL);
+const clients = new Map([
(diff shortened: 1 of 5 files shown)


packages/backend/src/lib/daoHausClient.ts
```diff
@@ -1,8 +1,15 @@
...
-import { getSdk } from './autogen/daohaus-sdk';
+import { getSdk, Sdk } from './autogen/daohaus-sdk';
 
-export const daoHausClient = getSdk(
-  new GraphQLClient(CONFIG.daoHausGraphqlURL),
-);
+export function clientFactory(chain: string): Sdk {
+  switch (chain) {
+    case 'polygon':
+      return getSdk(new GraphQLClient(CONFIG.daoHausPolygonGraphqlURL));
```
dan13ram, line 9: I'd suggest caching this locally. Otherwise a new GraphQLClient is created on every api call.

dan13ram (changes requested):
Everything looks good. Just requesting a minor change to increase performance

packages/backend/src/lib/daoHausClient.ts
mattcan: Good catch! I have updated that now.

mattcan:
I have modified the way the resolver works to be less surprising and handle errors a bit better. Full detail in https://github.com/MetaFam/TheGame/pull/643/commits/bb600dbbed9fe0c46542c7185ea7bab35b303fa9

dan13ram (approved):
Awesome. Thank you, this is long overdue!

PR #645, opened 2021-06-10
infinite scroll for players page

dan13ram:
The design/UX has been fully reviewed by Dave. 

Latest deployment here: https://the-game-git-feat-pagination-metafam.vercel.app


alalonde:
Playing around with the live preview, there's some weirdness around searching now:

- Typed in "davor" in the search box and hit enter, it shows 0 players but still the 9 original player results.
- Separate issue, but it seems search should be case-insensitive and that "davor" should be matched
- When there are no matches, the label should not be "no more players available" but rather "There were no matches"

alalonde (changes requested):
I found a few issues (see previous comment), please investigate and fix

dan13ram:
@alalonde case insensitive search is already working. But the search is on the username of the player and not their 3box name. 
In this case Davor's username is dave and it can be searched for using "DAVE" or any such case insensitive search

dan13ram:
> Typed in "davor" in the search box and hit enter, it shows 0 players but still the 9 original player results.

I'm unable to reproduce this bug. 
I've fixed the other issue you mentioned :) Please review

alalonde:
> > Typed in "davor" in the search box and hit enter, it shows 0 players but still the 9 original player results.
> 
> I'm unable to reproduce this bug.

It is a bit difficult to reproduce reliably but I see it maybe 25% of the time, some sort of timing / race condition. Here's a video 

https://user-images.githubusercontent.com/1402582/123981755-ba069680-d97f-11eb-8e97-f4ee8d0d1324.mov

dan13ram:
@alalonde I was still unable to reproduce that bug, but I have made some changes to make the implementation better. 
Please check if it works now :)

packages/web/components/Player/PlayerFilter.tsx
```diff
@@ -116,11 +117,15 @@ export const PlayerFilter: React.FC<Props> = ({
...
+      setSearch('');
+      setQueryVariable('search', '%%');
     }
   };
 
-  const { filtersUsed } = useFiltersUsed(queryVariables);
+  const filtersUsed = useFiltersUsed(queryVariables);
+  const isSearchUsed = queryVariables.search !== '%%';
+  const searchText =
+    queryVariables.search?.substr(1, queryVariables.search.length - 2) || '';
```
dysbulic, line 128:
```suggestion
    queryVariables.search?.slice(1, -2) || '';
```

packages/web/lib/hooks/players.ts
```diff
@@ -120,7 +139,74 @@ export const useFiltersUsed = (
...
+    fetching,
+    players: fetchedPlayers,
+    totalCount,
+    error,
+  } = useFilteredPlayers(queryVariables);
+
+  const itemsPerPage = PLAYER_LIMIT;
+  const maxPage = Math.ceil(totalCount / itemsPerPage);
+  const currentOffset = useMemo(() => (queryVariables.offset as number) || 0, [
+    queryVariables.offset,
```
dysbulic, line 160:
```suggestion
    [queryVariables.offset],
```

packages/web/lib/hooks/players.ts
```diff
@@ -120,7 +139,74 @@ export const useFiltersUsed = (
...
+
+  useEffect(() => {
+    if (error || fetching) {
+      if (error || !shouldAppend.current) {
+        setPlayers([]);
+      }
+      return;
+    }
+    if (shouldAppend.current) {
+      setPlayers((_players) => [..._players, ...fetchedPlayers]);
```
dysbulic, line 196:
```suggestion
      setPlayers((players) => [...players, ...fetchedPlayers]);
```

dysbulic: Generally an underscore in front of a variable means it isn't referenced in the body of the function.

packages/web/pages/players.tsx
```diff
@@ -35,26 +35,67 @@ const Players: React.FC<Props> = () => {
...
+    observer.observe(cachedRef);
+
+    return () => observer.unobserve(cachedRef);
+  }, [loadMore]);
+
   return (
     <PageContainer>
       <HeadComponent url="https://my.metagame.wtf/players" />
-      <VStack w="100%" spacing="8">
+      <VStack w="100%" spacing="8" pb={{ base: '16', lg: '0' }}>
```
dysbulic, line 74:
```suggestion
      <VStack w="100%" spacing={8} pb={{ base: 16, lg: 0 }}>
```

dysbulic (commented):
I'll do some testing on the Vercel build & see if I hit any issues.

packages/web/lib/hooks/players.ts
dan13ram: I can't set it as `players` since that variable is used outside. But maybe some other variable name like `oldPlayers`

packages/web/lib/hooks/players.ts
dan13ram: Its already in an array ...

alalonde:
The previous issue I found seems to be fixed. Thanks. Now it looks like the "player type" filter isn't working. I select "Pioneer" and all player types seem to show. Interestingly, when I combine player type with another filter (e.g. skill) the filter is applied. But by itself it doesn't seem to work.

dan13ram:
@alalonde are you sure you are using the latest deployment ?

dan13ram:
Okay thanks for mentioning your browser. I'm able to reproduce the same issue with firefox on my local. Debugging now :)

dan13ram:
Thanks for your reviews @alalonde 🙏 I was able to write better code and resolve the issue with your help!
But please double check if all works fine on your end as well :)

alalonde (approved):

Issue #614, opened 2021-05-31
Create MetaCartel Ventures Guild


vrortvedt:
Certain Blood Mages of MetaCartel Ventures have expressed interest in creating an MCV MetaGame funding guild to be able to interact with the wider crypto community beyond our Moloch contract.

Our DAO is here: https://app.daohaus.club/dao/0x1/0x4570b4faf71e23942b8b9f934b47ccedf7540162/

We'd like to request the creation of an MG guild.

peth-yursick:
@hammadj @alalonde @mquellhorst

PR #648, opened 2021-06-12
migration for insertion of metacartel guild

mquellhorst:
Made a migration file for the guild table to insert the MetaCartel data 

Addresses #614

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -315,6 +315,7 @@
       - timezone
       - username
       filter: {}
+      allow_aggregations: true
   - role: public
     permission:
       columns:
@@ -328,6 +329,7 @@
       - timezone
       - username
       filter: {}
+      allow_aggregations: true
   update_permissions:
   - role: player
     permission:
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -315,7 +315,6 @@
       - timezone
       - username
       filter: {}
-      allow_aggregations: true
   - role: public
     permission:
       columns:
@@ -329,7 +328,6 @@
       - timezone
       - username
       filter: {}
-      allow_aggregations: true
   update_permissions:
   - role: player
     permission:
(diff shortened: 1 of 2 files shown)


dysbulic:
I'm somewhat confused. There is an existing [MetaCartel entry on MyMeta](https://my.metagame.wtf/guild/metacartel). If this is an update, shouldn't it be an `update` rather than an `insert`?

Also, it is likely prudent to include a `delete from guild where guildname = "metacartel"` in the down migration so that if, for some reason, it has to be reverted or rerun, it will succeed.

dysbulic:
Ok, is MetaCartel Ventures a different organization than MetaCartel?

¿Did this migration succeed? It seems like it should have failed with a conflict on `guildname`.

alalonde (changes requested):
This is for MetaCartel Ventures, which has its own DAO. See https://metacartel.xyz/ 

`guildname` must be unique. I believe `metacartel` already exists. You can query the prod mymeta API (api.metagame.wtf) to view the existing guild information.

Also, please reference the original issue (e.g. #issue-num) when creating PRs. Any additional guild information can be requested in that issue.

mquellhorst:
Ah ok I got the idea it was about MetaCartel but you're right, that's already on there and it needs to be MetaCartel Ventures. I'll change the migration and ref the original issue

alalonde:
Did you do a merge or rebase? If done properly, this PR should only show your 2 commits.

hasura/metadata/tables.yaml
```diff
@@ -315,7 +315,6 @@
       - timezone
       - username
       filter: {}
-      allow_aggregations: true
```
alalonde, line 318: I don't think you want to do this? Looks like a bad merge

dan13ram (approved):

hasura/metadata/tables.yaml
mquellhorst: Fixed it :)

alalonde (approved):
Thanks!

PR #663, opened 2021-06-23
added guild position (internal / external) [closes #489]

### LINKED ISSUE #489: [MyMeta] Add Internal Guilds to the guilds page
On the https://my.metagame.wtf/guilds page, there should also be internal guild pages.
The guild profiles should feature information from [this page ](https://www.figma.com/file/l9qXlzJFaNGvWTdxgJCPjf/MetaGame-Onepagers?node-id=0%3A1)
One of the containers should be like this:
Title: How to join
Body: Do one of these things (list evergreen quests from the above figma in bullet points), then get on discord & contact Head Innkeeper of that specific guild (names will be provided when you choose to tackle this issue.
The guild logos should also be ready by then, if not - ask in the #tavern or #artists-guild.

Dependant on #460

mquellhorst:
Added an enum table GuildPosition (INTERNAL/EXTERNAL)
The Guilds have a label indicating their position.

All the current guilds have the 'position' field set to EXTERNAL
The field 'position' in the guild table needs permission to be selected by players and public. This prob has to be done manually!


alalonde:
To close #489, this would need to actually insert the guilds, no?

I would assume we would want a separate table for internal guilds. Take a look at the mockups and see what data is shown, and whether the current schema can accommodate that data, where it is reasonably semantically similar to the data for external guilds.

mquellhorst:
Since we also want to create a way for guilds to add sections. I'ld suggest a section table. The content of the guilds can be stored there and we're not restricting ourselves to specific sections. This also makes it possible to implement internal guilds this way without causing any redundancy in the db. Guilds are guilds, the only difference is external / internal

Issue #508, opened 2021-05-04
Fix periodic vercel build errors


alalonde:
e.g.

```
@metafam/web: Error occurred prerendering page "/player/mlibty". Read more: https://nextjs.org/docs/messages/prerender-error
--
23:44:37.765 | @metafam/web: Error: `redirect` can not be returned from getStaticProps during prerendering (/player/mlibty)
23:44:37.765 | @metafam/web: See more info here: https://nextjs.org/docs/messages/gsp-redirect-during-prerender
```

alalonde:
This may be [quite easy](https://nextjs.org/docs/messages/gsp-redirect-during-prerender) but difficult to test...

dysbulic:
My understanding is the redirect is being returned thanks to [this code](https://github.com/MetaFam/TheGame/blob/dfff04ebaaa21b82403df9c5f4f1cc9361073eca/packages/web/pages/player/%5Busername%5D.tsx#L197-L208) which attempts to look up a user, and, if it isn't found, attempts to look up the lowercased version of the name. When that is found, it redirects to the lowercase url.

The username in question, "mlibty", is already lowercased, but I think the first lookup fails due to a backend timeout.

I'm pretty sure the solution is to [add retries to urql](https://formidable.com/open-source/urql/docs/advanced/retry-operations/) ([here](https://github.com/MetaFam/TheGame/blob/dfff04ebaaa21b82403df9c5f4f1cc9361073eca/packages/web/graphql/client.ts)) so that when it times out it doesn't return an error, but instead waits and tries again.

If it is a timeout error, setting a [fallback page](https://nextjs.org/docs/basic-features/data-fetching#fallback-pages) won't necessarily fix the problem since it could timeout twice in a row. This is less likely, so the error will occur less, but it doesn't address the fundamental issue.

alalonde:
> The username in question, "mlibty", is already lowercased, but I think the first lookup fails due to a backend timeout.

Just curious, have you verified this in the logs?

dysbulic:
There's not really anything in the logs other than the failure. I have seen timeouts when accessing Hasura.

It would explain the intermittent nature of the errors.

dan13ram:
I've verified that Timeout is indeed the issue. Added retries in #548 
But it still fails sometimes. 

Looks like even 2 retries are not enough sometimes.

Perhaps worth thinking about doing some load balancing on hasura.
[...]

Issue #309, opened 2021-01-28
Integrate Ceramic Network 


META-DREAMER:
Create a "hello world" integration of Ceramic / IDX into our repo (e.g. adding 3ID connect to login flow and setting / getting a field from the user profile)

https://twitter.com/ceramicnetwork/status/1354448316512284675?s=20

dysbulic:
For reference, there is [a profile record](https://developers.idx.xyz/guides/definitions/default/#basic-profile) associated with an IDX, but it is currently empty. So, the first time a user logs in with IDX, maybe the system should copy their 3Box profile into IDX?

You can experiment in the [web playground](https://ceramicstudio.github.io/web-playground/) which instantiates the `ceramic`, `idx`, & `did` objects for manipulation using the development console.

Setting and retrieving the profile looks like:

- `await idx.set('basicProfile', { name: 'Δυς' })`
- `await idx.get('basicProfile', did.id)`

dysbulic:
I checked with [3Box support](https://discord.gg/pYXBnKkt) and [3ID-Connect](https://github.com/ceramicstudio/3id-connect) will eventually handle migrating the user profile the first time a user uses IDX "closer to the mainnet release".

dysbulic:
The system which verifies accounts (Discord, Github, Twitter) and associates them with an IDX DID is [identitylink-services](https://github.com/ceramicstudio/identitylink-services). It is possible to run the services ourselves, but there will be a hosted version that is released closer to mainnet.

To support out current coverage, there would also need to be support for Discourse as well.

Github verification requires creating a Gist; Twitter tweeting. To verify on Discourse, a user could have to post a reply with their DID to a special thread dedicated to these verifications. It looks like there's [an API for reading posts in a particular topic](https://docs.discourse.org/#tag/Topics/paths/~1t~1{id}~1posts.json/get).

dysbulic:
I [forked identitylink-services](https://github.com/MetaFam/identitylink-services) to the MetaFam repo to look into adding Discourse support while still garnering my precious XP.

dysbulic:
Another aspect of the IDX integration is using it as the basis for determining a Player's accounts for XP distribution.

I'm working on [IDX as a SourceCred source of truth](https://github.com/sourcecred/sourcecred/issues/2718) for ETHDenver.

dysbulic:
[...]

PR #686, opened 2021-07-06
Add caching of 3Box profile information

dysbulic:
This PR adds a table to the postgres instance to store the information in 3Box's profiles to attempt to alleviate the intermittent timeouts (#508).

It alters an existing action which cached the 3Box verified accounts to pull information from IDX (#309), or, if that's not available, the soon to be retired 3Box site. Linked sites continue to come from 3Box as there is not currently an instance of [IdentityLink](https://github.com/ceramicstudio/identitylink-services) running anywhere to do the IDX-based account links.

The "collectible favorites" which allowed specifying top NFTs from a users collection are not a part of the IDX basic profile and support for them was simply removed.

Currently the `/actions/cache/updateAll` endpoint has no checks on who can execute it. Since the backend server is only accessible from the Hasura instance when deployed, I'm not sure there needs to be any protection.

Still left to investigate is subscribing to changes to the Ceramic documents to invalidate the cache. There are currently 500+ users in the database and I suspect listening to that many streams is infeasible. There's also the problem of discovering when a new profile is created which almost certainly takes polling.

As it is, entries are cached when first the user first logs into the site, and the `updateAll` could easily be run in a Hasura cron task.

The ultimate goal now is to reimplement the interfaces of [self.id](//self.id) in MyMeta's profile setup as well as publish the MetaGame profile additions to Ceramic, specifically the skill sets.


packages/backend/src/handlers/actions/3Box/cache3BoxProfile/handler.ts
```diff
@@ -1,8 +1,8 @@
 import { Request, Response } from 'express';
 
-import { updateVerifiedAccounts } from './updateVerifiedAccounts';
+import { updateCachedProfile } from './updateCachedProfile';
 
-export const updateBoxProfileHandler = async (
+export const cache3BoxProfileHandler = async (
```
hammadj, line 5: We should start using "IDX" instead of 3box for naming

packages/backend/src/handlers/actions/3Box/refresh3BoxCache/cache3BoxProfiles.ts
```diff
@@ -0,0 +1,16 @@
+import { CacheProcessOutput } from '../../../../lib/autogen/hasura-sdk';
+import { client } from '../../../../lib/hasuraClient';
+import { updateCachedProfile } from '../cache3BoxProfile/updateCachedProfile';
+
+export async function cache3BoxProfiles(): Promise<CacheProcessOutput> {
```
hammadj, line 5: the function naming and folder structure is kinda confusing. cache3BoxProfiles function is not inside the cache3BoxProfile folder. Seems like they should all be in the same directory.

packages/backend/src/handlers/actions/3Box/routes.ts
```diff
@@ -0,0 +1,11 @@
+import express from 'express';
+
+import { asyncHandlerWrapper } from '../../../lib/apiHelpers';
+import { cache3BoxProfileHandler } from './cache3BoxProfile/handler';
+import { refresh3BoxCacheHandler } from './refresh3BoxCache/handler';
```
hammadj, line 5: naming is confusing. there's cacheProfile, updateCachedProfile, and refreshCache and its not clear which one does what since they're used interchangeably. Would be good to unify the naming a bit

packages/backend/src/handlers/actions/routes.ts
```diff
@@ -1,17 +1,14 @@
...
 import { questsRoutes } from './quests/routes';
-import { updateBoxProfileHandler } from './updateBoxProfile/handler';
 
 export const actionRoutes = express.Router();
 
-actionRoutes.post(
-  '/updateBoxProfile',
-  asyncHandlerWrapper(updateBoxProfileHandler),
-);
+actionRoutes.use('/cache', cacheRoutes);
```
hammadj, line 11: route should have a more descriptive name like `/cacheIdxProfiles`

packages/backend/src/handlers/actions/3Box/refresh3BoxCache/cache3BoxProfiles.ts
```diff
@@ -0,0 +1,16 @@
...
+  const data = await client.GetPlayerIds();
+  const ids = data.player.map(({ id }) => id);
+
+  try {
+    await Promise.all(ids.map((id) => updateCachedProfile(id)));
```
hammadj, line 10: this would trigger all these requests to happen in parallel rather than one at a time, which would spike CPU/memory usage and also result in timeouts from the API getting hammered with hundreds of concurrent requests.  once we start growing out user base, this wont scale

If we want to iterate over all of them, we should use something like https://github.com/sergelerator/async-batch to limit the concurrency to 5-10 players at a time.

Also, this method of periodically fetching all user data would result in things getting out of sync (if period is too long), or over-fetching unnecessarily and using up a lot of resources (if period is frequent). The ideal strategy to minimize the chance of stale data + prevent having to fetch every single user periodically would be to return the cached data when its queried + trigger an update in the background so that next time someone visits that same player page they get fresh data. For players that no one visits, we don't waste time updating the data frequently and can do it once every 3-4 days instead.

It would also be nice to store a timestamp in the table of when the cached data was last updated and only fetch data from IDX when its X days stale. Basically a TTL.

hammadj (commented):
nice work, some comments on naming / scalability

packages/backend/src/handlers/actions/3Box/cache3BoxProfile/handler.ts
dysbulic: Done.

packages/backend/src/handlers/actions/3Box/refresh3BoxCache/cache3BoxProfiles.ts
dysbulic: There are folders for each of the action handlers. The update implementation is in the same directory as those folders.

packages/backend/src/handlers/actions/3Box/routes.ts
dysbulic: The names are now `updateSingleProfile` for the individual invalidation and `updateExpiredProfiles` for the process that checks for any profiles not updated in the last four days.

packages/backend/src/handlers/actions/routes.ts
dysbulic: It is now `idxCache/`.

packages/backend/src/handlers/actions/3Box/refresh3BoxCache/cache3BoxProfiles.ts
dysbulic: I used [Bottleneck](https://github.com/SGrondin/bottleneck#readme) with a `maxConcurrency` of 10 to rate limit the update processes. It is stored as a shared variable on the Express server, and there's a map for tracking its current contents.

There's a `last_checked_at` field on the cache entries now and a Hasura action that invalidates all entries more than four days old meant to be run from a cron job.

package.json
```diff
@@ -80,5 +80,8 @@
   },
   "resolutions": {
     "bcrypto": "5.2.0"
+  },
+  "dependencies": {
```
dan13ram, line 84: Hey are we using `stylis`? can be removed if not.

dan13ram (commented):
Fantabulous work!! Added some minor comments.

Are you planning to add a cron that updates expired profiles? Perhaps every 4 days in `hasura/metadata/cron_triggers.yaml` ?

package.json
dysbulic: It's referenced by `emotion` and without this I get the error [`stylis.middleware is not a function`](https://github.com/emotion-js/emotion/issues/2138).

dan13ram: I see. Then would it make sense to add it under `design-system` where `emotion` is declared ?

packages/backend/src/index.ts
```diff
@@ -7,6 +8,12 @@ import { errorMiddleware } from './lib/apiHelpers';
 
 const app = express();
 
+app.locals.limiter = new Bottleneck({
+  maxConcurrent: 10,
```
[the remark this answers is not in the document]
dysbulic: I was shooting for something pretty small so as to avoid any possibility of the load being an issue. It takes about five minutes to process all 500 users in the database, and most of the time only a small percentage of those will be in the caching queue.

package.json
dysbulic: I think I tried that and it didn't resolve the issue, but I'll try it that way again just to be sure.

packages/web/components/Guild/GuildJoin.tsx
```diff
@@ -77,7 +77,7 @@ export const GuildJoin: React.FC = () => {
               asking for your permission to collect certain relevant information
               about your guild.
             </Text>
-            <MetaButton size="lg" maxW="15rem" as="a" href={discordAuthUrl}>
+            <MetaButton size="lg" maxW="15rem" as="a" href={discordAuthURL}>
```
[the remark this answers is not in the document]
dysbulic:
```suggestion
            <MetaButton size="lg" maxW="15rem" as="a" href={discordAuthURL} isExternal>
```

packages/web/lib/hooks/opensea.ts
```diff
@@ -36,11 +36,9 @@ export const useOpenSeaCollectibles = ({
   const [loading, setLoading] = useState<boolean>(false);
   const owner = player.ethereum_address;
   const collectiblesFavorites = useMemo(
```
dysbulic, line 38:
```suggestion
```

packages/web/lib/hooks/opensea.ts
```diff
@@ -36,11 +36,9 @@ export const useOpenSeaCollectibles = ({
...
-    () =>
-      player && player.box_profile && player.box_profile.collectiblesFavorites
-        ? player.box_profile.collectiblesFavorites
-        : [],
-    [player],
+    // IDX's basic profile doesn't include a list of favorite NFTs
```
dysbulic, line 39:
```suggestion
```

packages/web/lib/hooks/opensea.ts
```diff
@@ -36,11 +36,9 @@ export const useOpenSeaCollectibles = ({
...
+    () => [], // player?.box_profile?.collectiblesFavorites ?? [],
```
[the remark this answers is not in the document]
dysbulic:
```suggestion
```

packages/web/lib/hooks/opensea.ts
```diff
@@ -36,11 +36,9 @@ export const useOpenSeaCollectibles = ({
...
+    [], // [player],
```
dysbulic, line 41:
```suggestion
```

packages/web/lib/hooks/opensea.ts
```diff
@@ -36,11 +36,9 @@ export const useOpenSeaCollectibles = ({
...
   );
```
dysbulic, line 42:
```suggestion
```

packages/web/lib/hooks/opensea.ts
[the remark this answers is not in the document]
dan13ram: @dysbulic I mean that since `collectibleFavorites` is always `[]`, we can remove the variable and its usage in the below `useEffect` under the first if check.

dysbulic:
> Are you planning to add a cron that updates expired profiles? Perhaps every 4 days in `hasura/metadata/cron_triggers.yaml` ?

Done. It checks once a day for profiles that haven't been updated in the last four days.

packages/web/lib/hooks/opensea.ts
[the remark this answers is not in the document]
dysbulic: Done.

packages/web/components/Guild/GuildJoin.tsx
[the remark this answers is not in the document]
dysbulic: I added this and then in looking at the site, this button sends them to authenticate in Discord which then sends them back to the site, so it is safe to leave this page.

package.json
dysbulic: I removed it and, for the time being, I'm not getting the middleware error, so I just left it out.

dan13ram (approved):

dan13ram:
Amazing stuff @dysbulic !!!

PR #690, opened 2021-07-07
Mobile layout for player filters

dan13ram:
Builds upon #645 

Deployed at https://the-game-git-feat-filters-metafam.vercel.app

PR #693, opened 2021-07-08
better search for timezone filter

dan13ram:
builds upon #690

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -32,6 +32,7 @@
     "@types/react": "17.0.6",
     "@types/react-dom": "17.0.5",
     "@types/react-select": "3.0.22",
+    "city-timezones": "^1.2.0",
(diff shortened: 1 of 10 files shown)


.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.19.0.cjs"
```
dan13ram, line 5: If yarn is greater than 1.19 locally, it throws up with weird errors when you do `yarn add` inside a workspace.

Their docs mention we shouldn't gitignore `.yarnrc` 
https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

alalonde (approved):

PR #694, opened 2021-07-08
Added force option to migrateSourceCredAccounts action

### LINKED ISSUE #666: XP updates to Player page
I know this has been raised in the past.  I looked but did not find any open or closed issues in our XP or Sourcecred repos.  

Page here: https://my.metagame.wtf/players does not reflect XP changes shown here: https://xp.metagame.wtf/#/explorer.

Is that something easy to fix and is there a longer term change to help keep the XP shown in sync?

alalonde:
Main reason for this is to fix what was probably a bad merge.  I also added a `force` option for people to run locally; otherwise there's no easy way to seed players into your local DB.

Should resolve #666  😈

### DIFF
--- a/packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
+++ b/packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
@@ -47,13 +47,16 @@ const parseAlias = (alias: SCAlias) => {
 };
 
 export const migrateSourceCredAccounts = async (
-  _: Request,
+  req: Request,
   res: Response,
 ): Promise<void> => {
   const ledgerRes = await ledgerManager.reloadLedger();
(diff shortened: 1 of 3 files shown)


packages/backend/src/handlers/actions/migrateSourceCredAccounts/handler.ts
```diff
@@ -102,23 +105,42 @@ export const migrateSourceCredAccounts = async (
       async (player) => {
         const vars = {
           ethAddress: player.ethereum_address,
-          identityId: player.scIdentityId,
```
alalonde, line 105: This should not have made it to develop/master

hammadj (approved):

dan13ram (approved):

Issue #687, opened 2021-07-06
Gracefully handle DAOhaus API availability problems


alalonde:
Currently in the test environment we're seeing lots of these:

```
Jul 6 03:25:31 AM  Pulling memberships failed: FetchError: request to https://api.thegraph.com/subgraphs/name/odyssy-automaton/daohaus-xdai failed, reason: connect ETIMEDOUT 104.18.10.19:443
```

I assume that the URL is correct, but the endpoint is down? If so we should anticipate outages and handle them gracefully.

@mattcan

mattcan:
I think what you are seeing is this: https://github.com/MetaFam/TheGame/blob/develop/packages/backend/src/handlers/remote-schemas/resolvers/daohaus/resolver.ts#L32-L39

I should have made that a `console.warn`

mattcan:
Work in progress branch that I started a few weeks ago. Quickly added this test now, and it is passing: https://github.com/MetaFam/TheGame/commit/c836130dc135982a5c87be72e2b2fb1ac7a98b82

(I'm getting tslint errors during commit that I don't understand, which is why I haven't pushed it previously)

PR #696, opened 2021-07-09
Multichain DAO tests + log noise reduction

mattcan:
Related: #687 

---

This adds tests to the daohaus resolver as there is some logic in pulling more than one subgraph.

There is a small tweak to the resolver under test to change the `console.error` to `console.warn` as the request failing is covered by providing a default value.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -48,6 +48,7 @@
   "devDependencies": {
     "concurrently": "6.1.0",
     "eslint-plugin-jest": "24.3.6",
+    "nock": "^13.1.0",
     "ts-node-dev": "1.1.6"
   }
 }
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -48,7 +48,7 @@
   "devDependencies": {
     "concurrently": "6.1.0",
     "eslint-plugin-jest": "24.3.6",
-    "nock": "^13.1.0",
+    "nock": "13.1.0",
     "ts-node-dev": "1.1.6"
   }
 }
--- a/packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
+++ b/packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
@@ -4,17 +4,17 @@ import { Resolver } from '../../../../../src/handlers/remote-schemas/autogen/typ
 import { getDaoHausMemberships } from '../../../../../src/handlers/remote-schemas/resolvers/daohaus/resolver';
 
 type MockResolver<TResult, TParent, TContext, TArgs> = (
-    parent?: TParent | any,
-    args?: TArgs | any,
-    context?: TContext | any,
-    info?: any
+  parent?: TParent | any,
+  args?: TArgs | any,
+  context?: TContext | any,
+  info?: any,
 ) => Promise<TResult> | TResult;
 
 function mockResolver<TResult, TParent, TContext, TArgs>(
-    resolver: Resolver<TResult, TParent, TContext, TArgs>
+  resolver: Resolver<TResult, TParent, TContext, TArgs>,
 ): MockResolver<TResult, TParent, TContext, TArgs> {
-    return (parent: TParent, args: TArgs, context: TContext, info) =>
-      resolver(parent, args, context, info);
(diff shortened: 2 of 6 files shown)


alalonde (approved):
Above and beyond! Well done

packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
```diff
@@ -0,0 +1,116 @@
+import nock from 'nock';
+
+import { Resolver } from '../../../../../src/handlers/remote-schemas/autogen/types';
+import { getDaoHausMemberships } from '../../../../../src/handlers/remote-schemas/resolvers/daohaus/resolver';
+
+type MockResolver<TResult, TParent, TContext, TArgs> = (
+  parent?: TParent | any,
```
dysbulic, line 7:
```suggestion
  parent?: TParent | unknown,
```

packages/backend/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "devDependencies": {
     "concurrently": "6.1.0",
+    "nock": "^13.1.0",
```
dysbulic, line 46: We're using [exact versions](https://github.com/MetaFam/TheGame/commit/51089ae49e8ee2a7a41936fb22b27c2bbe77aae0). *(I'm uncertain as to the effectiveness of doing so, but that is the current standard.)*

packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
dysbulic: I think using `unknown` instead of `any` for all of these will clear up the type errors in the pull request.

dysbulic (changes requested):
Looks good to me in general. I love to see the tests. I very much want to learn Jest & getting to see functional tests is a blessing.

The only real problem is the type errors in the resolver test ("unexpected any"s). I think switching to `unknown`s will resolve that.

packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
mattcan: So I took a look and there are a couple problems:
* linting happens differently on commit (no linting problems) and running from the root dir (brings up these linting errors)
* the root level lint hammers the entire set of code instead of obeying the configurations, which are configured to ignore the test directories
* `unknown` brings me to the same problem I had before: I need to do a bunch of type manipulation to make this work and that seems like a waste of time to me

I propose the fix is to do the following:
* leave the `any` (it is the [recommended solution](https://github.com/dotansimha/graphql-code-generator/issues/1763))
* add a `lint` command to each package that is missing it
* replace the root-level `lint` with `lerna run lint --parallel` so that it obeys the `.eslintrc` config

packages/backend/package.json
mattcan: Good catch, thanks

packages/backend/tests/handlers/remote-schemas/resolvers/daohaus/resolver.test.ts
alalonde: > 
> I propose the fix is to do the following:
> 
>     * leave the `any` (it is the [recommended solution](https://github.com/dotansimha/graphql-code-generator/issues/1763))
> 
>     * add a `lint` command to each package that is missing it
> 
>     * replace the root-level `lint` with `lerna run lint --parallel` so that it obeys the `.eslintrc` config

Yup, agree with all of this.  Wrangling types is sometimes not worth it for test code. You should be able to add an eslint override for the `test` directory also (see the `.eslintrc` in backend project for an example)

mattcan: So I misremembered how eslint works and I thought the pathing overrode the config (I know it happens with some other package but clearly my mind is going). Either way, I'm in progress on this but wasn't able to wrap it up this morning. Will get back to it in a couple days when I have time again.

dan13ram (approved):
Looks great. Sorry for the delay in review, could you rebase with `develop` so we can merge this in :)

PR #704, opened 2021-07-12
showing adjacent timezones for players filters

dan13ram:
also :
+ fix touchscreen issue
+ replaced about section with timezone in player cards
+ fix daoHaus membership links to point to app.daohaus instead of explorer (explorer is useless imo)


packages/web/pages/players.tsx
```diff
@@ -77,24 +82,47 @@ const Players: React.FC<Props> = () => {
           resetFilter={resetFilter}
           totalCount={totalCount}
         />
-        {error && <Text>{`Error: ${error.message}`}</Text>}
-        {!error && players.length && (fetchingMore || !fetching) && (
+        {error ? <Text>{`Error: ${error.message}`}</Text> : null}
```
dysbulic, line 85: ¿Why do you prefer `x ? y : null` over `x && y`?

packages/web/pages/players.tsx
```diff
@@ -77,24 +82,47 @@ const Players: React.FC<Props> = () => {
...
+    const isTimezoneSelected = useMemo(
+      () => queryVariables.timezones && queryVariables.timezones.length > 0,
+      [queryVariables],
+    );
+
+    return (
+      <VStack w="100%" ref={ref}>
+        {fetching ? <LoadingState color="white" /> : null}
+        {!fetching && !isTimezoneSelected && totalCount > 0 ? (
+          <Text color="white">No more players available</Text>
```
dysbulic, line 119:
```suggestion
          <Text color="white">No more players available.</Text>
```

packages/web/components/Player/Filter/AdjascentTimezonePlayers.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const {
+    players,
+    totalCount,
+    fetching,
+    fetchingMore,
+    error,
+    nextPage,
+    moreAvailable,
+  } = usePlayerFilter(variables);
```
dysbulic, line 55: I did some research on this and the way to include players w/o a profile in the query results is apparently to compose the where statement at this level and then pass that to the filter function rather than a list of variables.

I've been meaning to test it out, but haven't had the time yet.

I think requiring players to fill out their profile is always going to leave us with holes in the ranks.

dysbulic (commented):
I had a variety of semantic suggestions, but the overall implementation looks solid.

packages/web/components/Player/Filter/PlayerFilter.tsx
```diff
@@ -242,6 +242,16 @@ export const PlayerFilter: React.FC<Props> = ({
                 />
               </WrapItem>
             ))}
+            {availability && (
+              <WrapItem>
+                <FilterTag
+                  label={`Available >${availability.value} h/week`}
```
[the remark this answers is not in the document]
dysbulic: Given the behavior of the filter, the options should be ≥ as well.

packages/design-system/src/MetaFilterSelect.tsx
```diff
@@ -175,20 +175,35 @@ const SelectControl: React.FC<
...
     hasValue,
-    selectProps: { menuIsOpen, onMenuClose, onMenuOpen },
+    selectProps: { menuIsOpen, onMenuClose, onMenuOpen, showSearch },
   } = props;
 
   const buttonRef = useRef<HTMLButtonElement>(null);
 
+  const onClick = useCallback(() => {
+    if (menuIsOpen) {
+      if (onMenuClose) {
```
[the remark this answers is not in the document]
dan13ram: This is necessary because it can be undefined.

packages/web/pages/players.tsx
dan13ram: This i feel is much clearer to read.

packages/web/components/Player/Filter/AdjascentTimezonePlayers.tsx
dan13ram: Sounds good. Perhaps we can get to this in a future PR

packages/web/assets/no-players-found.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="80" height="64" viewBox="0 0 80 64" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
dan13ram: This changes how it looks. I prefer to leave it as is.

dysbulic (approved):
LGTM.

Issue #204, opened 2020-12-08
[MetaOS] Mega Issue - Design / Roadmap


peth-yursick:
Following Friday's product design raid, we're moving onto building the platform.

## [Platform Design](https://www.figma.com/file/n36ggiRBQEGzRUbGxWfkMX/Meta-Game-Copy?node-id=3198%3A3)

Not sure if we're building mobile-first from the get-go, but this is how I see it working on desktop vs. mobile.
![image](https://user-images.githubusercontent.com/55857222/101480653-4570f980-3954-11eb-9bac-302ec5569bae.png)
- The main differences are that on mobile, the "journal & goals" tabs would be collapsed by default & there would be no navigation bar, but the floating MG crystal instead. It would give the "navigation bar buttons" on single press, and open the full app drawer on double tap.
- If making the mobile version like this makes it impossible to have a single build for both mobile & desktop, we can drop the navigation bar from the desktop version and have also have a floating crystal that slowly follows the mouse as you navigate the page.

For anyone who missed out, here's the full outline of the platform:

### v0.1 - A navigation bar with an app drawer & an empty background.
![image](https://user-images.githubusercontent.com/55857222/101479219-22dde100-3952-11eb-87a2-e47a2ee7597e.png)
- A) navigation bar
- B) nav. with app drawer open
A user must be able to open any of these "apps" inside an iframe window, as suggested in this issue.
- First version of the app drawer ✅
- [Second version issues](https://github.com/MetaFam/TheGame/issues/254)

### v0.11 - The beginning of a dashboard in the background.
![image](https://user-images.githubusercontent.com/55857222/101479368-60db0500-3952-11eb-80c3-5260d6745669.png)
- The design is super-rudimentary, a mockup sort of thing. But unless it gets updated in the coming few days, just go with it. Simply have these containers that display XP & Seeds to begin with, you can drop the morale & carbon footprint for now.

###  v0.12 - A better dashboard.
@Polycarpik' Journal quest should be ready before this step, so it can be integrated as suggested in the "v0.11" rendering above.
The dashboard itself should have a better design by then, and it shouldn't just display "XP & Seeds" but also "+% this week" for each of them. Hopefully we also get the morale indicator by then & someone to estimate our carbon footprint.

### v0.13 - Make it mine! Populating the dashboard.
![image](https://user-images.githubusercontent.com/55857222/101479967-4e150000-3953-11eb-8c6b-fc17e03e39b5.png)
[...]

PR #712, opened 2021-07-16
Fix/cannot update dao listing while rendering link dao

mquellhorst:
State was set inside a component which is rendered inside another component.
This makes it impossible to update the parent component and results in an error message in the console.
See:
https://github.com/facebook/react/issues/18178

### DIFF
--- a/packages/web/components/Player/Section/PlayerMemberships.tsx
+++ b/packages/web/components/Player/Section/PlayerMemberships.tsx
@@ -13,7 +13,7 @@ import {
   useDisclosure,
 } from '@metafam/ds';
 import { PlayerFragmentFragment } from 'graphql/autogen/types';
-import React, { useState } from 'react';
+import React from 'react';
 import { isBackdropFilterSupported } from 'utils/compatibilityHelpers';
 
 import polygonImage from '../../../assets/chains/polygon.png';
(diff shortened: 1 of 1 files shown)


dan13ram:
Hey @mquellhorst I've made some slightly different changes to show the daoHaus urls instead of etherscan along with fixing that bug in #204

Can you check https://github.com/MetaFam/TheGame/blob/d3faafe5416984fc6ed210ba461809d6e188892e/packages/web/components/Player/Section/PlayerMemberships.tsx ?

dan13ram:
Maybe we will push this first and then I'll rebase my branch upon this. What say? @mquellhorst

packages/web/components/Player/Section/PlayerMemberships.tsx
```diff
@@ -90,17 +79,27 @@ const DaoListing: React.FC<DaoListingProps> = ({
...
     message = `XP: ${Math.floor(memberXp || 0)}`;
   } else {
     message = `Shares: ${memberShares}/${daoShares}`;
   }
 
-  const [isLink, setIsLink] = useState(false);
-
   return (
-    <LinkDao chain={chain || ''} address={address || ''} setIsLink={setIsLink}>
+    <LinkDao
```
dan13ram, line 98: chain and address are no longer necessary for this component, maybe we can remove it?

mquellhorst:
I see, yeah sounds good!

packages/web/components/Player/Section/PlayerMemberships.tsx
mquellhorst: You're right, I'll remove them then we can merge this so you can rebase and merge yours

dan13ram (approved):

PR #726, opened 2021-07-29
Updated Frontend.MD to display http://localhost:3000 in documentation. 

tommymaher15:


tommymaher15:
Correct me if this isnt right! 

thought I would quickly do this issue while I do some light reading on others and try gather some momentum.

PR #727, opened 2021-07-31
MyMeta: Make player tile into a link

dysbulic:
This PR changes the player tile to a link. Sublinks to GitHub, Twitter, etc. still function as before.

Paired With: @tommymaher15


dan13ram (approved):
Amazing! Just needs some typecheck fix looks like

PR #729, opened 2021-07-31
resolve moloch titles after DaoHaus schema change

mquellhorst:
Cause of DaoHaus changes in their schema we had failures during build.
The moloch title was moved to the metadata schema.

paired-with: @dysbulic


dysbulic (approved):
LGTM.

PR #749, opened 2021-08-10
Feature/internal guilds 2

mquellhorst:
Added an enum table GuildPosition (INTERNAL/EXTERNAL)
The Guilds have a label indicating their position.

All the current guilds have the 'position' field set to EXTERNAL
The field 'position' in the guild table needs permission to be selected by players and public.


dysbulic (approved):
I see nothing I would change.

PR #785, opened 2021-08-24
added a frequently asked question about Lerna

mquellhorst:
A question I got a couple of times already came up again so I've added it to the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ We're using the following stack:
 
 - [Jest](https://jestjs.io/)
 - [Prettier](https://prettier.io/)
-- [Lerna](https://github.com/lerna/lerna/tree/main/commands/add#readme)
+- [Lerna](https://lerna.js.org)
 - [TypeScript](https://www.typescriptlang.org/)
 - [Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ We're using the following stack:
 
 - [Jest](https://jestjs.io/)
 - [Prettier](https://prettier.io/)
-- [Lerna](https://lerna.js.org/)
+- [Lerna](https://github.com/lerna/lerna/tree/main/commands/add#readme)
```
alalonde, line 26: I think the original link is more appropriate here

README.md
```diff
@@ -66,6 +66,12 @@ Initial TS Compilation For Monorepo Packages
 yarn typecheck
 ```
 
+Adding a package to the codebase can be done with Lerna
```
alalonde, line 69: That new link should probably go here instead

alalonde (changes requested):

README.md
mquellhorst: I noticed the website itself doesn't contain much info for commands as `lerna add ...`
But yeah if I put the link at the other spot in the documentation sounds good. 
Thanks!

README.md
mquellhorst: Thanks for the feedback!

alalonde:
Waiting on @mquellhorst for requested changes..

mquellhorst:
> 
> 
> Waiting on @mquellhorst for requested changes..

Pushed the changes and added a some extra information about the --no-verify option

README.md
```diff
@@ -66,6 +66,12 @@ Initial TS Compilation For Monorepo Packages
...
+
+```shell script
+yarn lerna add <package> packages/<folder> [--dev] [--exact] [--peer]
+```
```
alalonde: This inline example seems like it would be useful

alalonde (approved):

PR #803, opened 2021-09-03
🛠 Added docker image of hasura for M1 Mac Users

Udit-takkar:
Sadly, official docker images of Hasura don't work on M1 yet. I tried this image few weeks ago while setting up the project locally https://hub.docker.com/r/fedormelexin/graphql-engine-arm64/ which starts Hasura properly without errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,8 +74,23 @@ yarn typecheck
 yarn docker:start
 ```
 
+**For M1 Mac Users**
+
+Official docker images of Hasura don't work on M1 yet
+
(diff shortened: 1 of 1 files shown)


alalonde:
Thanks! Couple of thoughts here:
- Guess I'll wait as long as possible before upgrading to M1!
- I appreciate your not just changing the hasura Dockerfile directly, now folks will just have to be careful not to commit this local change!

alalonde (approved):

PR #717, opened 2021-07-22
Create player skill rank migration

chair28980:
Create migration for player skill rank addressing https://github.com/MetaFam/TheGame/issues/441

Issue #441, opened 2021-03-27
[MyMeta] Add "top 3 skills"


peth-yursick:
Have people first specify their "top 3 skills" then display those in a separate line on their profiles.

In the edit / setup profile flow, duplicate the skill picker on the same page, where the first picker is limited to three choices, and the second filters out the three chosen skills. 

Will probably need to add a `featured` column to the `player_skill` table.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 1.0 OTHER   attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/MetaFam/TheGame/441/100025501).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/MetaFam/TheGame/441/100025501).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $1,921,420.49 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

mquellhorst:
The bounty has expired. Is this still planned?

luxumbra:
@mquellhorst Looks like it's still in the roadmap. @peth-yursick is this still needing doing?

mquellhorst:
Yeah this is still open. 
Dunno whether the bounty is still available since it seems to be expired. You'll get XP anyway

luxumbra:
@mquellhorst yeah, I was more asking as it's a good one for one of the cohorts to take on. If it's still wanted, I'll let the engaged octo-builders know it is available.

chair28980:
Looking to pick this one up when design is ready.

davort:
@peth-yursick 
**"Captain obvious" alert! 😄**

For the sake of absolute clarity, can you confirm that the "skills" we are talking about in this issue are what's  currently called "superpowers" (in the profile setup)? Screenshot 👇 
![CleanShot 2021-06-12 at 13 05 24](https://user-images.githubusercontent.com/1573304/123463803-474a9380-d5ec-11eb-85f5-ba57205fe94e.png)

If that's true, might we simply re-do this existing view, so that it matches what you described in https://github.com/MetaFam/TheGame/issues/441#issue-842609609 ?

[...]

PR #807, opened 2021-09-05
Create player skill rank migration (v2)

richteri:
Create migration for player skill rank addressing schema changes for #441 based on PR https://github.com/MetaFam/TheGame/pull/717

Adds id and rank fields to player_skill


alalonde (approved):
Looks pretty straightforward; thanks

PR #816, opened 2021-09-10
Fix duplicate MemberOf label in patron tile

alalonde:
This was an artifact of the guild-onboarding work


mquellhorst (approved):
Nice catch

PR #822, opened 2021-09-15
Added ability to edit guild form info during onboarding process

alalonde:
Fixed a couple of issues:

- If you tried to go through the onboarding process a second time (with the same guild), it would erroneously redirect you to another guild page. I fixed the graphql query and it will now redirect you to the `/join/guild/<guildname>` page.
- Fixed initial population of selected discord roles in the two role selection dropdowns.

### DIFF
--- a/packages/backend/src/handlers/actions/guild/discord/oauthHandler.ts
+++ b/packages/backend/src/handlers/actions/guild/discord/oauthHandler.ts
@@ -54,7 +54,6 @@ export const handleOAuthCallback = async (
         const getGuildResponse = await client.GetGuild({
           id: existingGuild.guild_id,
         });
-
         const successResponse: DiscordGuildAuthResponse = {
           success: true,
           guildname: getGuildResponse.guild[0].guildname,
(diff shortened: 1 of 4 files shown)


dysbulic (approved):
All these changes look reasonable to me assuming you've checked for side-effects.

The first time I tried to run the flow at `/join/guild/` I got an error after the Discord auth screen: "`Could not load your guild information from Discord. State did not match`". *(There should be a period ending that error.)* When I ran it again, it succeeded.

A small UI criticism. When I was completing the form, I didn't initially have an admin role defined. To get the form to pick it up after creating one I had to reload the page which discarded all my existing inputs. It would be nice if there was a `🗘` icon by the input to refresh the roles list. I don't know how probable it is that someone would create a new role for their MetaGame admins, but it could certainly happen.

alalonde:
Thanks for testing / feedback!

> A small UI criticism. When I was completing the form, I didn't initially have an admin role defined. To get the form to pick it up after creating one I had to reload the page which discarded all my existing inputs. It would be nice if there was a `🗘` icon by the input to refresh the roles list. I don't know how probable it is that someone would create a new role for their MetaGame admins, but it could certainly happen.

Added 😁

PR #828, opened 2021-09-17
Fixed megamenu link to go to edit profile flow

alalonde:
Also, closes https://github.com/MetaFam/TheGame/issues/830

### DIFF
--- a/packages/web/components/MegaMenu.tsx
+++ b/packages/web/components/MegaMenu.tsx
@@ -312,6 +312,13 @@ const PlayerStats: React.FC<PlayerStatsProps> = ({ player }) => {
         <MenuList mt="8px" color="black" fontFamily="exo2">
           <MetaLink
             color="black"
+            href={`/player/${player.username}`}
+            _hover={{ textDecoration: 'none' }}
+          >
+            <MenuItem>View Profile</MenuItem>
+          </MetaLink>
+          <MetaLink
+            color="black"
             href="/profile/setup/username"
             _hover={{ textDecoration: 'none' }}
           >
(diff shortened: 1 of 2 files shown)


mquellhorst:
I noted this one before approving the merge and proposed to add the edit link to the menu but this seems to be intentional left out on @davort request communicated by @phys3

davort:
You are correct @mquellhorst - the idea is that it's very easy to jump into "Edit" mode once you're looking at your own profile page. At the same time, "View profile" was always missing from the menu, and my bet is that people go look at their profiles as equally, if not more often than editing them 😊 Design decision.

mquellhorst:
This PR is up already so rather have @alalonde change this one.
@davort what you say about just having both the edit and view profile links in the dropdown for now?
This way we can prob get it in the merge with master now

packages/web/components/MegaMenu.tsx
```diff
@@ -311,10 +312,10 @@ const PlayerStats: React.FC<PlayerStatsProps> = ({ player }) => {
         <MenuList mt="8px" color="black" fontFamily="exo2">
           <MetaLink
             color="black"
-            href={`/player/${getPlayerName(player)}`}
+            href="/profile/setup/username"
             _hover={{ textDecoration: 'none' }}
           >
-            <MenuItem>View Profile</MenuItem>
```
mquellhorst, line 317: Discussed with @davort 
Keep both links for now please :)

mquellhorst (changes requested):

alalonde:
OK, I will add a "view profile"  link to the menu and "edit profile" to your own player page.

alalonde:
@mquellhorst Added these both, please re-review

mquellhorst (approved):

Issue #835, opened 2021-09-20
Edit Profile Guide is Glitchy


Udit-takkar:
https://user-images.githubusercontent.com/53316345/134089698-5188c5fc-52ce-4a75-9d68-5cfdba785d22.mov

Description :- When Trying to forward/backwards the forward, backward buttons and Meta Game icon seem to be glitchy.
Just after clicking forward button they change their state . Something  like below for just a moment.

<img width="1440" alt="Screenshot 2021-09-21 at 5 02 49 AM" src="https://user-images.githubusercontent.com/53316345/134090056-13aaea83-0815-42d7-bf3b-ee07c1445a6f.png">
. 

Steps to Reproduce  :-  
1) Connect to Walltet

2) Go to Edit Profile Option

3) Press forward/backward Button

Udit-takkar:
The page re-renders everytime button is clicked. The image has to load everytime. in the file `components/setup/SetupHeader`. The image tag is used from chakra-ui instead of `next/image`. 
Image tag from `next/image` also gives an attribute priority which enables pre loading of image.

Udit-takkar:
It seem to work with the usage of `next/image`.

https://user-images.githubusercontent.com/53316345/134113864-ba3ccc24-4e32-470f-8659-0e57c01c4561.mov

PR #836, opened 2021-09-22
fix: Setup Header Images

Udit-takkar:
#835

### DIFF
--- a/packages/design-system/src/BoxedNextImage.tsx
+++ b/packages/design-system/src/BoxedNextImage.tsx
@@ -9,10 +9,10 @@ export const BoxedNextImage: React.FC<
   const { src, alt } = props;
   return (
     <Box
+      pos="relative"
       {...{ ...props, src: undefined, alt: undefined }}
       style={{
         objectFit: 'contain',
-        position: 'relative',
       }}
     >
       <NextImage {...{ src, alt }} layout="fill" objectFit="contain" />
--- a/packages/web/components/Setup/SetupHeader.tsx
+++ b/packages/web/components/Setup/SetupHeader.tsx
@@ -1,9 +1,10 @@
-import { Box, Flex, Grid, Image, ResponsiveText } from '@metafam/ds';
+import { Box, Flex, Grid, ResponsiveText } from '@metafam/ds';
 import BackImage from 'assets/Back.svg';
 import LogoImage from 'assets/logo.png';
 import SkipImage from 'assets/Skip.svg';
 import { FlexContainer } from 'components/Container';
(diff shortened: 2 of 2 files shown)


Udit-takkar:
After Fixing:- 
https://user-images.githubusercontent.com/53316345/134290221-0e384671-c685-4f2d-bad9-106692972f45.mov

alalonde:
Thanks! In the future, you can just push your branch directly to the MetaFam repo. Then Vercel will deploy it automatically

dysbulic:
I'm on the fence on this one. In most places we are using [`BoxedNextImage`](//github.com/MetaFam/TheGame/blob/develop/packages/design-system/src/BoxedNextImage.tsx) to leverage `next/image` without requiring specification of a size in pixels.

The only real downsize is if someone has their screen zoomed `1rem` is not necessarily `16px`. I don't know if that's a big enough need to warrant a change.

Udit-takkar:
@alalonde @dysbulic  any suggestions/changes in this one?

alalonde (changes requested):
Generally speaking, we should aim to use these shared components from the `ds` package as much as possible. Would it suffice in this case?

dysbulic:
@Udit-takkar, so far as how to move forward, the consensus seems to be switch to `BoxedNextImage` out of the `design-system` package.

Udit-takkar:
@dysbulic  ok just saw the package now.

Udit-takkar:
If we use  BoxedNextImage then it would look something like this
Screenshot 2021-09-30 at 11 57 40 PM

To fix we just have to override position : "relative" in 
```
<Box
      {...{ ...props, src: undefined, alt: undefined }}
      style={{
        objectFit: 'contain',
        position: 'relative',
      }}
    >
      <NextImage {...{ src, alt }} layout="fill" objectFit="contain" />
    </Box>
```

dysbulic:
> To fix we just have to override position : "relative" in
> 
> ```
> <Box
>       {...{ ...props, src: undefined, alt: undefined }}
>       style={{
>         objectFit: 'contain',
>         position: 'relative',
>       }}
>     >
>       <NextImage {...{ src, alt }} layout="fill" objectFit="contain" />
>     </Box>
> ```

If you have,
```
<Box
  position="relative"
  objectFit="contain"
  {...props}
>
```
will a definition of `position` in `props` override the setting of `relative`?

Udit-takkar:
> relative

No It didn't

Udit-takkar:
oh fixed now was using props first @dysbulic

alalonde (approved):

PR #845, opened 2021-09-29
Added requestPolicy option to useUser hook

### LINKED ISSUE #363: [MyMeta] Editing skills doesn't appear to save changes, but actually does
If you e.g. add a skill and click the next button, then go back, the skill you just added is not there. Nor will it appear on your player card on the home page. Refreshing the page re-fetches the data and the next skill does show up.

Thus, skill changes are persisted to the database but not to the app state.

alalonde:
Resolves #363


leewardbound (approved):
Looks good @alalonde ! We chased our tail for hours on this one, I think this is the fix we needed!

Issue #311, opened 2021-01-28
[MyMeta] Add the role selection screen to the profile creation flow


peth-yursick:
Currently, role selection is done through a Discord bot. We want to add it to the profile creation flow (Connect button [here]( https://my.metagame.wtf/players)). 

We probably want new tables such as `guild_role` and `guild_player_role` to store these new roles. This should support guilds that have a discord integration as well as those that do not. 

Ideally, choosing roles would also apply the appropriate Discord role. For this to work, we'd first need to know a player's discord ID, which we could ask for in a prior step.

alalonde:
I assume the roles should be the same as those in discord? I'm refactoring the profile setup stuff currently for #278 so I'll tackle this once that is done.

IxaBrjnko:
Let's see what the discussion brings tomorrow, it may be better to use a simplified set of roles for the onboarding process...

alalonde:
Sounds good; I wasn't privy to the conversation so any links / details here would be appreciated.

dysbulic:
Since there's going to be a Discord bot that knows what everyone's roles are (#184), should we also have it mirror roles to a Player's profile.

Specifying information in two places is a recipe for things getting out of sync.

dan13ram:
I believe this should be worked on once we have discord oAuth implemented. 
Ref: https://github.com/MetaFam/TheGame/pull/446

davort:
Will this need any design contribution?

peth-yursick:
> 
> 
> Will this need any design contribution?

Just a little bit! 😇
The page for selecting the role can be either pretty much the same as the color selection or player type selection screens, just different icons & text

davort:
> > Will this need any design contribution?
> 
> Just a little bit! 😇
> The page for selecting the role can be either pretty much the same as the color selection or player type selection screens, just different icons & text

In that case, please feel free to ping me about this issue whenever you feel is a good time for a designer to give their 2 cents 😉

peth-yursick:
> 
> 
> > > Will this need any design contribution?
> > 
> > 
> > Just a little bit! 😇
> > The page for selecting the role can be either pretty much the same as the color selection or player type selection screens, just different icons & text
> 
> In that case, please feel free to ping me about this issue whenever you feel is a good time for a designer to give their 2 cents 😉

Whenever! How about tomorrow? Could do it pretty much any time before the designers call or at the call

davort:
[...]

PR #846, opened 2021-09-29
Choose roles in MyMeta part 2: Frontend

alalonde:
For #311 

There are a few Hasura permissions fixes in here so unfortunately the Vercel build is not working..


alalonde:
Triggering another deploy

PR #861, opened 2021-10-15
Enable hasura console on previews

polimyl:
Alec asked me if I can enable the Hasura console on pull request previews, so here it is!
Clicking on the "Hasura" link in the "Successfully deployed..." message should now open the Hasura console.


alalonde (approved):
Easiest PR test ever!

Issue #865, opened 2021-10-17
Pressing backspace in the availability input is bugged


vidvidvid:
See the attached gif:

![availability press backspace](https://user-images.githubusercontent.com/35112939/137637011-b924ddc9-b30e-414d-b8a4-438455df69be.gif)

Udit-takkar:
yeah If I double press it then the last digit appear again

PR #869, opened 2021-10-18
fix: availabilty input bug fixed

Udit-takkar:


packages/web/pages/profile/setup/availability.tsx
```diff
@@ -17,12 +17,15 @@ const AvailabilitySetup: React.FC<DefaultSetupProps> = () => {
...
-      setAvailability(player.availability_hours.toString());
+  useEffect(() => {
+    if (user?.player) {
+      const { player } = user;
+      if (player.availability_hours && !availability) {
+        setAvailability(player.availability_hours.toString());
+      }
     }
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
alalonde, line 27: There should be a way of doing this where this comment is not necessary.. :)

alalonde (changes requested):

alalonde (changes requested):
I believe these eslint rules are generally a best practice. Please either implement this in a way where the rule is not violated, or provide an explanation as to why this is a unique case that cannot be solved any other way.  By providing the latter, you will also be demonstrating sufficiently deep understanding of react hooks and the render lifecycle.

Udit-takkar:
the edit profile setup has been changed now.  now there is a different way to update availability hours. although this same problem exist in all of the input elements if i go to update them. is there any way i can access the onboarding flow again after being registered?

vidvidvid:
> now there is a different way to update availability hours.

how is it different?

> although there are lot of bugs in all of the input elements if i try to update them.

can you specify the bugs and why are they relevant to this issue?

> is there any way i can access the onboarding flow again after being registered?

are you asking for yourself, as a developer or for the user? as a developer, you can navigate to `/profile/setup/username` and make a change in `packages/web/lib/hooks/index.ts`
`if (!redirectTo) return;`
to
`return;`
but this is just for testing purposes. For the user, they will no longer have access to the setup wizard once they complete the steps. They will edit the profile in the dialogs in their profile view.

> I think we are moving away from updating the profile with the page `/profile/setup/availability`

What do you mean?

Udit-takkar:
@vidvidvid  
Bugs like If i try update my username the url does remains the same with old username.
and same bug like in this #865 for input elements like username, availability hours. 

I was asking if #865 need a fix as we will be using `profile/setu/availability` for update but i think as we are saving each step to database so if user goes previous step he should be able update it.

vidvidvid:
> @vidvidvid Bugs like If i try update my username the url does remains the same with old username. and same bug like in this #865 for input elements like username, availability hours.

The issue with changing the name was already observed and it would make sense to try and figure out how to deal with it. One of the possible solutions might be to immediately change the route name, but I'm not sure if that would work. I just opened an [issue for it](https://github.com/MetaFam/TheGame/issues/889).

> I was asking if #865 need a fix as we will be using `profile/setu/availability` for update but i think as we are saving each step to database so if user goes previous step he should be able update it.

The issue needs a fix, that's for sure. The setup wizard and editor both share the core of this component and it still doesn't work. Neither in the setup wizard, nor in the profile page (you can access the dialog that has an input to change availability by clicking the edit button in the player hero).

Udit-takkar:
> issue for it

I tried using router and it didn't break anything

vidvidvid:
how did you use it?

Udit-takkar:
> how did you use it? - did it reload the page?

No it did not reload. Just router.push() with the updated username in the save function after the username has been updated successfully

vidvidvid:
ok, nice solution :)

Udit-takkar:
@vidvidvid  should i also update new edit form in this PR?

vidvidvid:
yes, of course :)

alalonde (approved):

PR #885, opened 2021-10-29
Dashboard structure

vidvidvid:
Split the current dashboard page into components, one of which is a wrapper component for each of the sections `GridItem` and the rest are contents of each section `XP`, `Seed`, `LatestNews`, `Leaderboard` and `Calendar`.

I also made a `utils` file `dashboard` which contains most of the static values for defining the layout of the the dashboard grid. Feel free to add anything in it if you think it belongs there.

`dashboard` page now has the following structure of the grid items: 

```
...
<Box key="xp" className="gridItem">
  <GridItem title="XP" sx={gridConfig.xp}>
    <XP />
  </GridItem>
</Box>
<Box key="seed" className="gridItem">
  <GridItem title="Seed" sx={gridConfig.seed}>
    <Seed />
  </GridItem>
</Box>
...
```
If I wanted to include the `Box` with the `key` and `className="gridItem"` in the `GridItem` component, it somehow didn't render properly. The rendered div was missing many class names even if I tried adding `className="gridItem"` onto the `GridItem` component, so I guess the children of `ResponsiveGridLayout` have to be directly `Boxes` with certain classnames?

I was also previously trying to figure out how to get data from the public calendar, but I commented it all out, because I couldn't get it to work yet.

Anyway, give it a try, perhaps clean the local storage first and check if everything is saving properly.


packages/web/pages/dashboard.tsx
```diff
@@ -37,29 +35,6 @@ export interface ContainerQueries {
...
-
-export const gridDataSm = [
-  { i: 'latest', x: 0, y: 3, w: 4, h: 3 },
-  { i: 'xp', x: 0, y: 0, w: 2, h: 2 },
-  { i: 'seed', x: 2, y: 0, w: 2, h: 2 },
-  { i: 'cal', x: 0, y: 5, w: 2, h: 4 },
-  { i: 'leaders', x: 2, y: 5, w: 2, h: 4 },
-];
 
 export const initLayouts = {
```
luxumbra, line 39: Out of curiosity, why not move `initLayouts` out to `utils/dashboard.tsx` with the other layout constants and just have the one import for `initLayouts`?

luxumbra (commented):
Looks good. All seems to be working as it was, with no additional bugs. 👍🏻 

> If I wanted to include the Box with the key and className="gridItem" in the GridItem component, it somehow didn't render properly. The rendered div was missing many class names even if I tried adding className="gridItem" onto the GridItem component, so I guess the children of ResponsiveGridLayout have to be directly Boxes with certain classnames?

Yeah, I had that issue. I wonder if using `forwardRef` or something to pass the required props from the `GridLayout` to `GridItem` would make this work and also potentially fix the height bug we have. I did look at using forwardRef but my brain hurt, so moved on to Container Queries and adding stuff direct to the main file. 😅

packages/web/pages/dashboard.tsx
vidvidvid: good remark, we can put at least all of the static values to the `utils/dashboard.tsx`, i missed this one. :)

vidvidvid:
> Yeah, I had that issue. I wonder if using `forwardRef` or something to pass the required props from the `GridLayout` to `GridItem` would make this work and also potentially fix the height bug we have. I did look at using forwardRef but my brain hurt, so moved on to Container Queries and adding stuff direct to the main file. 😅
We could save this as one of the possible improvements after we're done building the mvp.

Issue #888, opened 2021-10-30
Navigating back and forth between players doesn't update profile data


vidvidvid:
### What happened?

Navigating back and forth between players doesn't update profile data.
If the user clicks on another player's profile and navigates to their own profile from the menu by clicking View Profile and then uses browser navigation to navigate back to the previously viewed profile, the sections of the profile don't get updated.
(e.g. player/luxumbra -> player/vid -> player/luxumbra...)

### What did you expect to happen?

The data should update according to the user profile.

### How can we reproduce the problem (as minimally as possible)?

1. click on another player's profile 
2. click View Profile in menu 
3. navigate back to the previously viewed profile

### Browser and operating system used (if applicable)

Preview of the bug:
![navigate](https://user-images.githubusercontent.com/35112939/139543374-0a5ed2af-c38a-46e4-ab82-b7b3feee7a59.gif)

Udit-takkar:
yeah even when i open any profile from home . there is an edit profile button

Udit-takkar:
sometime the sections are not even updating 
<img width="1440" alt="Screenshot 2021-10-31 at 9 26 51 PM" src="https://user-images.githubusercontent.com/53316345/139591875-f033c333-b2a9-45ea-985e-2a699ca0a5f9.png">

vidvidvid:
that's the point of this issue. as the title suggests - navigating back and forth doesn't update the profile data (which is displayed in sections). i also noticed that the edit button remains. not good.

PR #892, opened 2021-11-01
fix: navaigating between profiles and edit button

Udit-takkar:
#888


vidvidvid (approved):
nice catch, thank you.

alalonde:
In the future, please push PR branches to the `MetaFam` org so that the CI builds happen automatically.

alalonde (approved):

Issue #337, opened 2021-02-09
[MetaOS] - Build The Dashboard


peth-yursick:
MetaGame needs a dashboard!
With the creation of the new metamenu, it became obvious metagame needs a "home" page as well.
Not the landing page - though that is needed as well...
The home page aka the dashboard where people can see:
- Their Seeds & XP stats as well as other MetaGame stats
- Latest posts, podcasts & videos
- Upcoming events
- The leaderboard
- ???

[The design](https://www.figma.com/file/A6Nh2tyBbRCpglAnm7JQDm/Mega-Menu?node-id=86%3A107)
The actual layout doesn't matter & will be updated.
Mainly what matters for now is the backend for this stuff to work.
Like plugging in google calendar, a way for pulling in the latest content, embedding the leaderboard, measuring "morale" etc.
Some of these things are relatively simple in how they should work, some will require a bunch of discussion.

Eventually, the thing should be modular so people should be able to customize, move the modules around & embed other dapps or things they want easily accessible through their dashboards.

Want to work on it?
Comment here, feel free to ask anything.

davort:
Hi, I am the designer behind these mockups. Feel free to ping me about anything regarding the Dashboard. Looking forward to talking and working with you!

vidvidvid:
After I'm done with the user profile / edit update, I would like to work on this.

PR #900, opened 2021-11-04
feature: Dashboard section for Seed done for mvp.

luxumbra:
Adds CoinGecko API to handle the token data & React Vis library for adding charts in SVG. 

I also added maxH on the grid for the the seed & xp sections as under 2 rows high, they aren't ideal.  

Relates to #337 & #691


packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+
+type MarketProps = {
+  identifier: string;
+};
+
+export const Seed: FC = () => {
+  const [token, setToken] = useState<TokenDataProps | null>();
+
+  useEffect(() => {
+    if (typeof token !== 'undefined') return;
```
mquellhorst, line 71: Prob cleaner to set token null by default and check whether it's not null here

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+  const [scale, setScale] = useState<boolean>(true);
+
+  const toggleScale = () => {
+    setScale(!scale);
+  };
+
+  function makePlots(days: Array<Array<number>>) {
+    // adding this as i couldn't work out how to fix the type error. Could do with a pair session for this stuff.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const plots: Array<string | any | AreaSeriesPoint> = [];
```
mquellhorst, line 220: String and any should be removed here

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+        <IconButton
+          aria-label="Toggle chart scale"
+          icon={<FaChartBar />}
+          borderColor="pinkShadeOne"
+          background="rgba(17, 17, 17, 0.9)"
+          color="pinkShadeOne"
+          _hover={{ color: 'white', borderColor: 'white' }}
+          variant="outline"
+          isRound
+          onClick={() => toggleScale()}
```
mquellhorst, line 248: just putting toggleScale between the curly braces should be enough

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+      </Box>
+      <Box
+        className="chartWrapper"
+        position="absolute"
+        width="100%"
+        height="100%"
+        bottom={0}
+        left={0}
+        zIndex={0}
+        sx={{
```
mquellhorst, line 155: I guess it would clean things up a bit more if you'ld put these styles in a const somewhere and just put the const here

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+            prices,
+          };
+
+          if (!token) {
+            return tokenData;
+          }
+          return null;
+        };
+        setToken(() => buildTokenData());
+        return () => {};
```
mquellhorst, line 107: doesn't need a return actually

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+              market_data.total_volume,
+            ).toFixed(2),
+            highLow7d: findHighLowPrice(prices, 7),
+            prices,
+          };
+
+          if (!token) {
+            return tokenData;
+          }
+          return null;
```
mquellhorst, line 104: just return is enough I think no null needed

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+        );
+        const chartJson = await chartResponse.json();
+
+        const { market_data, tickers } = tokenJson;
+        const { prices, total_volumes } = chartJson;
+
+        const buildTokenData = () => {
+          const tokenData: TokenDataProps = {
+            market: market_data,
+            ticker: ticker(tickers, 'balancer_v1'),
```
mquellhorst, line 89: I'ld move all these function to the top assign the values to const and construct the json object

packages/web/components/Dashboard/Seed.tsx
```diff
@@ -1,36 +1,303 @@
...
+        const tokenJson = await tokenResponse.json();
+        const chartResponse = await fetch(
+          `${apiUrl}coins/${tokenId + chartQuery}`,
+        );
+        const chartJson = await chartResponse.json();
+
+        const { market_data, tickers } = tokenJson;
+        const { prices, total_volumes } = chartJson;
+
+        const buildTokenData = () => {
```
mquellhorst, line 86: you can just take everything out of the function and pass the json object to setToken

mquellhorst (changes requested):

alalonde (approved):
Nice! Looks good overall. Some comments:

1. The coingecko logo (for me) is hard-enough-to-see that it's like it's not there. Not sure what they dictate in their Terms of Service, but something like "Powered by CoinGecko" in an (i) Icon or something might look better.
2. React-spring looks pretty snazzy but, to your point, seems overweight. I think it's probably fine to leave in there for the being, assuming next.js / webpack does sufficient tree-shaking or other optimizations to thin it down.
3. Chart animation looks great! My only nitpick is the fade from white / purple / back to white on the xxD text looks a bit odd.

alalonde:
@mquellhorst Have your suggestions / comments been addressed? Would like to move this one through as it's a prerequisite for a bunch of other stuff

alalonde (approved):

luxumbra:
Thanks for the Merge @alalonde 🙏🏻

PR #901, opened 2021-11-04
feature: Dashboard leaderboard section MVP

luxumbra:
Leaderboard is currently only showing the 7 most active players for the season. All styling is mostly done but needs to show the top 10 maybe?

### DIFF
--- a/packages/design-system/src/SelectSearch.tsx
+++ b/packages/design-system/src/SelectSearch.tsx
@@ -98,3 +98,45 @@ export const selectStyles: Styles = {
 export const SelectSearch: React.FC<SelectProps> = (props) => (
   <Select styles={selectStyles} {...props} />
 );
+
+export const metaFilterSelectStyles: typeof selectStyles = {
+  ...selectStyles,
+  multiValue: (s) => ({
+    ...s,
+    color: theme.colors.white,
+  }),
+  multiValueLabel: (s) => ({
+    ...s,
+    color: theme.colors.white,
+  }),
+  groupHeading: (s, { children }) => ({
+    ...s,
+    ...(selectStyles.groupHeading &&
+      selectStyles.groupHeading(s, { children })),
+    borderTop: `1px solid ${theme.colors.borderPurple}`,
+    margin: 0,
+  }),
+  option: (s, { isSelected }) => ({
+    ...s,
+    backgroundColor: 'transparent',
+    fontWeight: isSelected ? 'bold' : 'normal',
+    ':hover': {
+      backgroundColor: 'transparent',
+      color: theme.colors.white,
+    },
+    ':focus': {
+      boxShadow: '0 0 0 3px rgba(66, 153, 225, 0.6)',
+    },
+  }),
+  menu: () => ({}),
+  control: (s) => ({
+    ...s,
+    background: theme.colors.dark,
+    border: 'none',
+    ':hover': {},
+  }),
+  noOptionsMessage: (s) => ({
+    ...s,
+    borderTop: `1px solid ${theme.colors.borderPurple}`,
+  }),
+};
--- a/packages/design-system/src/index.ts
+++ b/packages/design-system/src/index.ts
@@ -14,7 +14,12 @@ export { MetaTag } from './MetaTag';
 export { MetaTile, MetaTileBody, MetaTileHeader } from './MetaTile';
 export { MultiSelect } from './MultiSelect';
 export { ResponsiveText } from './ResponsiveText';
-export { SelectComponents, SelectSearch, selectStyles } from './SelectSearch';
+export {
+  metaFilterSelectStyles,
+  SelectComponents,
+  SelectSearch,
+  selectStyles,
+} from './SelectSearch';
 export {
   getTimezonesFor,
   SelectTimeZone,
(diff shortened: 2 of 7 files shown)


vidvidvid:
2 concerns:
The order of the leaderboard looks a bit disorganised, why is that?
Screenshot 2021-11-04 at 19 47 53

There is a repeating error in the console that appears when `/dashboard` is loaded:
Screenshot 2021-11-04 at 19 53 40

luxumbra:
> 2 concerns: The order of the leaderboard looks a bit disorganised, why is that?

It is because the playerList query defaults to the seasonal leaderboard. The filter is missing here. It'll take me forever to add them in with my typescript skills, so feel free to do this. 

> There is a repeating error in the console that appears when `/dashboard` is loaded

Not sure about this but will take a look when I have refactored some of the Seed section.

vidvidvid:
> It is because the playerList query defaults to the seasonal leaderboard. The filter is missing here. It'll take me forever to add them in with my typescript skills, so feel free to do this.

Could you maybe than get the seasonal xp for the leading players and display that xp instead? We can always add filters later and other selections later.

> Not sure about this but will take a look when I have refactored some of the Seed section.

👍

Udit-takkar:
https://user-images.githubusercontent.com/53316345/142258002-74f55f31-e712-4fac-970b-4b0a719cd93c.mov

I got sorting options working

vidvidvid:
> Need to fix some parts of UI. But there is also an error
> 
> ```
> Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
> ```
> 
> When the leaderboard data is being fetched and till it is displayed for some reason the leaderboard component is rendering a lot

how's fixing the mentioned things coming along? need any help?

Udit-takkar:
yeah . I have tired commenting the things you said earlier (like onScreen,nextPage() etc. which are not used in this page but included) . Can't resolve this bug

Udit-takkar:
so this useEffect is causing problem in usePlayerFilter hook
```
  useEffect(() => {
    if (error || fetching) {
      if (error || !shouldAppend.current) {
        setPlayers([]);
      }
    } else if (shouldAppend.current) {
      shouldAppend.current = false;
      setPlayers((p) => [...p, ...fetchedPlayers]);
    } else {
      setPlayers(fetchedPlayers);
    }
  }, [fetchedPlayers, fetching, error]);
  ```

**Comment by Udit-takkar** [comment]
This useEffect is causing multiple re renders in usePlayerFilter hook

```
 useEffect(() => {
    if (error || fetching) {
      if (error || !shouldAppend.current) {
        setPlayers([]);
      }
    } else if (shouldAppend.current) {
      shouldAppend.current = false;
      setPlayers((p) => [...p, ...fetchedPlayers]);
    } else {
      setPlayers(fetchedPlayers);
    }
  }, [fetchedPlayers, fetching, error]);

```
If i remove fetchedPlayer from useEffect. it does not cause multiple re renders

**Comment by vidvidvid** [comment]
you found the right `useEffect` that was causing the trouble, but we can't unsubscribe from `fetchedPlayers`, because then we would not get any new players from scrolling on the `/players` site. the problem was with this piece of code that caused constant rerendering:
```
if (error || !shouldAppend.current) {
  setPlayers([]);
}
```
I believe it set up new players array, which in turn triggered a rerender and refetching of `fetchedPlayers` from `usePaginatedPlayers`, which triggered the useEffect hook again.

**Comment by vidvidvid** [comment]
I adjusted the styling of the filter popover and added a spinned when the data is loading:

### Before:
Screenshot 2021-11-25 at 16 27 14

**Comment by Udit-takkar** [comment]
> I adjusted the styling of the filter popover and added a spinner when the data is loading.

I was about to do it. thanks for stepping in and finishing it

**Comment by vidvidvid** [comment]
my pleasure :) i turned this draft into a pr, it should be ready for the review now.
this logic could be moved out of the rendering part of the component:
```
players.map((p, i) => {
  const playerPosition: number = i + 1;
  if (
    playerPosition <= 7 &&
    ((showSeasonalXP && p.season_xp >= 1) ||
      (!showSeasonalXP && p.total_xp >= 50))
  ) {
```
what do you think?

vidvidvid:
What is the status on this branch?

Udit-takkar:
LGTM. Just rebase and remove conflicts
Might have to ask @luxumbra  if anything has to be included?

luxumbra:
> LGTM. Just rebase and remove conflicts Might have to ask @luxumbra if anything has to be included?

LGTM too. Nice job. I don't think it needs anything else included. There's probably going to be some responsive tweaks with the media & container queries once we've got all the sections done for v1.

alalonde:
Can we remove the availability / username selections from the dropdown? They don't make sense in this context.

alalonde:
> Can we remove the availability / username selections from the dropdown? They don't make sense in this context.

Could one of @luxumbra @vidvidvid @Udit-takkar  take this on? I'm assuming it's pretty straightforward. LMK if not.

Udit-takkar:
Rebased and fixed conflicts

alalonde (approved):

alalonde:
Great team effort!

PR #902, opened 2021-11-04
feature: Dashboard calendar section MVP 

vidvidvid:
### Added the upcoming MG calendar events
~They include:~
~- title~
~- day & time~
~- description~
~- guests~
~- location~
Added the default iframe version of google calendar due to the lack of a smooth, flexible solution for implementing a public calendar.

### Implementation
~I had major problems when trying to make use of the service google account in order to fetch the calendar data, since fetching any kind of google data that is connected to accounts seems to require some sort of authentication. After a while of trying and researching, I found a way that bypasses the need for authentication:~
~I created a google sheet and a script that gets the data from the MetaGame calendar and fills out the fields in the sheet. Then I published the sheet as csv, fetched it in the web app and parsed the data. The data is being updated hourly.~

Google sheet of the upcoming 5 events: ~https://docs.google.com/spreadsheets/d/1Pb7Jxa_snve5UoDstGoR4PI5MdCPKUOWcZM_p0kx6FU/edit#gid=0~
~Script: https://script.google.com/home/projects/1mwxmZ3lX0A_TbD1-cAoQHqgCPR7DXi1PLXCmhel6-~UNWZqOD0SWB8jw3/edit
~Documentation for the CalendarApp: https://developers.google.com/apps-script/reference/calendar~
~Can someone confirm they can access the script?~

### Improvement ideas
~If someone will be keen to refactor this and try their hand on implementing the calendar with google service account, I wish you good luck but until then, I think this will work just fine.~
@polimyl suggested we could use the ics of the calendar, which would update real-time and provide an overall more elegant solution 👍.

### Preview
calendar


vidvidvid:
@polimyl i tried the ical solution and fetching the data again requires authorisation. google really seems to want to know exactly who is fetching what data and they are making it as difficult as possible to fetch the data with a service account. after days/hours spent on trying to get it to work, i finally gave up and implemented the ugly default iframe version of the calendar, which, on the bright side, requires just the calendar to be public and its `id`, so other organisations that will eventually pick up our platform will have a really easy time to adjusting it to their calendar.

vidvidvid:
Hail the default embeddable google calendar:
Screenshot 2021-11-15 at 21 52 51

polimyl:
oh god, please not the embed.
didnt see this earlier, but the ics is indeed publicly available.
metagame ics: https://calendar.google.com/calendar/ical/nih59ktgafmm64ed4qk6ue8vv4@group.calendar.google.com/public/basic.ics

vidvidvid:
@polimyl it can't be fetched, I'm getting the cors error and I couldn't get it to work with service account.

polimyl:
hmm, i can curl it fine though. maybe fetch and cache it from the backend or hasura every x minutes to work around cors?

alalonde (approved):
This does add useful functionality, so I will merge as-is. Let's implement the backend fetching in another PR. Presumably if we are fetching via hasura we can just use the google calendar REST API (or graphql! if they have one) for a cleaner implementation

alalonde:
@vidvidvid When we implement the v2 version, can we display the times in the local timezone? Maybe by passing the user's timestamp into the fetch query ?

vidvidvid:
> @vidvidvid When we implement the v2 version, can we display the times in the local timezone? Maybe by passing the user's timestamp into the fetch query ?

Yep, for sure. Even in this version I made sure to display the times based on the user's timezone and I'm sure we can find a way to display it after we manage to fetch the data.

PR #910, opened 2021-11-10
Docker image size optimization

### LINKED ISSUE #855: Optimize Frontend and Backend Dockerfiles
Both the Frontend and Backend Dockerfiles are pretty unoptimized at the moment. The built containers have a size of nearly 1GB, which is huge (the Hasura container has a size of less than 70MB for reference). The sizes could probably be drastically reduced and maybe even the build time could be lowered.

polimyl:
Fixes #855 
I have tried to optimize our Dockerfiles since the built images were huge. Sadly couldn't really optimize build time since I'm not experienced enough with yarn etc. and our eventual upgrade to yarn 2 could probably already decrease build time quite a bit.

Some data (single local build per entry, all builds from completely fresh state)
|Name|Old Image Size|New Image Size|Old Build Time|New Build Time|
|---|---|---|---|---|
|frontend|3.2 GB|683 MB|10m13s|10m43s|
|backend|3.11 GB|683 MB|8m35s|6m38s|
|discord-bot|2.08 GB|270 MB|4m17s|4m17s|

-> Overall final image size reductions of ~80% with similar build times

I have only done simple smoke tests with the new images though.

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -34,6 +34,9 @@ COPY packages/@types ./packages/@types/
 # Build
 RUN yarn backend:build
 
+# Delete devDependencies
+RUN yarn install --pure-lockfile --production --ignore-scripts --prefer-offline
+
 # Create completely new stage including only necessary files
 FROM node:14-alpine as app
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


dysbulic (approved):
This looks reasonable to me given my limited Docker knowledge…

Issue #793, opened 2021-08-27
Potentially replace react-timezone-select


dysbulic:
We are in the process of [integrating IDX](https://github.com/MetaFam/TheGame/issues/769) more deeply into MyMeta and there's the possibility we will [upgrade Next](https://github.com/MetaFam/TheGame/issues/792) as part of the process.

The timezone select component [requires transpiling](https://github.com/ndom91/react-timezone-select#-nextjs-users) to work with Next 10. There's a question if it will need that in v11 & if it does if it will work.

@dan13ram, what does the timezone select component do other than have a list of timezones & provide text search on the timezone field.

If it would be expedient to replace it, could we make do with a generic text search component rather than something timezone specific?

tommymaher15:
Apparently there has been an update on react-timezone-select which was as follows: 

"Update: Next.js 11.1.0 now supports ESM packages natively! Until 12.0.0 you need to include esmExternals: true in your next.js.config though."

is this worthwhile to look at or would it make more sense to continue and replace with a generic search as mentioned above?

mquellhorst:
Yeah is a good one to take a look at.
Currently the timezone package is being transpiled in `next.config.js`
Seems likely this solved the roadblock to upgrade of Next.js

richteri:
Hi Guys, I'm working on the [Next 11 & React 17 upgrade ticket](https://github.com/MetaFam/TheGame/issues/792) and tried to use the `esmExternals` option with & without transpiling modules.

The first issue was that `next` does not allow [global CSS imports](https://github.com/vercel/next.js/issues/19936) but seems to be fixed by using [next-global-css](https://github.com/bem/next-global-css).

Currently I can't get over the issue below:

```
@metafam/ds: - Compiling modules...
@metafam/web: [0] error - ../design-system/dist/ds.esm.js:10:0
@metafam/web: [0] Module not found: Package path ./dist/index.js is not exported from package /home/irichter/dev/TheGame/node_modules/react-timezone-select (see exports field in /home/irichter/dev/TheGame/node_modules/react-timezone-select/package.json)
@metafam/web: [0]    8 | import cityTimezones from 'city-timezones';
@metafam/web: [0]    9 | import TimezoneSelect from 'react-timezone-select';
@metafam/web: [0] > 10 | import { i18nTimezones } from 'react-timezone-select/dist/index.js';
@metafam/web: [0]   11 | import spacetime from 'spacetime';
[...]

Issue #792, opened 2021-08-27
[MyMeta] Upgrade to Next 11 & React 17


dysbulic:
@Cali93 & I would like to [web3.storage](//web3.storage) to house the [profile, header](https://github.com/MetaFam/TheGame/issues/27), and [background](https://github.com/MetaFam/TheGame/issues/443) images being added by users of the in-house front end to IDX.

We are having difficulty getting it to work and all signs point to it would likely work with the latest versions of Next and React.

React 17 has been out for almost a year, and it's useful in general to be using the current stable release.

@mquellhorst was in the programming session as well and concurred this seems like the most expedient path.

mquellhorst:
Involves #793

dysbulic:
[IDX](//idx.xyz) is being deprecated in favor of [Glaze](https://developers.ceramic.network/tools/glaze/overview/) which also requires Webpack 5 and motivates us to do this update.

mquellhorst:
React has been updated.
NextJS can be updated to version 12 but needs testing.

alalonde:
React, yes, Next, no.

peth-yursick:
> React, yes, Next, no.

looks like rich changed his mind lol
so the issue should remain open?
if so, should it be put on the priorities list on dework or kept here for now?

alalonde:
I'll leave that answer to @dysbulic, not sure what has changed since he created this. Probably no to dework though

dysbulic:
I'm upgrading to React 18 and Next 12 as part of #1405.

peth-yursick:
> I'm upgrading to React 18 and Next 12 as part of #1405.

Nice nice, closing this then

PR #916, opened 2021-11-12
Update to Webpack 5

dysbulic:
This pull request updates the version of Webpack used for building the website to v.5.

It was not necessary to upgrade Next.js (#792), nor was it necessary to replace the timezone select (#793), only to remove an erroneous type definition for that library.


alalonde (approved):
Nice work!

PR #927, opened 2021-11-18
Dashboard latest news: listen section

vidvidvid:
!! to the reviewer: **Please make sure that the #900 is merged to develop before approving and merging this pr** !!

## Overview

Dashboard latest news listening section.

## Improvement ideas

For future reference the player can be [styled with css](https://chromium.googlesource.com/chromium/blink/+/72fef91ac1ef679207f51def8133b336a6f6588f/Source/core/css/mediaControls.css?autodive=0%2F%2F%2F). 

## To-do

- [x] ~change the library to [react-audio-player](https://www.npmjs.com/package/react-audio-player)~
- [x] use default html `<audio>` player
- [x] show more/show less button
- [x] make everything pretty
- [x] refactor fetching of data
- [x] use swr package for client side data fetching

## Assets

**Preview**
dashboard_listen_section

### DIFF
--- a/packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
+++ b/packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
@@ -1,12 +1,19 @@
 import 'react-h5-audio-player/lib/styles.css';
 
-import { Box, Heading, Text } from '@metafam/ds';
+import { Box, Button, Heading, Text } from '@metafam/ds';
 import React, { useState } from 'react';
 import AudioPlayer from 'react-h5-audio-player';
 import { parse } from 'rss-to-json';
 
 export const Listen: React.FC = () => {
-  const [items, setItems] = useState([]);
+  const [items, setItems] = useState<
+    Array<{
+      title: string;
+      src: string;
+      showMore: boolean;
+      description: string;
+    }>
+  >([]);
 
   const getData = async () => {
     const podcastDataRss = await parse(
@@ -14,27 +21,48 @@ export const Listen: React.FC = () => {
       {},
     );
 
-    console.log('podcast data', podcastDataRss);
-    setItems(podcastDataRss.items);
-    console.log('items', items);
+    const preparedData = podcastDataRss.items.map((item) => ({
+      title: item.title,
+      description: item.description.replace(/<\/?[^>]+(>|$)/g, ''),
+      src: item.enclosures[0].url,
+      showMore: false,
+    }));
+    setItems(preparedData);
+  };
+
+  const findAndReplace = (src: string) => {
+    const foundIndex = items.findIndex((x) => x.src === src);
+    items[foundIndex].showMore = !items[foundIndex].showMore;
+    return items;
   };
 
   if (items.length === 0) getData();
-  console.log('items', items);
 
   return (
     <Box>
       {items.map((item) => (
         <Box>
-          <Heading size="xs" color="white">
-            {item.title}
-          </Heading>
-          <Text>{item.description.replace(/<\/?[^>]+(>|$)/g, '')}</Text>
           <AudioPlayer
             // autoPlay
+            header={
+              <Heading size="xs" color="black">
+                {item.title}
+              </Heading>
+            }
+            footer={
+              <Box>
+                <Text>
+                  {item.showMore
+                    ? item.description
+                    : `${item.description.substring(0, 250)}`}
+                </Text>
+                <Button onClick={() => setItems([...findAndReplace(item.src)])}>
+                  Show more
+                </Button>
+              </Box>
+            }
             preload="none"
-            src={item.enclosures[0].url}
-            onPlay={(e) => console.log('onPlay')}
+            src={item.src}
             // other props here
           />
         </Box>
--- a/packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
+++ b/packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
@@ -1,5 +1,5 @@
-import { Box, Heading, Text } from '@metafam/ds';
-import React, { useState } from 'react';
+import { Box, Heading, LoadingState, Text } from '@metafam/ds';
+import React, { useEffect, useState } from 'react';
 import { parse } from 'rss-to-json';
 
 export const Listen: React.FC = () => {
@@ -12,20 +12,33 @@ export const Listen: React.FC = () => {
     }>
   >([]);
 
-  const getData = async () => {
-    const podcastDataRss = await parse(
-      'https://anchor.fm/s/57a641c/podcast/rss',
-      {},
-    );
+  const [isLoading, setIsLoading] = useState(true);
 
-    const preparedData = podcastDataRss.items.map((item) => ({
-      title: item.title,
-      description: item.description.replace(/<\/?[^>]+(>|$)/g, ''),
-      src: item.enclosures[0].url,
-      showMore: false,
-    }));
-    setItems(preparedData);
-  };
+  useEffect(() => {
+    const getData = async () => {
+      try {
+        const podcastDataRss = await parse(
+          'https://anchor.fm/s/57a641c/podcast/rss',
+          {},
+        );
+
+        const podcasts = podcastDataRss.items.map((item) => ({
+          title: item.title,
+          description: item.description.replace(/<\/?[^>]+(>|$)/g, ''),
+          src: item.enclosures[0].url,
+          showMore: false,
+        }));
+
+        setIsLoading(false);
+        setItems(podcasts);
+      } catch (error) {
+        console.log(error);
+        setIsLoading(false);
+      }
+    };
+
+    getData();
+  }, []);
 
   const findAndReplace = (src: string) => {
     const foundIndex = items.findIndex((x) => x.src === src);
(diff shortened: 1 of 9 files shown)


Udit-takkar:
I don't get why are you fetching data with 
```
 if (items.length === 0) getData();
 ```
 
instead of getStatic props or something
any reason?

**Comment by vidvidvid** [comment]
i just set it up so i could get started working with data, i'll make one final cleanup and fix everything when i'm ready to turn this draft into an actual pull request 👍

**Comment by vidvidvid** [comment]
@Udit-takkar apparently `getStaticProps`/`getServerSideProps` can be used only on `next.js` pages and the `<Listen />` section where i was fetching the data is just a react component. I decided to fetch the data inside the component's useEffect, because fetching in the `dashboard.tsx` (with `getStaticProps`, for example) would mean a long chain of prop passing and using context to pass it to the `<Listen />` component doesn't really make sense, because other child components of dashboard don't need it. However, in the future, when we implement the 'all section' it might make sense to fetch all the listening, reading and watching data in the dashboard and pass it to the their respective sections, but we still need to decide on how how we'll implement this and I don't want to make any assumptions until we make a choice together.

**Comment by Udit-takkar** [comment]
> @Udit-takkar apparently `getStaticProps`/`getServerSideProps` can be used only on `next.js` pages and the `<Listen />
 Well then i think we should use `useSwr` for client side fetching. much better option than useEffect

**Comment by vidvidvid** [comment]
Why do you think it's better?

**Comment by vidvidvid** [comment]
Ahh, i see - it first provides the data from the cache and then makes the call and patches any necessary changes. Nice stuff. We could probably make an issue to refactor all (or most) of our client side fetching with [swr package](https://www.npmjs.com/package/swr).

**Comment by Udit-takkar** [comment]
Yeah sure

File: packages/web/components/Dashboard/LatestContentTabs/Listen.tsx [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+
+  const findAndReplace = (src: string) => {
+    const foundIndex = items.findIndex((x) => x.src === src);
+    items[foundIndex].showMore = !items[foundIndex].showMore;
+    return items;
+  };
+
+  return (
+    <Box pt={4}>
+      {isLoading ? (
```
dysbulic, line 49:
```suggestion
      {loading ? (
```

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+              {item.title}
+            </Heading>
+            <Text
+              textOverflow={item.showMore ? '' : 'ellipsis'}
+              overflow={item.showMore ? '' : 'hidden'}
+              whiteSpace={item.showMore ? 'normal' : 'nowrap'}
+              fontSize="sm"
+              cursor="pointer"
+              color="white"
+              onClick={() => setItems([...findAndReplace(item.src)])}
```
dysbulic, line 70: I'm surprised this works. You're mutating `items` *then* making a copy of it. Maybe because you're setting it immediately you don't see the side effects…

```suggestion
              onClick={() => setItems((items) => (
                toggleVisibility({
                  items: [...items], member: item.src
                })
              ))}
```

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+            >
+              {item.description}
+            </Text>
+            <Text
+              cursor="pointer"
+              color="blueLight"
+              pt={1}
+              pb={1}
```
dysbulic, line 78:
```suggestion
              py={1}
```

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+              fontWeight="bold"
+              onClick={() => setItems([...findAndReplace(item.src)])}
```
dysbulic, line 80:
```suggestion
              onClick={() => setItems((items) => (
                toggleVisibility({
                  items: [...items], member: item.src
                })
              ))}
```

packages/web/pages/join/guild/[guildname].tsx
```diff
@@ -20,26 +21,28 @@ const SetupGuild: React.FC = () => {
...
+  }, [data, guildName]);
 
   if (guild == null) {
     return <></>;
   }
 
-  const onSubmit = async (data: EditGuildFormInputs) => {
-    const { type, ...otherInputs } = data;
+  const onSubmit = async (editGuildFormInputs: EditGuildFormInputs) => {
+    const { type, ...otherInputs } = editGuildFormInputs;
```
Comment (dysbulic) [inline packages/web/pages/join/guild/[guildname].tsx:37]:
```suggestion
    const {
      type,
      discordAdminRoles: adminRoles,
      discordMembershipRoles: membershipRoles,
      ...otherInputs
    } = editGuildFormInputs;
```

packages/web/pages/join/guild/[guildname].tsx
```diff
@@ -20,26 +21,28 @@ const SetupGuild: React.FC = () => {
...
     const payload: GuildInfo = {
       ...otherInputs,
-      discordAdminRoles: data.discordAdminRoles.map((o) => o.value),
-      discordMembershipRoles: data.discordMembershipRoles.map((o) => o.value),
+      discordAdminRoles: editGuildFormInputs.discordAdminRoles.map(
```
Comment (dysbulic) [inline packages/web/pages/join/guild/[guildname].tsx:40]:
```suggestion
      discordAdminRoles: adminRoles.map(
```

packages/web/pages/join/guild/[guildname].tsx
```diff
@@ -20,26 +21,28 @@ const SetupGuild: React.FC = () => {
...
+        (o) => o.value,
+      ),
+      discordMembershipRoles: editGuildFormInputs.discordMembershipRoles.map(
```
Comment (dysbulic) [inline packages/web/pages/join/guild/[guildname].tsx:43]:
```suggestion
      discordMembershipRoles: membershipRoles.map(
```

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+      setIsLoading(false);
+    }
+  }, [data]);
+
+  const findAndReplace = (src: string) => {
+    const foundIndex = items.findIndex((x) => x.src === src);
+    items[foundIndex].showMore = !items[foundIndex].showMore;
+    return items;
+  };
```
dysbulic, line 45:
```suggestion
  const toggleVisibility = ({ items, member }) => {
    const index = items.findIndex((x) => x.src === member);
    items[index].showMore = !items[index].showMore;
    return items;
  };
```

dysbulic (changes requested):
Mostly just had some formatting and conciseness comments.

The only out-and-out error is you are mutating `items` in `findAndReplace` before you make a copy of it.

dysbulic:
I'm currently not seeing any output on the [Vercel test instance](https://the-game-git-feature-dashboard-latest-newslisten-metafam.vercel.app/). Also, I think you should keep the HTML tags in the description and use [`dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml). The block of text you've got currently is a pain to parse.

https://user-images.githubusercontent.com/181523/143087070-085197af-946f-4ded-a92e-0c8a94b2472b.mp4

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+      src: string;
+      showMore: boolean;
+      description: string;
+    }>
+  >([]);
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  const { data, error } = useSWR(
+    'https://anchor.fm/s/57a641c/podcast/rss',
```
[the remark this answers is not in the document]
vidvidvid: the way dashboard config is currently structured, we just import each exported variable separately like `import { podcastRSSURL } from '../config';` - is that fine or do you think we should refactor the config?

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
```diff
@@ -0,0 +1,101 @@
...
+  if (error) {
+    console.log(error);
+    setIsLoading(false);
+  }
+
+  useEffect(() => {
+    if (data) {
+      data.items = data.items.map((item) => ({
+        title: item.title,
+        description: item.description.replace(/<\/?[^>]+(>|$)/g, ''),
```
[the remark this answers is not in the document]
vidvidvid: tbh i just took a regex i found somewhere that looked liked it successfully stripped html off of the description - didn't put much effort into testing. thx for the suggestion!

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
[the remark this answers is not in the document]
dysbulic: Sounds good to me. I was just wanting the configuration value moved to the config file & `CONFIG.whatever` is the format used everywhere else.

There's a `"include": ["./src"]` in the TytpeScript config, so you should be able to reference dependencies from the root without a preceding path (i.e. `import from 'config').

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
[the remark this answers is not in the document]
dysbulic: I later had a change of heart and think you shouldn't be stripping the tags at all, but instead `dangerouslySetInnerHTML`.

¿What's the motivation for removing them?

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
vidvidvid: i couldn't go exactly for your solution, because i was getting `Assignment to property of function parameter 'items'.eslintno-param-reassign` at `items[index].showMore = !items[index].showMore;`, so I made a slight adjustment:
```
setPodcasts((items) =>
  toggleVisibility({
    items, <-- no need to make a new array here, because it's made in toggleVisibility
    src: podcast.src,
  }),
)
```
and `toggleVisibility`:
```
const itemsCopy = [...items];
const index = items.findIndex((item) => item.src === src);
itemsCopy[index].showMore = !items[index].showMore;
return itemsCopy;
```

vidvidvid:
I thought about using `dangerouslySetInnerHTML`, but something in the name made me avert from that idea :P However, if you think we could use it - sure, why not, it's definitely going to look nicer!

> The block of text you've got currently is a pain to parse.

What do you mean by that?

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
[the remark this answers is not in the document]
vidvidvid: It's because I didn't want to use `dangerouslySetInnerHTML` and just show the description as a block of text. Although it would probably look nicer, this is a potential vulnerability (since any script can be executed in it). So what do you think - do we risk it?

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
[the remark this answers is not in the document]
vidvidvid: thanks, noted for the future!

dysbulic:
> > The block of text you've got currently is a pain to parse.
> 
> What do you mean by that?

The cognitive workload of picking out sentences and meaning is significantly higher then with some whitespace. Parsing being the process of making sense out of a given input — mentally in this scenario.

packages/web/components/Dashboard/LatestContentTabs/Listen.tsx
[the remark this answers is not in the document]
dysbulic: Script or style injection are really the only dangers I know of, and the source text will be generated by someone in MetaGame.

That or Anchor goes rogue, which probably isn't going to happen. 😸

vidvidvid:
> The cognitive workload of picking out sentences and meaning is significantly higher then with some whitespace. Parsing being the process of making sense out of a given input — mentally in this scenario.

aaaah ok, i see what you mean. agreed. i'll just go with innerHtml than 👍

vidvidvid:
show_more_show_less

alalonde:
Dope! My only suggestion would be to make the Latest Content / tabs fixed, and just the content scrollable. @davort thoughts on the WIP? https://the-game-git-feature-dashboard-latest-newslisten-metafam.vercel.app/

vidvidvid:
We're aiming to first fetch all of the sections' data and then make a unified tab will all of the latest news and have an option to filter out the news by categories, so the structure with tabs is only temporary: ​https://discord.com/channels/629411177947987986/815357282631155833/909543635035508776
the proposed design: https://www.figma.com/file/M3BpWO13WJVp7YivScL6K9/Dashboard?node-id=1401%3A7911
@alalonde what do you think about this idea?

alalonde:
Ahh ok, it's just a stopgap then. Can we build a somewhat unified "content" box that can be used to display video, audio OR text? I'm just curious what it's going to look like with all three media types in a single view.

Udit-takkar (changes requested):
https://user-images.githubusercontent.com/53316345/143681842-b46880df-b70d-4981-856f-feada177c452.mov

I think this is how the scrolling should would. The tabs should not be scrollable but the content

vidvidvid:
> Ahh ok, it's just a stopgap then. Can we build a somewhat unified "content" box that can be used to display video, audio OR text? I'm just curious what it's going to look like with all three media types in a single view.

Yup, it's temporary and yes, we are planning to build a unified content box. We are still figuring how exactly it will look like, so I'm curious as well. :)

> This seems like the kind of thing where we just build the content fetching with a bare-bones MVP-style UI, then have our designers review once they know what sort of data is going to be displayed (assuming this hasn't already been designed)

Yup.

> I think this is how the scrolling should look. currently on your branch the parent is scrollable but The tabs should not be scrollable but the content

As discussed on discord, we don't have to fix this exact thing atm, because the tabs will be replaced with a dropdown with filter options - https://www.figma.com/file/M3BpWO13WJVp7YivScL6K9/Dashboard?node-id=1401%3A7911. But we'd have to make sure that the dropdown with filter options stays fixed than.

alalonde:
Can you do some minimal styling on the podcast description content: e.g.
- indented list items
- proper line / paragraph spacing
- padding etc. 

I took a stab at reworking the scrolling but my CSS-fu is quite rusty these days. It would be cool to style an inner `.scroll-container` or even another component that could be placed within a grid item to scroll vertically. The calendar and content widget could use it

vidvidvid:
Did some styling on the podcast info, it looks much nicer now - thx for the feedback.
formatting

alalonde (approved):

PR #928, opened 2021-11-19
Load timezone on client

vidvidvid:
## Overview

Displaying dashboard calendar in the timezone of the user that's currently viewing it.
Remove the `jstz` library and use Intl to get the timezone name. The coverage of Intl is 94% - should we proceed with using it or should we go for a more universal package for figuring out the timezone? https://caniuse.com/?search=Intl

## Implementation

Set up a hook that's the equivalent of componentDidMount and execute the setting of timezone on client.

### DIFF
--- a/packages/web/components/Dashboard/Calendar.tsx
+++ b/packages/web/components/Dashboard/Calendar.tsx
@@ -1,8 +1,16 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 const calId = 'nih59ktgafmm64ed4qk6ue8vv4@group.calendar.google.com';
 
 export const Calendar: React.FC = () => {
+  const [isComponentMounted, setIsComponentMounted] = useState(false);
+
+  useEffect(() => setIsComponentMounted(true), []);
+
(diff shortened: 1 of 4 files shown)


packages/web/components/Dashboard/Calendar.tsx
```diff
@@ -1,10 +1,17 @@
-import jstz from 'jstz';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 const calId = 'nih59ktgafmm64ed4qk6ue8vv4@group.calendar.google.com';
```
alalonde, line 3: This should probably be in an env variable

alalonde (approved):
Simple enough

packages/web/components/Dashboard/Calendar.tsx
vidvidvid: i moved it to `packages/web/config.ts` or should it be stored somewhere else? do we store the value into `.env` file and access it through `process.env`?

alalonde (approved):

PR #930, opened 2021-11-20
fix: dao memberships without title are also displayed

### LINKED ISSUE #863: Non-mainnet DAO memberships don't render properly on the setup / memberships page
image

Udit-takkar:
## Overview

Closes #863 

## Implementation

Made all DAO membership display properly on the `/profile/setup`.
Also added Image and Url link of the dao

Note: this change won't be visible on the deployed website. you would first have to allow this page to show.
For that you would have to go to `web/lib/hooks/index`  and comment some things

## Assets

Screenshot 2021-11-20 at 5 36 20 PM

### DIFF
--- a/packages/web/components/Setup/SetupMemberships.tsx
+++ b/packages/web/components/Setup/SetupMemberships.tsx
@@ -14,11 +14,7 @@ import { Membership } from 'graphql/types';
 import React, { useState } from 'react';
 
 import { useWeb3 } from '../../lib/hooks';
-import {
-  getDaoLink,
-  getImageMoloch,
-  LinkGuild,
-} from '../Player/Section/PlayerMemberships';
+import { getDaoLink, getImageMoloch, LinkGuild } from '../Player/PlayerGuild';
 
 export type SetupMembershipsProps = {
   memberships: Array<Membership> | null | undefined;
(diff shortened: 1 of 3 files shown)


packages/web/components/Setup/SetupMemberships.tsx
```diff
@@ -78,3 +81,43 @@ export const SetupMemberships: React.FC<SetupMembershipsProps> = ({
     </FlexContainer>
   );
 };
+
+type MembershipListingProps = {
+  member: Membership;
+};
+
+const MembershipListing: React.FC<MembershipListingProps> = ({ member }) => {
```
alalonde, line 89: Let's pull this shared component up into its own file, along with the `getDaoLink`, `getImageMoloch` etc functions, so that both  `SetupMemberships` and `PlayerMemberships` can use it

alalonde (changes requested):

packages/web/components/Setup/SetupMemberships.tsx
Udit-takkar: okay makes sense

Udit-takkar:
Made the changes and rebased
LGTM

alalonde (approved):
Thanks, looks good!

PR #944, opened 2021-12-02
Changelog out of commits

vidvidvid:
## Overview

Provides a neat overview of commits that follow the conventional commits structure in the form of a changelog.

I have included all of the commits that followed the conventional commits up until now in the `0.1.0` version of the `CHANGELOG.md`.

## Improvement ideas

We could/should discuss when and how to update the changelog. I suggest we do it on each PR - meaning that we could perhaps somehow add the command to the `git push` process and always bump the version by `0.0.1`, but this would mean that everyone would have to follow the commit structure and I don't think we should enforce that on anyone. Whenever we publish the version to master, we could bump it by `0.1.0`. I can bring it up in the next builder's align.

## Assets

Repo of standard version package: https://github.com/conventional-changelog/standard-version
Conventional commits overview: https://www.conventionalcommits.org/en/v1.0.0/
Check it out: https://github.com/MetaFam/TheGame/blob/e754ba1f76b8a4707b6471a17395b2511e603d07/CHANGELOG.md


alalonde:
I'm not a huge fan of auto-updating a changelog just because a commit-by-commit granularity is generally too fine to be very useful. Nobody new to the project is going to look at the changelog to see what is going on; it will just look like a bunch of gibberish because there is a ton of assumed context that only active contributors have. 

The only functionality I see this providing is automatic categorization of commits which is mildly useful; otherwise you can just view the [commits](https://github.com/MetaFam/TheGame/commits/develop) directly. 

@mquellhorst @dysbulic What are your thoughts?

vidvidvid:
This package won't auto update the changelog. I think it would be nice to have to get a feeling of how much of what kind of work has been done in certain time period and to serve historical purposes. It might also prove useful when searching for stuff.

alalonde (approved):

PR #946, opened 2021-12-02
Show all players, even those that haven't created a profile

### LINKED ISSUE #739: Dynamically build the filter & display users without a profile
Currently, if a player hasn't configured a profile they won't appear on [MyMeta](//my.metagame.wtf). This is because the [`GetPlayers` query](//github.com/MetaFam/TheGame/blob/354dc2222855d7150e6718474424b07fad73d281/packages/web/graphql/getPlayers.ts#L20-L64) includes the line `Player_Skills: { Skill: { id: { _in: $skillIds } } }`.

The other query parameters would [match all entries if `null` (in 1.3.3, not in 2.0)](https://hasura.io/docs/latest/graphql/core/databases/postgres/queries/query-filters.html#evaluation-of-null-values-in-comparison-expressions), but the `Player_Skills: { Skill }` one requires that the user have a skill entry.

The solution according to [Stack Overflow](https://stackoverflow.com/questions/68430432/how-to-disable-a-graphql-filter-if-the-variable-is-not-set) is to pass in the complete `where` statement rather than the individual values. The generated where should only include conditional for the fields that are being queried on.

My open question is how to deal with inserting the values. Either they are escaped and inserted into the generated `where`, or they're referenced as variables in the `where` and passed into the query.

### LINKED ISSUE #839: [MyMeta] Pull member names regardless of whether they have MyMeta profiles
I think it would be great if the guild profiles displayed all their discord members under "members" regardless of whether they have MyMeta profiles otherwise it will be pretty empty.
We'll be working on getting all the guilds that onboard to also have their members create MyMeta profiles, but it still kind of sucks to not have their members list displayed under memberships even as non-clickable.
Same with DAOHaus DAOs, it would be displaying only their public keys if they don't have a discord, but still..

Alec said this wouldn't be easy so let's have this on the backburner? But putting it here just so I don't forget.

alalonde:
## Overview

When the players filter was implemented, it inadvertently starting filtering out any players that hadn't set up a profile, or more specifically, any player without any skills set.

I also fixed the display of a player's name in the PlayerHero component to match that of the PlayerTile.

Closes #739

Need to test to see if this also resolves #839 as well

## Implementation

The graphql query was refactored to take the entire where clause as a variable, rather than just the filter variables.

### DIFF
--- a/packages/web/components/Player/Filter/AdjacentTimezonePlayers.tsx
+++ b/packages/web/components/Player/Filter/AdjacentTimezonePlayers.tsx
@@ -1,6 +1,6 @@
 import { Flex, Skeleton, Text, TimezoneOptions, VStack } from '@metafam/ds';
 import { PlayerList } from 'components/Player/PlayerList';
-import { GetPlayersQueryVariables } from 'graphql/autogen/types';
+import { PlayersQueryVariables } from 'graphql/getPlayers';
 import { usePlayerFilter } from 'lib/hooks/players';
 import { useOnScreen } from 'lib/hooks/useOnScreen';
 import React, { useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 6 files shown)


mquellhorst (approved):
LGTM
Ran locally and all works fine.
Doesn't seem to fix #839

Issue #771, opened 2021-08-19
Pull in player's guild memberships data from Discord


alalonde:
I am finishing up #184 which allows us to pull just about any data from a guild's discord server.  This allows us to improve at least two sections of MyMeta, the first of which this issue will cover:

### A player's list of guilds in their profile
(For reference, see the Memberships section here: https://my.metagame.wtf/player/metadreamer)

Currently we show some hardcoded information for Metagame: 
![image](https://user-images.githubusercontent.com/1402582/130112748-4f45b853-12e0-45bc-ab0b-26211b04ec01.png) 
But now we can migrate existing MyMeta guilds to get their information from a discord server as opposed to moloch. Here is what we currently show for Metadreamer's moloch-fetched DAO membership:  
![image](https://user-images.githubusercontent.com/1402582/130113414-058b6e64-d514-4059-9535-7effb0dc69cb.png)
These images have been manually curated and placed in the MyMeta codebase, whereas now we can just fetch the logo from a Discord CDN.  

### Outstanding questions
- Does it make sense to reorganize the memberships section now? I could see us now separating DAO memberships (which may be totally external to MetaGame) and guild memberships (which by definition are within MyMeta) 
- Currently we have "Player" hardcoded beneath a DAO as well as their pulled-in share (member share divided by total shares). "Player" doesn't really make sense here, I would recommend just removing this. 
- For guilds with their Discord server linked, we could e.g. fetch a player's highest role in their server and display that. 
- Undoubtedly there is more interesting information we could pull in from Discord and display as well. We have full access to their server via the Discord API.

peth-yursick:
Great to see this coming together! Thank you 🙏
> Does it make sense to reorganize the memberships section now? I could see us now separating DAO memberships (which may be totally external to MetaGame) and guild memberships (which by definition are within MyMeta)
What do you mean by reorganizing it? Might not be necessary to separate DAO memberships from guild memberships as guilds can have a DAO or not have it, I'd say prioritize data from daohaus if it exists, otherwise discord.
> Currently we have "Player" hardcoded beneath a DAO as well as their pulled-in share (member share divided by total shares). "Player" doesn't really make sense here, I would recommend just removing this.
> For guilds with their Discord server linked, we could e.g. fetch a player's highest role in their server and display that.
Yeah, I think that makes sense. And for the ones that only have daohaus, we could just have it say "member"?
[...]

PR #949, opened 2021-12-05
Chore/cleanup guild page

### LINKED ISSUE #893: Clean up guild page

Player page:
image

Current guild page:
image

- Adjust guild page column sizes to match the player page
- Get rid of sections that don't have any data: Skills, stats, gallery, quests. Implementing these would be in their own issues, let's just hide for now

### Why is this needed?

UX standardization

vidvidvid:
## Overview

**Disclaimer**

I used this issue as an onboarding issue for @BojanGeorgievski  or Rrash#0514

Guild page has been restyled to match the user profile.
The sections from follow up ideas are commented out.

Closes #893 

Boxes that were hidden and should be implemented next:
- [ ] skills
- [ ] stats
- [ ] quests
- [ ] gallery

## To-do

- [x] Change the default cover image for guilds (so it's not the same as for the user profile)
- [ ] Fix the minimum width for guilds (if the description is short, the boxes become narrow as well)

## Implementation

Copied the structure from user profile and slightly adjusted it.

## Assets

Screenshot 2021-12-30 at 18 32 22


firattale:
LGTM, but maybe we need another PR to fix those Github actions.

luxumbra (commented):
Changes on the preview site look good and the refactor looks sane to my novice eyes. :)

I did notice that no-one shows up in any of the guilds apart from MetaFam, when most of them have players in MetaGame. Is the case for most guilds on prod too.

vidvidvid:
true, i noticed that too, but i didn't look deeper into it. any idea why? were the non-MetaFam players ever displayed as members of a guild?

luxumbra:
> any idea why are they not appearing? 

No idea, sorry. Even the players (penguin, peth, MD) whose guild memberships show correctly on their profiles don't show in the guild profiles. Opolis & MetaFam are the only ones that show members on prod. 

> were the non-MetaFam players ever displayed as members of a guild?
I don't think so

dan13ram:
@vidvidvid The memberships shown in the player profile is fetched by the daoHaus subgraph, whereas the guilds are in our database alone. Therefore there is no link between the "dao" memberships and "guild" memberships. 

The guild memberships is supposed to be fetched from our discord bot, but I'm not sure how far along we are in that development.

vidvidvid:
@dan13ram alrighty then. would be nice if we somehow fixed that once we get higher outer guild involvement.

luxumbra (approved):
LGTM on the FE and code looks fine to my eyes.

dan13ram:
@vidvidvid @luxumbra guild memberships from discord is being tracked at https://github.com/MetaFam/TheGame/issues/771

packages/web/components/Setup/SetupTimeZone.tsx
```diff
@@ -46,6 +48,14 @@ export const SetupTimeZone: React.FC<SetupTimezoneProps> = ({
     onNextPress();
   };
 
+  const [isComponentMounted, setIsComponentMounted] = useState(false);
+
+  useEffect(() => setIsComponentMounted(true), []);
+
+  if (!isComponentMounted) {
```
dan13ram, line 55: Why do we need to do this ?

packages/web/components/Guild/Section/GuildPlayers.tsx
```diff
@@ -50,7 +50,7 @@ export const GuildPlayers: React.FC<Props> = ({ guildId, guildname }) => {
     <ProfileSection title="Players">
       {loadingPlayers && <LoadingState />}
 
-      {guildPlayers == null ? (
+      {guildPlayers == null || guildPlayers.length === 0 ? (
```
dan13ram, line 53: This shows up when `loadingPlayers` is true.  We want to show only loading state initially and then show either "no known players yet" or the list of players.

dan13ram (changes requested):
@vidvidvid Mostly looks good. Requested for a minor fix, then we can merge.

I'd like to merge this first and rebase my code in #1012 because both PRs make changes to `BOX_TYPES`

packages/web/components/Setup/SetupTimeZone.tsx
vidvidvid: A server side rendered page was trying to access `<SelectTimeZone>` component, which is a part of the design-system, and was not found until the client loaded. The builds and deploys failed and when I introduced this change they passed. I don't know how else to solve this problem, do you perhaps have any suggestion?

packages/web/components/Guild/Section/GuildPlayers.tsx
vidvidvid: ahh yeah, gotcha. will fix that.

alalonde:
> @vidvidvid The memberships shown in the player profile is fetched by the daoHaus subgraph, whereas the guilds are in our database alone. Therefore there is no link between the "dao" memberships and "guild" memberships.

See the first bullet-point [here](https://github.com/MetaFam/TheGame/issues/771#issuecomment-1002666049). IMO we should also populate DAO memberships. There is currently a `membership_through_discord` flag on the guild table; if that is false and a DAOhaus address exists, we could populate our guild_player table that way.

> 
> The guild memberships is supposed to be fetched from our discord bot, but I'm not sure how far along we are in that development.

Yes, existing guilds first have to link their discord servers (through the /join/guild) flow for those memberships to be populated.

dan13ram:
> See the first bullet-point [here](https://github.com/MetaFam/TheGame/issues/771#issuecomment-1002666049). IMO we should also populate DAO memberships. There is currently a `membership_through_discord` flag on the guild table; if that is false and a DAOhaus address exists, we could populate our guild_player table that way.
I like this idea. You think we can write a cron to update this every 24hrs or so ?

> Yes, existing guilds first have to link their discord servers (through the /join/guild) flow for those memberships to be populated.
 Sounds good to me!

dan13ram (approved):

packages/web/components/Setup/SetupTimeZone.tsx
dan13ram: Interesting. I tried a few different things and nothing fixed that issue. :/

vidvidvid: Yep, it's quite peculiar 🤷

PR #976, opened 2021-12-15
Update issue templates

mquellhorst:


polimyl:
something feels wrong here.  there are already two issue templates (bug-report and feature-request) that use the new and improved format, while this pr tries to add two new issue templates (bug_report and feature_request) that use the old format and have pretty much the same names as the existing ones without replacing them.

PR #979, opened 2021-12-20
Added !mg prefix to discord-bot commands, added help section

alalonde:
## Overview

Our bot used the !ac prefix which is tied to sourcecred. We should be using our own prefix to uniquely associate commands to our `MetaGameBot`. 

Also, I ported the `help` implementation over to this repo so that all commands are handled by the same codebase.  

After this gets promoted to prod, we should disable the old bot completely.


mquellhorst (approved):
LGTM

PR #941, opened 2021-11-29
[WIP] #911 - Enhance the styling of the MegaMenu

perbruno:
Still need to sort the hover color on avatar

## Overview

This PR tries to make MetaMenu panes closer to what is proposed on figma, based on #911

PR #987, opened 2021-12-21
Feature megamenu styling

### LINKED ISSUE #911: Enhance the styling of the MegaMenu

New styles for the panes that open up when a top-level navigation item is clicked.

Here is a demo in Figma that shows it in action: https://www.figma.com/proto/A6Nh2tyBbRCpglAnm7JQDm/MetaMenu?page-id=86%3A107&node-id=684%3A281&viewport=322%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=684%3A281

Please contact me about any details re: implementing this design. There are things like background blur or drop-shadow effects, that might go unnoticed at first glance. I will gladly be your helper while you build this in 👍

### Why is this needed?

These panes are currently still very plain-looking, and don't fit all too well with the rest of the website. The new styling will make it feel more like an integral part of the website, which it really is.

vidvidvid:
## Overview

Updated the styling of the megamenu with colors and animations.

Closes #911 
Closes #941 

- [ ] Based on the [figma demo](https://www.figma.com/proto/A6Nh2tyBbRCpglAnm7JQDm/MetaMenu?page-id=86%3A107&node-id=684%3A281&viewport=322%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=684%3A281) the icon's background gets animated when leaving the menu item is not hovered anymore. Since the item's background then changes from `linear-gradient` to normal color, the css can't handle that by default and we'd have to make use of some hacky techniques, which I'm not very skilled in, so I think we can just wait until this kind of animation becomes supported by css.
_we might have to wait for years_

## Assets

https://user-images.githubusercontent.com/35112939/146969413-f58936c6-62bd-4700-95ee-417c975eba11.mov

paired-with @perbruno

### DIFF
--- a/packages/web/components/MegaMenu.tsx
+++ b/packages/web/components/MegaMenu.tsx
@@ -139,6 +139,7 @@ const DesktopMenuItem = ({
       display="flex"
       className="desktop-menu-item"
       href={url}
+      width="full"
       m={1}
       padding="1rem"
       borderRadius="0.618vmax"
--- a/packages/web/components/MegaMenu.tsx
+++ b/packages/web/components/MegaMenu.tsx
@@ -139,13 +139,16 @@ const DesktopMenuItem = ({
       display="flex"
       className="desktop-menu-item"
       href={url}
+      m={1}
       padding="1rem"
+      borderRadius="0.618vmax"
       _hover={{
-        bg: 'rgba(0,0,0,0.56)',
-        borderRadius: '0.618vmax',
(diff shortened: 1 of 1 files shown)


luxumbra (approved):
LGTM. Thanks @vidvidvid 🙏🏻  Although, having said that, I notice in the Figma also, that there's a bit of a box shadow on hovered menu items. 😄 
image

PR #1004, opened 2022-01-03
Handle Redirection on client side

### LINKED ISSUE #988: Dashboard should only be visible when authenticated

When a user accesses `/dashboard` and is not logged in, they should be forwarded to `/players`.

### Why is this needed?

The dashboard is full of personalized content, so if there is no one to personalize to the page doesn't make sense.

firattale:
## Overview
The dashboard page should be restricted for unauthorized users.

Closes #988 

Redirection has been made on the client side, it would be better if we can do in on server side, but I can't find a way to check if user is authenticated on the server side, any ideas?

## Implementation

**Side effects**

## Assets

### DIFF
--- a/packages/web/pages/dashboard.tsx
+++ b/packages/web/pages/dashboard.tsx
@@ -16,6 +16,7 @@ import { LatestContent } from 'components/Dashboard/LatestContent';
 import { Leaderboard } from 'components/Dashboard/Leaderboard';
 import { Seed } from 'components/Dashboard/Seed';
 import { XP } from 'components/Dashboard/XP';
+import { useUser } from 'lib/hooks';
 import { ReactElement, useEffect, useState } from 'react';
 import { Layout, Layouts, Responsive, WidthProvider } from 'react-grid-layout';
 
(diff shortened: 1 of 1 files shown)


luxumbra (commented):
Works for me on the preview. 👍🏻  Agreed with the server side over client side but I am not sure how / if it can be done. @mquellhorst or @alalonde, any thoughts sers?

dan13ram:
I think we should show a loading state and then redirect. Or perhaps just show a huge connect button to login.  
Currently the whole dashboard is displayed for a split second before redirection.

luxumbra:
> I think we should show a loading state and then redirect. Or perhaps just show a huge connect button to login.
> Currently the whole dashboard is displayed for a split second before redirection.

Good point. A large login / connect button and maybe some text "Connect to view your dashboard or create your profile" would be an improvement over a brief flash of the dash.

firattale:
There was a debate here  https://github.com/MetaFam/TheGame/issues/988 about this issue, it is better to close this PR and start with the new plan @luxumbra @dan13ram

Issue #1030, opened 2022-01-08
Redirect to 404 when player doesn't exist


dan13ram:
Working upon #1013
We should updating wherever we display 404 errors with a redirect to this page.
For example, when a player is not found in `/player/<username>`
https://test.metagame.wtf/player/thisplayerdoesnotexist

firattale:
I can take this since it is related to 404 page.

firattale:
@dan13ram this PR https://github.com/MetaFam/TheGame/pull/1033 closes this issue, I also realized we had same problem for guild page and quest completed page too.

PR #1033, opened 2022-01-10
feat/ add 404 redirection for nested routes

firattale:
## Overview
Custom 404 Page does not shown for nested routes. That has been fixed with correct redirection from the 404 page to the homepage or dashboard.

A path like "/player/404", "/guild/404" and "/quest/[id]/completed" were using regular 404 pages and I changed them to show our custom 404 page.

https://github.com/MetaFam/TheGame/issues/1030

Closes #

https://github.com/MetaFam/TheGame/issues/1030

## Implementation

I have to add fallback:false and hideTopMenu:true props to the getStaticProps, otherwise the page tries to build the page with header and then redirects to 404 , but now it is just showing the custom 404 page without any header.

Also the redirection to homepage from 404 has been fixed according to the @vidvidvid suggestion.

@vidvidvid 's comment:
image

### DIFF
--- a/packages/web/pages/guild/[guildname].tsx
+++ b/packages/web/pages/guild/[guildname].tsx
@@ -164,7 +164,7 @@ export const getStaticPaths: GetStaticPaths<QueryParams> = async () => {
     paths: guildnames.map((guildname) => ({
       params: { guildname },
     })),
-    fallback: false,
+    fallback: 'blocking',
   };
 };
 
--- a/packages/web/pages/guild/[guildname].tsx
+++ b/packages/web/pages/guild/[guildname].tsx
@@ -13,8 +13,8 @@ import {
   GetStaticPropsContext,
   InferGetStaticPropsType,
 } from 'next';
-import Error from 'next/error';
 import { useRouter } from 'next/router';
+import Page404 from 'pages/404';
 import React from 'react';
 import { BoxType } from 'utils/boxTypes';
 import { getGuildCoverImageFull } from 'utils/playerHelpers';
@@ -40,7 +40,7 @@ const GuildPage: React.FC<Props> = ({ guild }) => {
   }
 
   if (!guild) {
-    return <Error statusCode={404} />;
+    return <Page404 />;
   }
 
   const getBox = (name: string): React.ReactNode => {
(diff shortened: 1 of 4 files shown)


packages/web/pages/guild/[guildname].tsx
```diff
@@ -165,7 +165,7 @@ export const getStaticPaths: GetStaticPaths<QueryParams> = async () => {
     paths: guildnames.map((guildname) => ({
       params: { guildname },
     })),
-    fallback: true,
```
Comment (dan13ram) [inline packages/web/pages/guild/[guildname].tsx:168]: @firattale fallback should be true here

packages/web/pages/player/[username].tsx
```diff
@@ -300,7 +300,7 @@ export const getStaticPaths: GetStaticPaths<QueryParams> = async () => {
     paths: playerUsernames.map((username) => ({
       params: { username },
     })),
-    fallback: true,
+    fallback: false,
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:303]: here as well, fallback should be true

Comment (dan13ram) [reply packages/web/pages/player/[username].tsx:303]: We should deal with these cases as described in https://nextjs.org/docs/basic-features/data-fetching#fallback-true

Comment (firattale) [reply packages/web/pages/player/[username].tsx:303]: If we add fallback:true, the page tries to render with header, but since there is no data, it immediately turns back to 404 page which doesnt have any header. I think that is not a pretty UX. If fallback is not true like here, the page doesnt force itself for showing the content, if it is not available, it renders the 404 page without a hassle.

Comment (dan13ram) [reply packages/web/pages/player/[username].tsx:303]: @firattale fallback: true is necessary in case new players login for the first time and their profile page is not rendered yet on the server.

Comment (firattale) [reply packages/web/pages/player/[username].tsx:303]: I think fallback:blocking may help here as it is working similar to true but doesnt fallback to 404.

`fallback: 'blocking'

This is quite similar to fallback: true, except that it does not return the dummy loading page when a page that hasn't been cached is hit for the first time

Instead, it just makes the browser hang, until the page is rendered for the first time.

Future requests to that page are quickly served from the cache however, just like fallback: true.

 `

Comment (firattale) [reply packages/web/pages/player/[username].tsx:303]: wdty? @dan13ram

Comment (dan13ram) [reply packages/web/pages/player/[username].tsx]: Sounds great to me. But I just hope the blocking is not too long. Otherwise, it makes sense to display a loader to the user.

dysbulic (approved):
I was unaware of the `blocking` as a `fallback` option. I've been seeing a 404 when navigating to profile pages & it seems like this'd fix it.

dan13ram (approved):
Lets rebase this and merge :)

PR #1039, opened 2022-01-12
fix: bugs in sort/filter ui

### LINKED ISSUE #786: Selecting an already-selected sort-by option in player filter causes 500
On the players page, sort by seasonal XP is preselected. If you click it again, you end up with a white screen and "an unexpected error occurred"

dan13ram:
Closes #786

### DIFF
--- a/packages/web/components/Player/Filter/DesktopFilters.tsx
+++ b/packages/web/components/Player/Filter/DesktopFilters.tsx
@@ -89,7 +89,9 @@ export const DesktopFilters: React.FC<Props> = ({
           value={[sortOption]}
           onChange={(value) => {
             const values = value as ValueType[];
-            setSortOption(values[values.length - 1]);
+            if (values[values.length - 1]) {
+              setSortOption(values[values.length - 1]);
+            }
           }}
           options={sortOptions}
         />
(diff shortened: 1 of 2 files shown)


vidvidvid:
I noticed this issue is also present in quest filters, nice that you have solved it. Do you think there would be a way to just not trigger the `onChange` contents if `<MetaFilterSelectSearch />` has the `disableEmpty` flag?

dan13ram:
@vidvidvid that's what I tried at first, but there was some issue and onChange was always called. So I made this simpler fix.

vidvidvid:
hey i think i got it - pass `disableEmpty` as a flag and then grab it in the `<SelectOption> `:
```
  const {
    isSelected,
    data: { value: optionValue },
    selectProps: { onChange, value: selectValue, disableEmpty },
  } = props;
```

and then fix the onClick in the returned element:

```
return (
    <Flex
      ...
      onClick={isSelected && !disableEmpty ? clearValue : undefined}
      ...
    >
```

vidvidvid:
Oh nevermind. It doesn't solve it. Please ignore my last comment. 😅

dan13ram:
> Oh nevermind. It doesn't solve it. Please ignore my last comment. sweat_smile

@vidvidvid Yes, like I said I tried this first but it didn't work. So I resorted to the current code. 
Can you approve it if you think its good to merge?

vidvidvid (approved):
Let's merge!

PR #1042, opened 2022-01-13
Fleshed out guild join page

### LINKED ISSUE #896: Standardize guild onboarding copy

We currently have [this Notion page](https://meta-game.notion.site/Guild-Onboarding-Why-How-e3d7383d86de4f7f8240a20cd92b8efb) discussing Guild Onboarding. A lot of this content should probably be in our _actual_ guild onboarding page in [MyMeta](https://my.metagame.wtf/join/guild). This will require some design work and copy work.

### Why is this needed?

Because potential new guilds should have more context about joining: why, if they're a good fit, etc.

alalonde:
## Overview

Fleshes out the guild join page with additional copy that can help the applicant determine fit.
Closes #896 

- Seems that guilds should be able to join MetaGame without linking Discord along the way. We would need an alternative flow which basically goes right to the form with possibly the DAO address required so we can pull memberships from DaoHaus
- This page still hasn't been properly designed :D

## Assets
image

### DIFF
--- a/packages/web/components/Guild/GuildJoin.tsx
+++ b/packages/web/components/Guild/GuildJoin.tsx
@@ -168,12 +168,7 @@ export const JoinCopy: React.FC = () => {
           <MetaLink href="/">MetaGame</MetaLink>
         </ListItem>
         <ListItem>
-          <MetaLink
-            isExternal
-            href="https://wiki.metagame.wtf/docs/wtf-is-metagame/wtf-is-metagame"
-          >
-            WTF Is MetaGame
-          </MetaLink>
+          <MetaLink href="/learn/wiki">WTF Is MetaGame</MetaLink>
         </ListItem>
         <ListItem>
           <MetaLink
--- a/packages/web/components/Guild/GuildJoin.tsx
+++ b/packages/web/components/Guild/GuildJoin.tsx
@@ -1,14 +1,11 @@
 import {
   Box,
-  Flex,
-  HStack,
   Image,
   ListItem,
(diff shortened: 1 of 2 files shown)


alalonde:
Mobile version is no bueno currently, should be some simple tweaks to make it better.

packages/web/components/Guild/GuildJoin.tsx
```diff
@@ -105,3 +87,238 @@ export const GuildJoin: React.FC = () => {
...
+      </UnorderedList>
+      <Text fontWeight="bold">Read more about MetaGame:</Text>
+      <UnorderedList>
+        <ListItem>
+          <MetaLink href="/">MetaGame</MetaLink>
+        </ListItem>
+        <ListItem>
+          <MetaLink
+            isExternal
+            href="https://wiki.metagame.wtf/docs/wtf-is-metagame/wtf-is-metagame"
```
vidvidvid, line 168: maybe you could link them to `/learn/wiki`

alalonde: ah, good idea

mquellhorst:
Two things:

- I notice that when I'm not logged in to Notion I'm unable to access the manifesto. Maybe we can direct somewhere else or move that page from Notion?
- Any thoughts on how we link to this page from the Guilds overview page or from the menu (Contribute as Guild)?

mquellhorst (approved):
The access of the Notion page should be changed so anyone can see it without the need to login

PR #1067, opened 2022-01-20
Added cron trigger to sync guild memberships from Discord

### LINKED ISSUE #897: Guild memberships aren't kept up to date

We need a scheduled Cron Trigger in Hasura to run the `syncDiscordGuildMembers` action for all guilds that have `membership_through_discord` set. 

That action may need additional error handling as well.

### Why is this needed?

So changes in discord membership are reflected in MyMeta

alalonde:
## Overview

When guilds go through the Discord server linking process, we sync their members with our own database, populating the guild_player table.  Until now, those members were out of date as there was no job to keep them in sync.

This resolves #897 by way of a new Cron trigger in Hasura.

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -128,14 +128,6 @@
   - name: GuildType
     using:
       foreign_key_constraint_on: type
-  - name: metadata
-    using:
-      manual_configuration:
-        remote_table:
-          schema: public
-          name: guild_metadata
-        column_mapping:
-          id: guild_id
   array_relationships:
   - name: guild_players
     using:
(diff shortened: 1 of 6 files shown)


hasura/metadata/tables.yaml
```diff
@@ -435,7 +425,6 @@
       - availability_hours
       - color_mask
       - player_type_id
-      - profile_layout
```
dan13ram, line 438: @alalonde why are we removing this ?

alalonde: Hmm, that snuck in there again. There should be no changes in this file. My local hasura metadata is messed up.

alalonde:
OK, should be gtg now.

dan13ram (approved):
Looks good to me!

PR #1070, opened 2022-01-21
Switch to discordx

### LINKED ISSUE #1038: Convert the Discord bot to ESM using discordx

The library currently used as the basis for the Discord bot, [`@typeit/discord`](https://github.com/OwenCalvin/discord.ts), has been abandoned and is no longer maintained.

The recommended alternative is [`discordx`](https://discord-ts.js.org).

### Why is this needed?

`node-fetch` can no longer be imported using a `require` function as of v3. It is solely supporting ESM (ECMAScript Modules).

There are [instructions](//github.com/node-fetch/node-fetch/issues/1279#issuecomment-915062146) describing how to convert a project to ESM in the `node-fetch` repository.

This would also be an opportunity to do the migration to slash commands (#514).

alalonde:
## Overview

- Upgrades Discord.js to v13
- Overhauls the discord bot to use `discordx` as the old `discord-ts` package is no longer maintained

Closes #1038 

- [ ] Convert current "simple commands" to slash commands as they have much better discoverability and UI
- [x] Split up current bot into a MyMeta bot and MetaGame bot. The MyMeta bot should only need readonly access to their server data, while the MetaGame bot currently mutates user roles in the MetaGame discord.

## Implementation

No new functionality here, just the upgrades. Should be invisible to consumers


alalonde:
This is close. Running the discord-bot by itself has no issues, but when imported by the backend package it falls over. We could probably have a separate entrance point when included in this manner, since dependent packages should not have any interest in discordx, but just the discord.js client.

alalonde:
This PR includes an upgrade to node.js 16.6 (required by Discord.js v13 and discordx).

Unfortunately, Vercel [doesn't currently support > v14](https://github.com/vercel/community/discussions/37) so this is effectively blocked.

alalonde:
We are no longer dependent on Vercel, so this should be good to go.

Issue #1062, opened 2022-01-19
Split the Google Cloud Run GitHub Action into separate jobs


dysbulic:
### What would you like to be added?

Currently the cloud deployment GitHub action is a single [job](//docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobs) broken into about a dozen steps. The steps in a job execute serially whereas multiple jobs will run in parallel.

The [`needs`](//docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) keyword should be used to make sure that containers are built before they are deployed & that the backend starts before Hasura and Hasura before the frontend.

### Why is this needed?

It takes about half an hour for an instance to deploy if all the containers need to be rebuilt.

The existing structure could likely be half an order of magnitude faster if properly parallelized.

Rexben001:
I can work on this @dysbulic

dysbulic:
@Rexben001, the GitHub deployment action has undergone significant revision in the [`feat/glaze-lookup` branch](//github.com/MetaFam/TheGame/blob/feat/glaze-lookup/.github/workflows/gcp-deploy.yaml), and that is hopefully going to get merged in the next couple days.

In any case, you should work from that Action rather than the one that's in `develop`.

So far as ordering, the important thing to remember is the backend has to start before Hasura will start, and Hasura has to be running for the frontend to *build* (not start, build).

All of the teardown and image removal tasks could be in parallel, as could the building of the backend and Hasura.

dysbulic:
@Rexben001, I went ahead and did this in #1078 because I needed to move the seeding step & the decomposition into jobs was relatively straightforward.

PR #943, opened 2021-12-01
Reading and Writing Profile Info From Ceramic

dysbulic:
## Overview

**What features/fixes does this PR include?**

This pull request does these primary things:
1. It changes the process for loading data into Hasura from Ceramic from IDX to Glaze (#917)
2. It migrates the proxy for [web3.storage](//web3.storage) from the Express backend to a Next.js API endpoint (#848)
3. It enables saving of profile data into a split between Ceramic's `basicProfile` and an extended profile document (#920)
4. It changes the profile return values from Hasura from underscored to camel case so they can be destructured with regular React variables
5. It updates the version of Node used for the front and backends to v16
6. It fixes the Google Cloud Run deployments to update when changes are made to a pull request
7. It refactors the rate limiting of requests to the Ceramic daemon (imperfectly #1061)

## Follow up Improvement Ideas

* React is upgraded to 17.0.2 for compatibility with `react-form-hook`.
* Time zone is uniformly capitalized as timeZone since it is two words.
* The typing on the profile editing page is nearly elegant. I am using a union type to check which fields go to which stream and to handle the mapping between Hasura ids and Ceramic ones.

## Implementation

**Describe technical (nontrivial / non-obvious) parts of your code**

The main thing this PR does is it moves all the profile fields into their own table and defines the appropriate types for accessing them.

**Side effects**

The biggest side effects are from the fact that I standardized all the variable names as camel case when migrating them. The most pernicious was `timeZone`, which despite being two words was a mishmash of capitalizations.

PR #1078, opened 2022-01-25
[WIP] Within-app profile update post editing

### LINKED ISSUE #887: Optimistic updates for profile editing

It would be nice if instead of waiting for a couple of seconds, the users would get the impression that their data was updated immediately after pressing the save button. In a vast majority of the cases there would be no errors, but if they occur, we would later on have to patch the data and display an error message. Eventually we also might want to think about implementing error tracking software.

We would want this implemented in all of the updatable fields - which includes:
- player hero fields
  - username
  - availability
  - time zone
- skills
- color disposition
- player type

The speed of updating now:
update

### Why is this needed?

It's a nice to have feature that improves the UX.

dysbulic:
## Overview

This PR adds the [Jotai state management library](//jotai.org) to propagate changes made by editing a player's profile to the display version.

The code in #943 writes to the Ceramic Network from the user's browser and notifies Hasura to renew its cache for that user. There is a time delay between when the user saves their changes and when an updated version of the profile page is available. This is exacerbated by caching by Next.js. This PR directly updates the profile page within the app so the user sees their inputted information.

There are also a couple of consistency changes, most notable switching `PlayerFragmentFragment` to `Player` throughout. Also the removal of the `gql` `fake-tag` in favor of the `/* GraphQL */` identifier.

`useUser` now simply returns a `Player` object rather than a `user` object with a `player` attribute. The name of the returned object is still `user`. Also, rather than using the same query for the player avatar in the top menu with a dozen `@skip if $forLoginDisplay`s, there is now a separate `useLoginBasics` query that only returns the details needed for that avatar.

Closes #887 though "optimistic updates" is something of a misnomer because the editing flow will successfully complete the write to Ceramic and Hasura cache invalidation before it will update the values in Jotai.

Also, divides the GitHub deployment Action into separate jobs to address #1062 & moves the seeding step before the frontend generation step for #1083.

There are a couple places we are prop drilling which could potentially be cleaned up using Jotai.

On a side note, there's also a Hasura migration to make skill / player combinations unique. Somehow the test deployment had double entries for all the players' skills for a bit & this should be prevented at the database level. It's unrelated, but the Hasura schema in this branch is ahead of `develop`, so I just put it here.

## Implementation

The PR adds a `useProfileField` hook that provides access to the profile data that can be overwritten with local state.

It also adds a `useSaveCeramicProfile` that takes hashes of values and images, and sets the appropriate profile stream properties. *(Values are stored in two separate streams, a basic one defined by 3Box and an extended one of my creation.)* That hook also handles setting the local state values for the UI.


dysbulic:
This PR has been broken into smaller requests for easier reviewing: #1103, #1115, #1116, #1124, & #1127.

PR #1082, opened 2022-01-26
fix: better meta tags for pages

### LINKED ISSUE #1081: Set appropriate (and normal) page titles

Using the [`HeadComponment`](//github.com/MetaFam/TheGame/blob/f5295c324244f2973ea419c588e784c142a8091b/packages/web/components/Seo.tsx#L11) it is possible to customize the metadata associated with a page.

While testing this out, I set one of the titles to “Ⲙⲉⲧⲁ𝓖ⲁⲙⲉ” which I thought looked cool, but it went over decidedly poorly with a designer.

So, the various pages should be given descriptive titles using the standard Latin unicode glyphs. Also, the image for a player’s profile page should be their profile image.

### Why is this needed?

There was some kerfuffle over the fanciful glyphs, and, as someone pointed out, they are likely bad for SEO.

Adding desciptive titles in general will just make the pages a little more easily navigable.

dan13ram:
Fixes meta tags for various pages

Closes #1081

### DIFF
--- a/packages/web/pages/_app.tsx
+++ b/packages/web/pages/_app.tsx
@@ -17,7 +17,7 @@ const App: React.FC<WithUrqlProps> = ({
     <CSSReset />
     <Head>
       <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-      <title>Ⲙⲉⲧⲁ𝓖ⲁⲙⲉ</title>
+      <title>MetaGame</title>
       {CONFIG.gaId != null && (
         <>
           <script
--- a/packages/web/components/Seo.tsx
+++ b/packages/web/components/Seo.tsx
@@ -9,32 +9,44 @@ export type HeadMetaProps = {
 };
 
 export const HeadComponent: React.FC<HeadMetaProps> = ({
-  title = '𝗠eta𝗚ame',
+  title = 'MetaGame',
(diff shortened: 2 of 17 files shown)


packages/web/pages/_app.tsx
```diff
@@ -17,7 +17,7 @@ const App: React.FC<WithUrqlProps> = ({
     <CSSReset />
     <Head>
       <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-      <title>Ⲙⲉⲧⲁ𝓖ⲁⲙⲉ</title>
+      <title>Metagame</title>
```
dysbulic, line 20:
```suggestion
      <title>MetaGame</title>
```

dysbulic (changes requested):

packages/web/pages/_app.tsx
dan13ram: @dysbulic fixed this :)

dysbulic:
On the subject of excessive XP for PRs, there's not alot to this one.

You could consider doing what is described in #1081 and set the titles on the various pages appropriately. Something like “MetaGame: Players” or “MetaGame: Guilds” or whatever.

Also, I've already set the profile details page to include the profile name, but I think the image associated with that page should be their profile image.

dysbulic (approved):

PR #1100, opened 2022-02-01
Ensure that only known users can go through the guild setup process

### LINKED ISSUE #1063: Require login before going through the Guild Join process

In the Join Guild page, require logging in before going through the process.

Also give more information about the DAO address as some guilds have multiple candidates.

### Why is this needed?

If you don't log in, it appears like the process still works, but fails when fetching your guild's roles.

alalonde:
## Overview

- Only known users (those with a connected profile) can go through the guild setup process. This enforces that. 
- Form visual improvements
- Added more detail to the "DAO address" field

Closes #1063 

## Implementation

**Side effects**

I made a CSS change to the shared `Field` component to make it less opinionated.

## Assets

image

### DIFF
--- a/packages/web/components/Guild/GuildForm.tsx
+++ b/packages/web/components/Guild/GuildForm.tsx
@@ -10,7 +10,7 @@ import {
   VStack,
 } from '@metafam/ds';
 import { SelectOption } from '@metafam/ds/src/MultiSelect';
-import { Field } from 'components/Forms/Field';
+import { Field, FieldDescription } from 'components/Forms/Field';
 import { MetaLink } from 'components/Link';
 import {
   DiscordRole,
@@ -160,11 +160,7 @@ export const GuildForm: React.FC<Props> = ({
   }, [workingGuild, guildMetadata, roleOptions, reset]);
 
   return (
-    <Box
-      w="100%"
-      maxW="40rem"
-      sx={{ '.description': { ml: 4, mt: 1, fontSize: 'sm' } }}
-    >
+    <Box w="100%" maxW="40rem">
       <VStack>
         <Field label="Guildname" error={errors.guildname}>
           <Input
@@ -185,9 +181,9 @@ export const GuildForm: React.FC<Props> = ({
             isInvalid={!!errors.guildname}
             background="dark"
           />
-          <span className="description">
+          <FieldDescription>
             A unique identifier for your guild, like a username.
-          </span>
+          </FieldDescription>
         </Field>
         <Field label="Name" error={errors.name}>
           <Input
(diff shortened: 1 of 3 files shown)


vidvidvid:
tested the workflow, looks good to me!
the only improvement suggestion would be to perhaps to add the button to connect inside the
> Please log in or create a player profile by pressing the "Connect"

but it's not required at all, since the continuation of this message is clear enough.

packages/web/components/Guild/GuildForm.tsx
```diff
@@ -180,7 +185,9 @@ export const GuildForm: React.FC<Props> = ({
             isInvalid={!!errors.guildname}
             background="dark"
           />
-          <span>A unique identifier for your guild, like a username.</span>
+          <span className="description">
```
dysbulic, line 188: I would use `chakra.span` or `<Text as="span">`, remove the class, and move the `Box`'s `sx` attribute down here.

Though, honestly, given how the elements are used on the page, you could use a `block` element, so just `<Text/>`.

packages/web/components/Guild/GuildForm.tsx
```diff
@@ -197,7 +204,7 @@ export const GuildForm: React.FC<Props> = ({
             isInvalid={!!errors.name}
             background="dark"
           />
-          <span>
+          <span className="description">
```
dysbulic, line 207: If you want to repeat the style, create a local component like:

```typescript
const Description: React.FC = ({ children }) => (
  <Text as="span" ml={4} mt={1} fontSize="sm">
    {children}
  </Text>
);
```
```suggestion
          <Description>
```

packages/web/components/Guild/GuildForm.tsx
```diff
@@ -239,23 +248,35 @@ export const GuildForm: React.FC<Props> = ({
...
           />
-          <span>If your guild has a DAO, enter its address here.</span>
+          <span className="description">
+            If your guild has a DAO, enter its address here. This is the address
+            that will be used to look up your DAO's information from the{' '}
+            <MetaLink
+              isExternal
+              href="https://thegraph.com/hosted-service/subgraph/odyssy-automaton/daohaus"
+            >
+              DaoHaus Supergraph
```
dysbulic, line 276:
```suggestion
              DaoHaus Subgraph
```

dysbulic:
> perhaps to add the button to connect inside the

You could link the “create a profile” text to [/me](//test.metagame.wtf/me). That will present them with a login prompt.

dysbulic (changes requested):

packages/web/components/Guild/GuildForm.tsx
dan13ram: Its called Supergraph

dan13ram (approved):
LGTM

dysbulic (approved):

Issue #1080, opened 2022-01-26
Have the Cloud Run Action create deployments


dysbulic:
### What would you like to be added?

There is [a GitHub Action](//github.com/MetaFam/TheGame/blob/develop/.github/workflows/gcp-deploy.yaml) which builds and deploys the system for any open pull requests.

Currently, it notifies the user that the deployment is available by commenting on the associated PR.

There is [a deployment Action](//github.com/bobheadxi/deployments) that will:
* Display that the deployment is starting and then change state once complete (or failed).
* Appear streamlined in the PR discussion like Vercel's deployments.
* Place entries in the [GitHub deployments list](//github.com/MetaFam/TheGame/deployments).

### Why is this needed?

It is mostly just prettier & not a high priority issue. It would distinguish the deployment messages from the other comments, and make the conversations panel a little easier on the eye.

Simultaneously, it should be a relatively simple matter of adding a couple steps to the deployment Action, so it is low-hanging fruit.

Issue #1083, opened 2022-01-26
For the test deployments, seed the database before generating the frontend


dysbulic:
### What would you like to be added?

Currently, when a Cloud Run test instance is generated, at the point the frontend is built Hasura is empty. Because of this, the players list page initially displays as empty.

If the database seeding step was moved before the frontend generation in [the deployment Action](//github.com/MetaFam/TheGame/blob/develop/.github/workflows/gcp-deploy.yaml) it should fix that.

### Why is this needed?

The default players list reports there are zero players in the system.

dysbulic:
This is a pretty low effort task. An interested party might tackle #1080 or #1062 in the same PR to warrant the XP.

PR #1103, opened 2022-02-02
Break GitHub Deployment Action Into Jobs

### LINKED ISSUE #1062: Split the Google Cloud Run GitHub Action into separate jobs

dysbulic:
## Overview

This PR takes what were a list of steps in the GitHub Action to deploy to Google Cloud Run and converts them into jobs with appropriate dependencies between them.

Because each job can potentially run on a separate instance, some of the steps like checking out the repository and authenticating to gcloud needed to be duplicated.

Also, this PR comments out the build step of the continuous integration test suite. A build is performed by the deployment, so that one was redundant.

Additionally, it moves the seeding step to before the frontend generation because the existing order was causing Next.js's initial cache to claim there are no players in the database.

Closes #1062 & #1083.

Currently when the deployment is complete a comment is created in the PR giving the frontend and Hasura URLs. I would like to switch to using GitHub's deployments (#1080) which are more streamlined in the PR conversation flow & display under environments for the PR.

## Implementation

Environment variables are not shared between Action jobs. Also they can't be accessed from within the same step where they are defined. This PR defines an `env` job which generates all the various variables & exposes them as job outputs.


polimyl (changes requested):
the delete/remove job names are inconsistent and the database ones dont mention also deleting/creating an user.
otherwise lgtm

dysbulic:
@polimyl, I revamped the changes & I think I addressed your issues. The changes include:

* Removal of the commented out build portion of the CI Action
* Removal of references to the database being "local" in the seeding script
* Consistency in the naming of tasks for different subsystems
* Mention of the user creation task when setting up the db
* Division of the db and user creation into separate steps
* Addition of a step to remove the db user
* Migration from a comment in the PR to a GitHub deployment for notification to close #1080
* Division of long task commands into multiple lines

.github/workflows/gcp-deploy.yaml
```diff
@@ -19,14 +19,36 @@ env:
...
+            by ${{github.event.pull_request.user.login}}"
+          ref: ${{github.event.pull_request.head.sha}}
+
   cancel-previous:
     name: Cancel Existing Workflows
     runs-on: ubuntu-latest
 
     steps:
-      - name: Cancel Existing Runs
+      - name: Cancel Running Actions
```
polimyl, line 51: the previous "cancel existing runs" might actually be more accurate since this just cancels currently running runs of this action for this pr

.github/workflows/gcp-deploy.yaml
```diff
@@ -48,158 +70,192 @@ jobs:
...
 
   delete-db:
-    name: Delete Existing Database
+    name: Delete Hasura’s User & Database
     runs-on: ubuntu-latest
     needs: [env]
 
     steps:
-      - name: Set Up gcloud CLI
+      - name: Set Up GCloud CLI
```
polimyl, line 137: offical name is actually just "gcloud", all lowercase (according to its docs), same below

.github/workflows/gcp-deploy.yaml
```diff
@@ -48,158 +70,192 @@ jobs:
...
           sleep 10
-          PGPASSWORD=${{secrets.GCP_POSTGRES_PASSWORD}} dropdb -h /tmp/cloudsql/${{needs.env.outputs.CLOUDSQL_CONNECTION_NAME}} -U postgres ${{needs.env.outputs.DB_NAME}} -f
+          PGPASSWORD=${{secrets.GCP_POSTGRES_PASSWORD}} \
+            dropdb -h /tmp/cloudsql/${{needs.env.outputs.CLOUDSQL_CONNECTION_NAME}} \
+            -U postgres ${{needs.env.outputs.DB_NAME}} -f
           kill $PID
 
   create-db:
-    name: Create New Database
+    name: "Create New SQL User & Database"
```
polimyl, line 165: why does this have quotation marks but the delete step doesnt? also the dleete step mentions hasura but this doesnt (fine with either, but both should be the same imo)

.github/workflows/gcp-deploy.yaml
```diff
@@ -48,158 +70,192 @@ jobs:
...
+      - name: "Delete Service: ${{needs.env.outputs.BACKEND_SERVICE}}"
         continue-on-error: true
-        run: gcloud -q run services delete ${{needs.env.outputs.BACKEND_SERVICE}} --region ${{needs.env.outputs.CLOUDRUN_REGION}}
+        run: |
+          gcloud -q run services delete ${{needs.env.outputs.BACKEND_SERVICE}} \
+            --region ${{needs.env.outputs.CLOUDRUN_REGION}}
 
   delete-backend:
-    name: Remove Backend Image
+    name: Remove Backend Container Image
```
polimyl, line 208: job id uses the word "delete" while the anme uses the word "remove"

.github/workflows/gcp-deploy.yaml
```diff
@@ -48,158 +70,192 @@ jobs:
...
-      - name: Set Up gcloud CLI
+      - name: Set Up GCloud CLI
         uses: google-github-actions/setup-gcloud@v0.2.1
         with:
           project_id: ${{needs.env.outputs.PROJECT_ID}}
           service_account_key: ${{secrets.GCP_SA_KEY}}
           export_default_credentials: true
 
-      - name: Delete Backend Image
+      - name: "Delete Container: ${{needs.env.outputs.BACKEND_TAG}}"
```
polimyl, line 227: i have recently learned the proper definitions of this terminology: a container image is a static file thats more or less just a snapshot of a container while a container is a running instance of a container image
you seem to mix those up in multiple places (i also did in the original workflow)

polimyl (changes requested):
might be a bit too nitpicky, but eh

.github/workflows/gcp-deploy.yaml
dysbulic: The ampersand requires quotes.

polimyl: the delete-db job also got an ampersand but no quotes though. guess you will have to add quotes there then

dysbulic: I would've sworn that VSCode told me I needed quotes, but apparently not…

polimyl (approved):
lgtm

Issue #1035, opened 2022-01-11
Add button to reset profile layout to default


dan13ram:
### What would you like to be added?

Improve upon the profile layout introduced in #1012 and allow players to reset their layout to default with a button click.

### Why is this needed?

Quote from @vidvidvid 
>   if i attempted to reset layout I assumed i'd get the default layout, but the layout changed to the last saved version. Would it make sense to add another button for resetting the layout to its default values?

LouieRichardson99:
I will start work on this.

dan13ram:
> There is also no "cancel" button to discard current changes. I added something and didn't like it, and had to refresh to "discard" the changes

@alalonde There should be a "Reset" button right ? Didn't that reset your change?

vidvidvid:
Maybe cancelling should reset and exit the editing mode?

vidvidvid:
since there has been no reported progress on this, i'll be fixing this in #1104
since there is already a cancel button, do we agree we add a button to revert to the default layout or should we just close the issue?
@dan13ram @alalonde ?

vidvidvid:
@LouieRichardson99 i assume you've abandoned the issue?

alalonde:
> @alalonde There should be a "Reset" button right ? Didn't that reset your change?

Ah, to me, "reset" means to reset your saved configuration to the default. But "cancel" should revert any unsaved changes.

vidvidvid:
Solved in #1104

Issue #1003, opened 2021-12-31
MegaMenu - Styling enhancements


davort:
After a few iterations, the MetaMenu is almost there when it comes to its look & feel 🙌

Just a few more smaller changes, and it will be perfect!

Please use the below image for visual reference:
![megamenu-finetune.png](https://images.zenhubusercontent.com/603d32d6edb5330e5cd0dd54/16a6138f-ae14-42d5-b415-4e224aae90c3)

To do:

1. Make the logo vertically centered (relative to the header).
2. Set the `font-weight` for top-level links to `700`
3. Set the `color` for active/expanded top-level link to `var(--chakra-colors-cyan-300)`
4. Make the opened menus horizontally centered relative to the viewport.

Let me know if there's any questions or concerns and we'll figure it out together!

LarrySimiyu:
@davort I can take a look at this :)

davort:
Hi @LarrySimiyu - that's great! Let me know if you'd need anything else re: design before you begin 😊

vidvidvid:
@davort some of the stuff is already fixed, but i'll make sure to check what else is missing and include the solutions in #1104

vidvidvid:
@LarrySimiyu unless you plan to deliver this anytime soon? :)

LarrySimiyu:
@vidvidvid @davort so sorry I got caught up with some major family stuff past month and job applications. I apologize for my lack of communication since picking this ticket up. Won't happen again. @vidvidvid Thank you for picking this up.

vidvidvid:
@LarrySimiyu np, life happens - if possible just let us know you about it next time :)

vidvidvid:
Solved in #1104

Issue #962, opened 2021-12-09
Several UI and Developer Experience improvements


firattale:
### What would you like to be added?

1-) Add a Custom 404 Page #1013 
2-) Scrolling in the Homepage creates a white blank at top and at bottom. We need to prevent that or change the color to have better color consistency. https://my.metagame.wtf/players #1104 
3-) It is good to have path aliases in the import of other components. That would save time to to copy other component's imports or changing any file location. #1082 

### Why is this needed?

1-)  We have a regular Next Js 404 Page, It would be good to use a custom 404 Page, But design needed.

<img width="1440" alt="Screenshot 2021-12-08 at 17 26 48" src="https://user-images.githubusercontent.com/34514974/145370069-053b9812-39c3-4c24-a8ed-73688f639dd2.png">
2-)
<img width="1440" alt="Screenshot 2021-12-08 at 17 26 06" src="https://user-images.githubusercontent.com/34514974/145370436-cd9fc4a2-555e-40fc-8615-a80fee08dbb7.png">
3-)

vidvidvid:
Nice suggestions. May I ask what is the url of the 'home page' you're referring to?

firattale:
hi @vidvidvid , it is happening here https://my.metagame.wtf/players

vidvidvid:
ohh, but this is just standard website design, i believe. it's happening on pretty much every website on mac/chrome. we could look into how to change that, but i don't think it needs fixing. just check github, for example, same thing happens here:
![white](https://user-images.githubusercontent.com/35112939/145374357-c4c49030-f577-42fb-a328-558d657eac54.gif)

firattale:
Yes @vidvidvid , that's true, but it is white because in light mode, when you switch to dark mode, that area becomes dark and consistent with theme UI. So we can change it to Metagame purple color.

vidvidvid:
ahhh, true!

vidvidvid:
the file for setting the theme can be found in `packages/design-system/src/theme/index.ts`

mquellhorst:
Could you also look into the flickering effect the animations have in the profiles in this PR?

https://user-images.githubusercontent.com/25387592/146056177-6e5cf2b4-d0ed-4394-ba46-b79d25c5ff4a.mov

firattale:
@mquellhorst this might happen because of my recent PR, but haven't seen anything like that in my profile, I will check that.

firattale:
[...]

PR #1104, opened 2022-02-03
[WIP] UI improvements

### LINKED ISSUE #1101: Unwanted border when hovering items in MegaMenu
### What happened?

When hovering a megamenu item this unwanted grey border appears:

Screenshot 2022-02-01 at 22 39 53

No border when hovering:

Screenshot 2022-02-01 at 22 41 20

1. Go to https://test.metagame.wtf/community/players 
2. Open any of the megamenu sections
3. hover one of the items

Not only the new border appeared, the paddings are off and the background color of the hovered item is different than the one in our production version: https://my.metagame.wtf/community/players

vidvidvid:
## Overview

Gonna tackle multiple smaller UI improvements in this PR.
- #1101 
- triangle of the megamenu is not touching the open modal section
- #1035 
- #962 / 2) Scrolling in the Homepage creates a white blank at top and at bottom. We need to prevent that or change the color to have better color consistency. https://my.metagame.wtf/players
- #1003 

Closes #1101 

### Megamenu Before
Screenshot 2022-02-03 at 11 50 07

### Megamenu After
Screenshot 2022-02-03 at 11 50 14


davort:
First of all, amazing work @vidvidvid 🙌
Re: buttons for "Edit layout" flow →

1. As a rule of thumb, button labels should always _state the action_ that happens when the button is pressed - not a noun or adverb. So, in case of "Default" it would be better for the label to say "Reset to default". It is a 3-word label, but we have enough room to accommodate for it. Better to keep it clear for the user.
2) Once "Reset to default" is pressed, and the layout is reset, that button isn't needed anymore 🙂 Best to hide it until the layout is changed again. That said, and to avoid the UI from "jumping around", consider moving this button to the far left.
3) The icon next to "Cancel" might be misinterpreted. It's a trash can. Perhaps an "X" would be a better choice?

Don't hesitate to ping me again should you have any questions. Thanks!

packages/web/components/Container.tsx
```diff
@@ -9,10 +9,10 @@ export const PageContainer: React.FC<Props> = ({ children, ...props }) => (
     bgSize="cover"
     bgAttachment="fixed"
     w="full"
-    minH="100vh"
-    px={[4, 8, 12]}
-    pt={[4, 8, 12]}
-    pb={[20, 20, 20, 12]}
+    h="calc(100vh - 84px)"
```
dan13ram, line 12: imo we should refrain from using px, and use rem

packages/web/pages/player/[username].tsx
```diff
@@ -453,6 +460,21 @@ export const Grid: React.FC<Props> = ({
               Cancel
             </MetaButton>
           )}
+          {changed && canEdit && (
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:463]: instead of adding a new button here, could we add a pop over menu with three vertical dots ? 
this kinda clutters the UI

Comment (vidvidvid) [reply packages/web/pages/player/[username].tsx:463]: i thought the same, but to quote @davort :
> It is on the longer side, but 3 words is still quite conventional, IME.
Another way to go is to just use icons, with tooltips showing up on-hover → like in Gmail or Google Drive
Right now, we only display these few options, so having a longer button label, and thus a wider button, is completely fine.
I prefer to err on the side of clarity for the user in this case, as long as we have enough space to afford it 🙂

Comment (vidvidvid) [reply packages/web/pages/player/[username].tsx:463]: the suggestion is to change `Default` -> `Reset to default`

vidvidvid:
The deployment is failing due to this error:
```
***
  errors: [
    ***
      extensions: [Object],
      message: 'Uniqueness violation. duplicate key value violates unique constraint "profile_username_key"'
    ***
  ]
***
***
  extensions: ***
    path: '$.selectionSet.insert_profile.args.objects',
    code: 'constraint-violation'
  ***,
  message: 'Uniqueness violation. duplicate key value violates unique constraint "profile_username_key"'
***
```
@dysbulic any idea what's causing this/how to solve this?

packages/web/pages/player/[username].tsx
Comment (dan13ram) [reply packages/web/pages/player/[username].tsx:462]: I see alright then :)

dan13ram:
@vidvidvid We need to fix this "Reset to default" button for mobile view as well. Currently it doesn't look great.. 
Also lets change the icon for Cancel from "trash-can" to "cross" (x) as suggested by Dave.

vidvidvid:
@dan13ram will fix it on mobile, thx for pointing it out. as for the x icon - this must have gotten messed up when rebasing, will fix it.

vidvidvid:
@dan13ram i replaced the cancel button icon, that's how it looks like on mobile now (not perfect, but better imo):
Screenshot 2022-02-06 at 14 47 11

davort:
> @dan13ram i replaced the cancel button icon, that's how it looks like on mobile now (not perfect, but better imo): Screenshot 2022-02-06 at 14 47 11

Hey @vidvidvid - 
1. "Reset to default" might work well here even if it just said "Reset". We might even consider keeping it like that for all layouts. 
2. Which leads me to the following question: what if the user presses the button by mistake? Both "Reset" and "Cancel" are destructive actions, so might be good to add some friction – a modal asking for confirmation of the pressed button's action.

What do you think?

vidvidvid:
@davort sure! I'm up for replacing `reset to default` to just `reset` and adding a confirmation dialog makes sense. will follow up with the update today or tomorrow.

vidvidvid:
@davort here's the update:
confirmation_modal

dan13ram (approved):
I wonder if @davort has any thoughts on the white background of the confirmation modals, but otherwise everything awesome! 

You do have to rebase with `fix/layouts` and lets merge #1105 before merging this one @vidvidvid

vidvidvid:
> You do have to rebase with `fix/layouts` and lets merge #1105 before merging this one @vidvidvid

Alright, will take care of it!

davort:
> I wonder if @davort has any thoughts on the white background of the confirmation modals, but otherwise everything awesome!

Let's leave it as is for now. Not a bad thing that it sticks out so much 😉

PR #1105, opened 2022-02-04
Bunch of UI fixes

### LINKED ISSUE #1026: Show selected role in profile

We have added role selection to the setup profile flow, but it's currently not shown anywhere. 

### Why is this needed?

So folks have a visual of their own profile details

### LINKED ISSUE #1041: Embedded-URL blocks don't render
### What happened?

I tried adding a block on test.metagame.wtf. It seems that it is trying to pre-render the typed URL in the little window but it doesn't load. In the console I see e.g. 

```
GET https://rlp-proxy.herokuapp.com/v2?url=ht
[HTTP/1.1 400 Bad Request 67ms]
```

https://user-images.githubusercontent.com/1402582/149390659-81f36ea7-9e44-45fa-b97a-181cffd6b9f6.mov

I didn't expect any preview, but when the little window showed, I expected it to show a preview of the rendered URL. 

https://test.metagame.wtf

Couple of other suggestions:
- Make the modal a fixed height so it doesn't bounce around when typing
- Debounce the preview window to only render every second or two rather than for each keystroke

dan13ram:
## Overview

- [using next api for fetching metadata + eslint updates](https://github.com/MetaFam/TheGame/commit/cf101ec87458dd0fe935a203002d86ae3c1a7c42)
    - closes #1041
    - moved to using next api and html-metadata-parser to obtain og tags for EmbeddedUrlSection
    before: 
embed-before
    after:
embed-after
     
- [can edit roles from profile page](https://github.com/MetaFam/TheGame/commit/33ce796acf810cf0d90bd20506ee8a901a7b673c)
    - closes #1026 
    - added support to edit roles directly from profile 
roles

- [fixed high level layouting](https://github.com/MetaFam/TheGame/commit/da7b739caf9169e655358623dcfb0665f734f248)
    - fixed bugs in MegaMenu and iframe pages wrt padding, scrolling and general layout issues
    - iframe pages had unnecessary padding and small height
    - some pages had content at the bottom cut off due to mobile megamenu
    - created new components `MegaMenu` in https://github.com/MetaFam/TheGame/pull/1105/files#diff-2c9573e05d28aaee8d6fc645e4d9ee8411393ac2eacacdecf18c441f4006b98e and `EmbedContainer` in https://github.com/MetaFam/TheGame/pull/1105/files#diff-5f25970ce29022bab63c20d6879f0d047e041287adb5cf79ddb510572a223b26R37
    
- [fixed default player profile layout](https://github.com/MetaFam/TheGame/commit/c0a3a10ea4bac71a34f31141a52416667e4543e6)
    - after conferring with peth, changed the default profile layout to include only hero, dao memberships, nft gallery and skills.


vidvidvid:
awesome stuff! :D will review the pr tomorrow.

packages/web/components/Player/Section/PlayerHero.tsx
```diff
@@ -181,11 +191,14 @@ export const PlayerHero: React.FC<Props> = ({
...
-              {player.profile.emoji}
-            </Text>
+            <Text fontSize="1.25rem">{player.profile.emoji}</Text>
           </PlayerHeroTile>
         )}
+        {/* player?.profile?.colorMask && (
+          <PlayerHeroTile title="Color Disposition">
+            <ColorDispositionDisplay {...{ person, personalityInfo }} />
+          </PlayerHeroTile>
+        ) */}
```
vidvidvid, line 201: what's up with that part? was this ever even displayed?

packages/web/components/Player/Section/PlayerHero.tsx
```diff
@@ -314,3 +327,47 @@ const TimeZoneDisplay: React.FC<TimeZoneDisplayProps> = () => {
     </Flex>
   );
 };
+
+export const ColorDispositionDisplay: React.FC<ColorDispositionProps> = ({
```
vidvidvid, line 331: is this used anywhere?

packages/web/components/Player/Section/PlayerHero.tsx
dan13ram: ah. nice catch @vidvidvid didn't realise this got in. 
so peth asked me to completely remove a separate section for "color decomposition" and add it to player hero section itself.

but it's a slightly longer than just that as it involves supporting editing it as well in the `EditProfileForm` that dysbulic has built. 

so I just left this in so I can uncomment it later on. Can remove it if you think it's unnecessary

packages/web/components/Player/Section/PlayerHero.tsx
dan13ram: see reply to above

packages/web/components/Profile/EmbeddedUrlSection.tsx
```diff
@@ -7,191 +7,122 @@ import {
...
-        borderRadius,
-        textAlign,
-        margin,
-        backdropFilter: 'blur(7px)',
-        borderWidth: '0',
-        overflow: 'hidden',
-        '&:hover': {
-          cursor: 'pointer',
-        },
+      onClick={() => window.open(url ?? inputUrl, '_blank')}
```
vidvidvid, line 90: i think this is causing the double tab opening

packages/web/components/Profile/EmbeddedUrlSection.tsx
```diff
@@ -7,191 +7,122 @@ import {
...
-              a: {
-                color: 'white',
-              },
-            },
-          }}
-        >
-          <h3>
-            <LinkOverlay href={url} isExternal>
+        <Box p={4}>
+          <LinkOverlay href={url ?? inputUrl} isExternal>
```
vidvidvid, line 113: and this

vidvidvid (changes requested):
Great work! It's just that when for example I add a youtube link to the profile, it opens in two separate tabs when I click on it, could you check what's up with that?

Availability got stretched in the setup wizard:
Screenshot 2022-02-06 at 13 05 10

It would probably be good to set up a modal footer, which was always visible, because for example the footer of color disposition is not visible until i scroll down on my screen. But we could look into solving this in another issue.
modal

packages/web/components/Player/Section/PlayerHero.tsx
vidvidvid: np, we can leave it there. so if i understand correctly we're just removing the color composition all together until we find a way to support editing it in the edit form? could we just add it to the player hero section and add an edit button next to it? then we don't have to find a way to edit it in the `EditProfileForm`/

dan13ram: I'm leaving this extra component I wrote so it can be used once we figure out how to integrate color disposition as well into the EditProfileForm. I think it should lie in the same modal because all that information is stored on Ceramic. But I might be wrong here, let me know what you think @vidvidvid 

Otherwise there are no changes, people can still add/remove ColorDecomposition as part of their profile.

dan13ram:
@vidvidvid Thanks for your review! I've fixed those couple things that you've highlighted :)

packages/web/components/Player/Section/PlayerHero.tsx
vidvidvid: Cool, let's leave it there! :)

vidvidvid (approved):

PR #1106, opened 2022-02-06
feat: show reset to default button only when changed

dan13ram:
## Overview

Adds an extra variable `isDefaultLayout` in `player/username` page which returns if the current layout is the same as default layout or not.

### DIFF
--- a/packages/web/utils/layoutHelpers.ts
+++ b/packages/web/utils/layoutHelpers.ts
@@ -1,0 +1,157 @@
+import {
+  getBoxLayoutItemDefaults,
+  GRID_ROW_HEIGHT,
+  LayoutItem,
+  ProfileLayoutData,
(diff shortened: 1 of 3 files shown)


packages/web/utils/layoutHelpers.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+    if (i.y === j.y) {
+      return 0;
+    }
+    return -1;
+  }
+  return -1;
+};
+
+export const isSameLayouts = (
```
dan13ram, line 131: @vidvidvid This the method I've added. Most of the other changes are moving stuff around and clean up.

vidvidvid (approved):
Great that you've separated the helper functions, now the [username] page is much cleaner :)

PR #1115, opened 2022-02-08
Miscellaneous extraneous changes pulled from #1078

### LINKED ISSUE #1098: Connect button doesn't change to profile icon when a new user connects
### What happened?

Generally when a user hits the Connect button and authenticates, the Connect button is replaced by their profile icon.

When a user that isn't in the system connects for the first time, the Connect button disappears but then comes back.

If the page is reloaded the profile icon appears.

The Connect button to change to the profile picture.

Connect to the site with an ETH account that isn't in the system.

@dan13ram has suggested this may be because the `useUser` hook doesn't send an updated value for `user` after the connection process is complete.

_No response_

dysbulic:
## Overview

#1078 is at about 4,000 lines changed and includes both code implementing features and unrelated functional and stylistic changes. This set of changes is about ⅓ of the total and should be largely independent.
    
These changes include:
    
* have Dependabot ignore `patch` version changes
* switch the cache invalidation Hasura action to asynchronous
* add a uniqueness constraint on the combination of player and skill ids
* print a comprehensible error message in the db seeding script on network errors
* add a mutex on the account creation process & allow it to happen only once *(subsequent runs were attempting to create another account with the same info & were failing b/c of the uniqueness constraint on the ETH address & returning `null` (closes #1098))*
* adds some more profile editing related types
* renames `HighLow7dType` to `HighLowType`
* refactors the `ColorBar` component to include links to a color explanation
* replaces `fake-tag`'s `gql` string identifier with the dependency-free `/* GraphQL */`
* rename `lib/hooks/useUserXp.ts` to `useUserXP.ts`
* fixed some typings to remove `any`s
* bumped some library versions & fixed the version numbers of some others

Closes #1098 

## Implementation

These are all relatively straightforward changes. The only even slightly complex code is in the refactor of the `ColorBar` where I created a one-off `Images` component which takes a function as a child *(a somewhat cleaner alternative to a `render` attribute)*.

There's also the reentrancy guard I put on calls to `createPlayer`. The first call sets a state variable to `creating` and all subsequent calls just randomly timeout w/ a linear increase in time until the original call sets the state to `completed`. At that point they look up the newly created player id.


packages/backend/src/handlers/auth-webhook/users.ts
```diff
@@ -4,24 +4,64 @@ async function createPlayer(ethAddress: string) {
   const { insert_profile: insert } = await client.CreatePlayerFromETH({
     ethereumAddress: ethAddress,
   });
-  if (insert?.affected_rows !== 1) {
-    throw new Error('Error while creating player');
+  if (insert?.affected_rows !== 2) {
```
vidvidvid, line 7: how come the condition for throwing an error changed from `insert?.affected_rows !== 1` to `insert?.affected_rows !== 2`?

packages/web/graphql/mutations/guild.ts
```diff
@@ -1,7 +1,7 @@
-import gql from 'fake-tag';
+export {};
```
vidvidvid, line 1: what exactly does this do?

vidvidvid (approved):
Looks great! Can you just please explain why did you choose to rename `colorDisposition` to `magicDisposition` at some places and leave it as it is at some other places?

packages/web/graphql/mutations/guild.ts
dysbulic: Since the file doesn't have any statements in it that is required for it to compile as a ESM module.

`yarn typecheck` told me I needed it.

dysbulic:
> [W]hy did you choose to rename `colorDisposition` to `magicDisposition` at some places and leave it as it is at some other places?

`magicDisposition` is the key in the Ceramic extended profile because "color disposition" is too ambiguous in that context.

`colorDisposition` is the name that is used throughout the profile display and editing flow. The goal was to not associate it too strongly with Magic: The Gathering.

PR #1116, opened 2022-02-08
Clean up and simplify the type system

dysbulic:
## Overview

* switches the type used for player / user objects from `PlayerFragmentFragment` to `Player`
* adds `dedupeOperationSuffix` to `codegen.yml` which causes the `…FragmentFragment` types to just be `…Fragment`
* removes the `player` variable from the return result of `useUser`; the `user` that is returned is now what `user.player` was
* moves `useUser` and `useWeb3` into their own files which are reexported from `index.ts` to deal with a circular dependency issue
* finishes the replacement of `fake-tag`'s `gql` with `/* GraphQL */` begun in #1115

These type changes are to clarify and simplify the next set of changes which are a significant revamp of the profile setup wizard as well as the addition of Jotai for local state management. As such, there is no specific issue created for this.

I'd like to see

```yaml
namingConvention:
  transformUnderscore: true
```

added to the `codegen.yml`. It would cause the removal of underscores in the generated types, bringing them more in line with all the other types in the system.

The problem is there are both `PlayerRole` *(holds the values of an enum)* and `player_role` *(joins the `player` and `role` tables)* tables. With `transformUnderscore` turned on, both these tables generate the same type names.

`PlayerRole` would need to be renamed, likely to just `Role`.

## Implementation

This was mostly alot of search & replace. There are lots of changed lines, but it's all naming and very little logic.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,6 @@ updates:
     directory: / # default location of `.github/workflows`
     schedule:
       interval: daily
+    ignore:
+      - dependency-name: '*'
+        update-types: [version-update:semver-patch]
--- a/.github/workflows/PR-CI.yml
+++ b/.github/workflows/PR-CI.yml
@@ -17,18 +17,18 @@ jobs:
 
     steps:
       - name: Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.4.0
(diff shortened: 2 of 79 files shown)


vidvidvid (approved):
Very nice changes, they add a lot to brevity of the code.

PR #1124, opened 2022-02-08
Add Jotai and refactor Ceramic interaction to a hook

dysbulic:
## Overview

This is pretty much everything from #1078 except for the panels in the setup wizard and editing modals for the player's profile.

The next PR which will include the actual content panes is the one that will actually close issues. This is just the necessary infrastructure.

Currently, `ProfileWizardPane` and `GenericWizardPane` are extremely similar to one another. I wrote the profile pane & then needed a more general component to handle skills which aren't stored in Ceramic and aren't stored in the `profile` table. The next pull request, #1127, will include a refactor to remove much of the repeated code and have the profile pane leverage the general purpose one.

## Implementation

Much of the meat of this PR is in the `useProfileField` and `useSaveCeramicProfile` hooks. The former accesses the properties of a player object with the ability to override those values with ones stored in Jotai. The later encapsulates the logic for dividing profile fields into writes to the appropriate streams *(there are two separate streams for basic & extended profile information)*.

This PR also includes the `ProfileWizardPane` component which handles loading values and both saving them to Ceramic and to Jotai.

I backported a change from #1127 to give the deployment environment for the gcloud instance a PR-specific name. When a new run happens, it deactivates any existing deployments in the same environment. This was causing deployments to be removed cross-pull request.


packages/design-system/src/StatusedSubmitButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  status = null,
+  ...props
+}: {
+  label?: Maybe<string>;
+  status?: Maybe<string | ReactElement>;
+}) => (
+  <MetaButton
+    type="submit"
+    _focus={{ filter: 'brightness(1.75)' }}
+    sx={{ ':hover': { filter: 'rotate-hue(-90)', border: '2px solid green' } }}
```
dan13ram, line 18:
```suggestion
    _hover={{ filter: 'rotate-hue(-90)', border: '2px solid green' }}
```

packages/web/components/MegaMenu/MegaMenu.tsx
```diff
@@ -114,9 +118,10 @@ const Logo = ({ link }: LogoProps) => {
         _active={{ bg: 'transparent' }}
       >
         <Image
-          src="/assets/logo.png"
-          height={`${height}px`}
-          width={`${width}px`}
+          src={LogoImage}
+          w={`${width}px`}
+          h={`${height}px`}
```
dan13ram, line 123: Why is this in pixels? imo we should move to `rem`

packages/design-system/src/StatusedSubmitButton.tsx
dysbulic: Yeah, I got this in #1127. I was trying to figure out why this wouldn't work by using the `sx` attribute rather than `_hover`. It's because  `rotate-hue` is supposed to be `hue-rotate`.

packages/utils/src/did/index.ts
```diff
@@ -22,38 +23,40 @@ export async function createToken(
...
-    console.error('Token verification failed', e);
-    return null;
+  connectedAddress?: string,
+): Promise<Maybe<Claim>> {
+  const rawToken = Base64.decode(token, 'base64');
+  const [proof, rawClaim] = JSON.parse(rawToken);
+  const claim: Claim = JSON.parse(rawClaim);
+  const claimant = claim.iss;
+
+  if (connectedAddress != null && claimant !== connectedAddress) {
```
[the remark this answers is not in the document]
dysbulic: I thought about this. The address used to generate `claimant` is retrieved in the same way as the address used to generate `connectedAddress`, so regardless of whether the wallet checksums or not they both have the same capitalization.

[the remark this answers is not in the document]
dan13ram: Sounds good. Its great that you already thought about it :D

dan13ram (changes requested):
@dysbulic I'm unable to update my color decomposition from the player profile page.
colors

Could you take a look and fix this?

dysbulic:
> I'm unable to update my color decomposition from the player profile page.

This PR doesn't include the revised content panes for editing profile fields. Those are in #1127 *(color disposition editing works for that one, I'm still finishing with skills)*.

The only way to pull that functionality into this request would be to add another 3,000+ changed lines.

dan13ram (approved):
Thanks for clarifying ! :) These hooks look absolutely awesome

Issue #994, opened 2021-12-27
Update player profile display after editing


dysbulic:
### What would you like to be added?

When a player edits their profile, the system makes a call which dirties the Hasura cache for that user. The caching process [downloads the streams](https://github.com/MetaFam/TheGame/issues/921#issuecomment-1001332960) and updates the cache.

This introduces a delay between when a user hits "`Save`" on the form and when their profile page updates.

Next.js caching of statically compiled pages exacerbates this problem further to the point that it can take several minutes before a page refresh will show the updated information.

Some sort of state management needs to be introduced, be it prop drilling, `useReducer`, or some other library to update the profile page in the browser when edits are made. 

### Why is this needed?

The current behavior would likely lead users to believe that their changes weren't saved correctly because the profile shows the previous values for several minutes before updating.

Issue #1114, opened 2022-02-07
Players without a row in the profile table can't log in


dysbulic:
### What happened?

@vidvidvid was trying to log in, and the connect button would switch to a spinner for a second, then back to the button.

Further investigation revealed the result of the `getMe` query was returning `null`. This is because a player's username is now located in the profile table, the `me` table is a join between the player & profile tables, and @vidvidvid didn't have a row for their player id in the profile table.

### What did you expect to happen?

To be able to log in.

### How can we reproduce the problem (as minimally as possible)?

Open the Hasura console, delete the profile row with your player id, and lose the ability to log in.

### Is there anything else we need to know?

This should be simple to fix with a Hasura migration which creates rows in the profile table for all the player ids *(if they don't already exist)*.

`yarn hasura console` → `Data` Tab → `SQL` in the left-hand column → Check "`This is a migration`" → Write the needed SQL

dysbulic:
> The 201 was causing problems, which was why new users couldn't login.

@dan13ram, that's a good catch, but that isn't the basis for this ticket. For one, I made the change to `auth-webhook` after seeing this bug, but, more importantly, this fix is about new users, my problem is with users that already have an entry in the 'player' table.

if you connect to [api.metagame.wtf](//api.metagame.wtf), you'll note there are 887 entries in the `player` and `me` tables, but only `885` in the `profile_cache` table. Once `develop` is merged, `profile_cache` will become `profile` & there'll only be, at most, 885 entries in `me` because it is a composite view of a join of the `player` and `profile` tables. The `me` table is what is used to drive `useUser`.

The statement `select username from player where id not in (select player_id from profile_cache)` shows the two users that would be unable to log in.

The SQL we need is `insert into profile (player_id) select id from player where id not in (select player_id from profile)`. I'll add a migration to #1127.

Issue #1097, opened 2022-02-01
Users connecting for the first time should be directed to the profile setup wizard


dysbulic:
### What happened?

When a user connects for the first time an account should be created for them in Hasura & they should be directed to the profile setup wizard.

Currently that does not happen.

### What did you expect to happen?

New users should be directed to the profile wizard, they aren't.

### How can we reproduce the problem (as minimally as possible)?

Sign in to the site with an ETH account that isn't in the system.

alalonde:
This is broken on prod as well.

dysbulic:
> This is broken on prod as well.

@alalonde, did this ever work?

I'm leaning toward pulling down the Ceramic profile in the browser for new MyMeta users so that I only request they make a profile if they haven't elsewhere. Also new users won't be cached yet, so we run the risk of overwriting an existing profile (#1132) if we don't check.

What I don't know how to do is how to tell when a user is new to MyMeta. I switched up `/auth-webhook` *(where users are created)* to return a status of 201 instead of 200 when a user is created. I've not been able to find any evidence of who calls that endpoint though.

Is it called by Hasura internally & passed the bearer token sent to the GraphQL endpoint by the browser?

Maybe I'll get a non-zero `affected_rows` only for situations where a user was created.

I added a reentrancy check to prevent multiple calls to `createPlayer` because if they're on a profile page when they log in there's four or five calls in quick succession & some were making it into `createPlayer` and failing the uniqueness check on ETH address trying to create the same account multiple times. *(This was the reason for #1098.)* This means only one call will know a new user was created & I guess I'll have to put the logic to push them to the wizard in the `useUser` hook… Except any query might be the one to cause the creation of a new user, not just `useUser`…

I could always just check their profile & determine if it's empty (`createdAt === updatedAt`?). If they skip out of the profile creation process without entering anything, I could touch `updatedAt` to keep them from being asked again…

alalonde:
> What I don't know how to do is how to tell when a user is new to MyMeta. I switched up /auth-webhook (where users are created) to return a status of 201 instead of 200 when a user is created. I've not been able to find any evidence of who calls that endpoint though.

Yes, I believe this is called by Hasura.

> I could always just check their profile & determine if it's empty (createdAt === updatedAt?). If they skip out of the profile creation process without entering anything, I could touch updatedAt to keep them from being asked again…

[...]

Issue #1096, opened 2022-02-01
The /me page is broken in develop


dysbulic:
### What happened?

[test.metagame.wtf/me](//test.metagame.wtf/me) should display the profile of the current user, but instead it is returning a 500 server error.

Initial investigation suggests that it has something to do with the `PersonalityInfo` object not being correctly passed to the color disposition component.

### What did you expect to happen?

The page not to error out.

### How can we reproduce the problem (as minimally as possible)?

Visit [test.metagame.wtf/me](//test.metagame.wtf/me).

### Is there anything else we need to know?

When testing I was getting a “different number of hooks run” error, but it went away as I was tracing the execution though I didn't knowingly fix it.

PR #1127, opened 2022-02-09
Refactor Setup Wizard & Profile Modal Configuration Panes

### LINKED ISSUE #887: Optimistic updates for profile editing

dysbulic:
## Overview

* These are the configuration panes for editing profile fields in the setup wizard & the profile display. They utilize the `ProfileWizardPane` component which takes a render function as a child.
* Also the browser was hanging when loading the profile details which forced me to go digging into the mass of spaghetti that is the editable profile layout. I'm still not happy with how it it done, but I simplified some of the variable names & eliminated quite a bit of code in the configuration steps.
* When a user types into the fields in the profile hero editing interface the input will grow in size to accommodate their text up to the width of the modal. *(I wanted to resize the input on load, but a `text` `input` doesn't have an `onLoad` event & my experiments with `ref`s proved fruitless.)*
* I changed the `BoxType` type from an `enum` to a `const {} as const` object which can be iterated over.
* I eliminated a couple of the props that someone was drilling through 3–4 layers of components by moving their creation closer to where they are used.
* I tweaked the UI in several places where elements were out of line or overflowing their containers.
* The switch to loading profile data via a hook broke the fade out/in animation for values changes. It was very rarely going to be triggered and is currently still inoperative.

Closes #887 & #994 & #1168.

There is a page reload between each of the profile setup wizard's screens. This both wipes the local cache quicker than Next.js can respond with a new page and it causes Ceramic to demand credentials again for every page. It should be relatively simple to cycle through the components without a page load.

Also, `useUser` is called alot. Once values have been written into Jotai, the `useUser` values no longer have any effect. It should be possible to use an `atom` to cache a copy of the `user` object and potentially significantly improve performance. Ideally, MyMeta would subscribe to changes to its GraphQL query for a user's player record so that it could be notified of any changes and propagate them.

## Implementation

One interesting component set are the fields in the player hero. I am using a `Wrap` element to contain them, so each one needs to be contained in a `WrapItem` to perform correctly. When some of the elements aren't present, however, I return `null` so nothing is rendered. A `WrapItem` around a `null` will throw off the `space-between` justification on the `Wrap`. So, I needed a conditional inclusion and accomplished it with a `Wrapper` attribute that is `React.Fragment` by default, but which can be set to `WrapItem` for the elements that need it.

---

Paired With: @tenfinney


vidvidvid:
@dysbulic could you perhaps fit into this PR also https://github.com/MetaFam/TheGame/issues/1097 ?

dysbulic:
> [C]ould you perhaps fit into this PR also: #1097?

To deal with a first time user, I'm going to need to load their Ceramic profile in the browser lest I inadvertently overwrite it (#1132) *(also, they won't be in the cache)*.

I'm still uncertain how to tell that a new user was created. I've got [some ideas](//github.com/MetaFam/TheGame/issues/1097#issuecomment-1040050047), but it's going to take some testing to figure it out.

I've still got to format the various panes of the setup wizard & fixing #1096 is a higher priority than directing users to the wizard. I'll be surprised if #1097 makes it into Friday's release.

packages/design-system/src/MetaButton.tsx
```diff
@@ -16,8 +16,9 @@ export const MetaButton: React.FC<
     fontSize="sm"
     bg="purple.400"
     color="white"
-    {...{ ref }}
-    {...props}
+    transition="background 0.5s, filter 0.75s"
```
dan13ram, line 19: I would reduce the transition duration to 0.25s and 0.5s. Also is this necessary and/or mentioned by the design team?

packages/web/pages/player/[username].tsx
```diff
@@ -87,58 +90,81 @@ export const PlayerPage: React.FC<Props> = ({
...
 export default PlayerPage;
 
-const getBoxKeyFromTarget = (target: HTMLElement | null): string =>
-  (target?.offsetParent as HTMLElement)?.offsetParent?.id ?? '';
+const makeLayouts = (editing: boolean, layouts: Layouts): Layouts =>
+  Object.fromEntries(
+    Object.entries(layouts).map(([key, items]) => [
+      key,
+      items.map((item) =>
+        item.i === 'hero' ? { ...item, isResizable: editing } : item,
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:123]: This seems like a rebase error to me. This method is not necessary and not used in `develop` branch

Comment (dan13ram) [reply packages/web/pages/player/[username].tsx:123]: All these helpers were moved to `utils/layoutHelpers.ts`

packages/web/pages/player/[username].tsx
```diff
@@ -345,54 +360,26 @@ export const Grid: React.FC<Props> = ({
...
-              onClick={() => setExitAlertCancel(true)}
-              leftIcon={<CloseIcon />}
+              onClick={handleCancel}
+              leftIcon={<DeleteIcon />}
             >
               Cancel
             </MetaButton>
           )}
-
-          <ConfirmModal
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:383]: These modals shouldn't be removed?

packages/web/pages/player/[username].tsx
```diff
@@ -345,54 +360,26 @@ export const Grid: React.FC<Props> = ({
           justifyContent="end"
           variant="ghost"
           zIndex={10}
+          isAttached
           h="3rem"
           mb="1rem"
         >
-          {changed && canEdit && !isDefaultLayout && (
-            <MetaButton
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:352]: We need this button as well.

packages/web/pages/player/[username].tsx
```diff
@@ -345,54 +360,26 @@ export const Grid: React.FC<Props> = ({
...
+              colorScheme="purple"
               _hover={{ background: 'purple.600' }}
               textTransform="uppercase"
               px={12}
               letterSpacing="0.1em"
               size="lg"
               fontSize="sm"
-              onClick={() => setExitAlertCancel(true)}
-              leftIcon={<CloseIcon />}
+              onClick={handleCancel}
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:377]: And we want CloseIcon instead of DeleteIcon here as per Dave's design.

packages/web/graphql/mutations/idxCache.ts
```diff
@@ -1,7 +1,5 @@
 export const InsertCacheInvalidation = /* GraphQL */ `
   mutation InsertCacheInvalidation($playerId: uuid!) {
-    updateIDXProfile(playerId: $playerId) {
-      success
-    }
+    updateIDXProfile(playerId: $playerId)
```
dan13ram, line 3: I think @alalonde had added the success param here?

packages/web/components/Setup/SetupMemberships.tsx
```diff
@@ -37,7 +37,7 @@ export const SetupMemberships: React.FC<SetupMembershipsProps> = ({
   return (
     <FlexContainer mb={8}>
       <MetaHeading mb={5} textAlign="center">
-        Memberships
+        Member&shy;ships
```
dan13ram, line 40: Is this a typo?

dan13ram (changes requested):
Great stuff bringing this together :D

But I think some old code has resurfaced which has reintroduced old bugs. As well as some new bugs perhaps.

Here, the loading state should show up in the center but its showing on the left side.
setup-skills-bug

Here, skills select is too long. Should ideally be shorted and inside window.
skills-select-too-long


Profile Hero section is somewhat borked => should be no title bar. Also there's an extra boxShadow at the bottom of the section.

How it looks now:
profile-hero

How it should look (as in test.metagame.wtf):
former-profile-hero

Also I'm trying to setup my profile but edits aren't updated on my profile. I'm testing on the gcp deploy => https://frontend-pr-1127-mjhnbmqqna-uk.a.run.app/

vidvidvid:
I also tested saving data through the setup wizard, but it didn't update it upon exiting the setup wizard. I also tried adding various boxes to the user profile, yet they didn't persist on refreshing.
Was also testing on https://frontend-pr-1127-mjhnbmqqna-uk.a.run.app/

packages/design-system/src/MetaButton.tsx
dysbulic: Removed.

packages/web/pages/player/[username].tsx
Comment (dysbulic) [reply packages/web/pages/player/[username].tsx:123]: Removed from `[tsername].tsx`.

packages/web/lib/hooks/useField.ts
```diff
@@ -3,11 +3,15 @@ import { ExplorerType, Player, Profile } from 'graphql/autogen/types';
 import { Atom, atom as newAtom, PrimitiveAtom, useAtom } from 'jotai';
 import { useMemo } from 'react';
 
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
dysbulic: `useUser` includes `useWeb3` which includes `Web3Context` which includes `clearJotaiState` from `lib/hooks/useField`.

`useField` also includes `useProfileField` which uses a global object in `useField` to keep track of which profile fields have Jotai overrides. The `clearJotaiState` simply resets the global object.

I could maybe up the global state object in Jotai and split up the files… I'll try.

---

Ok, so accessing a Jotai `Atom` is done using a hook, and clearing the state of the overrides is done from the `disconnect` method. It isn't possible to call a hook from a regular method, only another hook, so I can't easily make this work without the circular dependency.

If you have a suggestion as to an alternate way to implement the global state of the "override" values (values that have been changed in the UI, but which are still potentially stale in Hasura), I'm open to it,

packages/web/graphql/mutations/idxCache.ts
dysbulic: I switched this method to asynchronous to speed it up. It now returns a UUID that can be used to query for it's final value. That query method is in `schema.graphql`.

I originally had this as:
```graphql
    updateIDXProfile(playerId: $playerId) {
      updateIDXProfile: uuid
    }
```
But I get an error: “`GraphQLDocumentError: Field "updateIDXProfile" must not have a selection since type "uuid!" has no subfields.`” I'm pretty sure what I have is how you designate it returns a single field.

packages/web/pages/player/[username].tsx
Comment (dysbulic) [reply packages/web/pages/player/[username].tsx:377]: Changed.

packages/web/graphql/mutations/idxCache.ts
dan13ram: Ah i see. Thanks for clarifyng this! :D

hasura/migrations/1645463938163_guarantee_players_exist_in_the_profile_table/up.sql
```diff
@@ -0,0 +1,6 @@
+INSERT INTO profile (player_id)
```
dan13ram, line 1: @dysbulic this should be closing #1114 ?

packages/web/pages/player/[username].tsx
```diff
@@ -364,35 +389,34 @@ export const Grid: React.FC<Props> = ({
...
+              aria-label="Cancel Layout Edit"
+              colorScheme="purple"
               _hover={{ background: 'purple.600' }}
               textTransform="uppercase"
               px={12}
               letterSpacing="0.1em"
               size="lg"
               fontSize="sm"
-              onClick={() => setExitAlertCancel(true)}
+              onClick={handleCancel}
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:402]:
```suggestion
              onClick={() => setExitAlertCancel(true)}
```

hasura/migrations/1645463938163_guarantee_players_exist_in_the_profile_table/up.sql
dysbulic: ✅

dysbulic:
@dan13ram, when fetching a user's roles, `requestPolicy` is set to `network-only`.

If I'm reading [the blame](//github.com/MetaFam/TheGame/blame/develop/packages/web/components/Setup/SetupRoles.tsx#L45) correctly, you added that. Why?

dysbulic:
> Here, the loading state should show up in the center but its showing on the left side. setup-skills-bug

This should be fixed.

dysbulic:
> Profile Hero section is somewhat borked => should be no title bar. Also there's an extra boxShadow at the bottom of the section.

Removed title bar & inner box shadow.

packages/web/pages/player/[username].tsx
```diff
@@ -500,11 +535,9 @@ export const getStaticProps = async (
   }
 
   const player = await getPlayer(username);
-  const personalityInfo = await getPersonalityInfo();
```
Comment (dan13ram) [inline packages/web/pages/player/[username].tsx:503]: The reason I had moved this here was to reduce load time on frontend, if this static data could be loaded on server side itself. But never mind :)

Comment (dysbulic) [reply packages/web/pages/player/[username].tsx:503]: I figured that out when I saw the same pattern for skills.

The problem was it was drilled through three or four components completely unrelated to the color disposition, and was just out of place. Maybe we could get the same effect using an `Atom` or `Context` in the future without the prop drilling.

packages/web/components/Player/Section/PlayerHero.tsx
```diff
@@ -19,67 +20,49 @@ import {
...
   WrapItem,
 } from '@metafam/ds';
+import { Maybe } from '@metafam/utils';
 import BackgroundImage from 'assets/main-background.jpg';
 import { FlexContainer } from 'components/Container';
 import { EditProfileForm } from 'components/EditProfileForm';
 import { PlayerAvatar } from 'components/Player/PlayerAvatar';
-import { PlayerContacts } from 'components/Player/PlayerContacts';
+import { PlayerContacts as Contacts } from 'components/Player/PlayerContacts';
+import { PlayerBrightId as BrightId } from 'components/Player/Section/PlayerBrightId';
```
dan13ram, line 29: PlayerBrightId was removed from PlayerHero and made exclusive to `PlayerContacts`. Could we revert this to how it was? 

This is because we do not want to see "Verify On BrightId" right below the username and also its not working. So it should be disabled as it is in PlayerContacts.

packages/web/components/Player/Section/PlayerHero.tsx
```diff
@@ -95,90 +78,46 @@ export const PlayerHero: React.FC<Props> = ({
...
             }}
           />
         </Box>
       )}
-      <Box textAlign="center" mb={8} mt={2}>
+      <Box align="center" mb={8} mt={2}>
         <PlayerAvatar
-          w={{ base: 32, md: 56 }}
-          h={{ base: 32, md: 56 }}
+          w="min(var(--chakra-sizes-56), 100%)"
```
dan13ram, line 88: This might be causing some issues as seen in the screenshot below. ideally we want it to be a circle ?
Screenshot from 2022-02-28 16:22:59

packages/web/components/Player/Section/PlayerHero.tsx
dysbulic: I'll add it to [the list](//github.com/MetaFam/TheGame/issues/1146#issuecomment-1053911232).

packages/web/components/Player/Section/PlayerHero.tsx
dysbulic: Yeah, my fix for elements with default values wasn't quite right.

It's now keeping the default values (emoticons in this case) from loading.

I'm going to get back to it shortly…

tenfinney (approved):
LGTM.

PR #1148, opened 2022-02-17
Fix player login + other minor issues

### LINKED ISSUE #1022: Disconnect user on wallet account change

When a user changes their account in their web3 wallet, they should be disconnected from the site.

### Why is this needed?

People change their connected account in MetaMask they remain connected to the site with their old address. Web3Modal caches the connection information so the address will persist even across page reloads.

When a user changes the connected account, they should be disconnected and the cached information removed.

### LINKED ISSUE #1096: The /me page is broken in develop

### LINKED ISSUE #1114: Players without a row in the profile table can't log in

### LINKED ISSUE #1128: Landing page is borked
### What happened?

Huge backgrounds: https://test.metagame.wtf/landing

Normal sizes of slides.

https://test.metagame.wtf/landing

Looks like removing `min-height: 100%;` from parent components of slides solves the problem.

dan13ram:
## Overview

- Cleaned up Web3Context + auth webhook on the backend to fix login issues.
- Fixed /me page and landing pages
- Refactored MegaMenu components into separate files

Closes #1022
Closes #1096
Closes #1114
Closes #1128

## Implementation

**Side effects**

## Assets


dan13ram:
Login in Vercel deployment won't work since there is a small change in the backend. But should work on the gcp deployment.

vidvidvid:
@dan13ram yep, works fine!

there are a couple of changes that were made in a [recent PR](https://github.com/MetaFam/TheGame/pull/1124/files#diff-43c199e7f717ce03089f7b2568d1822baf78d11e4b7e86f861c979bb9a7204ac), which messed up the `MegaMenu` a bit. since you worked to `MegaMenu`, could you please squeeze in a fix for the following things?

1. the logo got bigger and is not centered anymore
logo

2. the hover effect is different now
hover

3. the paddings (presumably) made the menu modal bigger, could you also please revert that?

PR #1163, opened 2022-02-23
Display a guild's DAOs on their page and support more than one

### LINKED ISSUE #1142: [MyMeta] Add DAO info to guild pages
Right now, there are no links to guild's DAO even if they have one.

I think we should add a fiield for a link to "your DAO", "your multisig" &/or "your treasury" to the guild application form. 
Every guild that has a DAO on DAOHaus or a multisig, should also have a link on their profile connected to it.

alalonde:
## Overview

There are a few outstanding issues with our DAO / guild / player relationships, with the main end-goal being having a new set of tables (dao and dao_players) that we can use as a cache to avoid the very expensive daohaus queries we're currently making.  My plan for this is:

1. Create those two new tables, including a guild_id in the dao table so that guilds can associate one or more daos / multisigs / treasury addresses. Also add the relevant fields currently being returned from daohaus 
2. Refactor the current guild / moloch_address persistence code to instead persist 1 or more daos. This includes updating the guild create / edit form 
3. Add any of a guild's inputted DAO data to the guild page
4. Remove the current daohaus remote schema relationship for a player in favor of a more generic action to be called to update the cached data in the new dao tables
5. Implement the cache update using the current IDX player caching logic as a model

This PR tackles 1 - 3, 4/5 will be addressed in later PRs.

Closes #1142

## Follow-up Items

- Add another page `/guild/[guildname]/edit` rather than just linking to the `/join/guild` page. It works, but the UI is not totally accurate

### DIFF
--- a/hasura/migrations/1645304381138_move_guild_dao_addresses/up.sql
+++ b/hasura/migrations/1645304381138_move_guild_dao_addresses/up.sql
@@ -1,0 +1,4 @@
+INSERT INTO dao(contract_address, network, guild_id) 
+SELECT moloch_address, 'mainnet', id FROM guild WHERE moloch_address IS NOT NULL;
+
+UPDATE dao SET network = 'xdai' WHERE contract_address IN ('0x30c9aa17fc30e4c23a65680a35b33e8f3b4198a2', '0x5219ffb88175588510e9752A1ecaA3cd217ca783');
--- a/packages/web/components/Guild/GuildForm.tsx
+++ b/packages/web/components/Guild/GuildForm.tsx
@@ -15,9 +15,10 @@ import {
 import { SelectOption } from '@metafam/ds/src/MultiSelect';
 import { Field, FieldDescription } from 'components/Forms/Field';
 import { MetaLink } from 'components/Link';
+import { utils } from 'ethers';
 import {
   DiscordRole,
-  GuildDao,
+  GuildDaoInput,
   GuildFragment,
   GuildType_Enum,
   Maybe,
@@ -25,7 +26,12 @@ import {
 } from 'graphql/autogen/types';
 import { useRouter } from 'next/router';
 import React, { useEffect, useMemo } from 'react';
-import { Controller, useFieldArray, useForm } from 'react-hook-form';
+import {
+  Controller,
+  FieldError,
+  useFieldArray,
+  useForm,
+} from 'react-hook-form';
 
 const validations = {
   guildname: {
@@ -48,6 +54,7 @@ const validations = {
   },
   daoAddress: {
     required: true,
+    validate: (address: string) => utils.isAddress(address),
   },
   daoNetwork: {
     required: true,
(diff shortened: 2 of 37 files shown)


hasura/migrations/1645304381138_move_guild_dao_addresses/up.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO dao(contract_address, network, guild_id) 
+SELECT moloch_address, 'mainnet', id FROM guild WHERE moloch_address IS NOT NULL;
+
+UPDATE dao SET network = 'xdai' WHERE contract_address IN ('0x30c9aa17fc30e4c23a65680a35b33e8f3b4198a2', '0x5219ffb88175588510e9752A1ecaA3cd217ca783');
```
dan13ram, line 4: We should ideally be calling this `gnosis-chain` or `gc` instead of `xdai`. But this is a change we have to do in multiple places, so perhaps best tracked in a separate PR. 

https://github.com/MetaFam/TheGame/issues/1170

dan13ram: Let's also remember to double check these addresses are the only gnosis chain daos that we have onboarded so far while all of the rest are on mainnet, before we merged this to master.

packages/web/components/Guild/GuildHero.tsx
```diff
@@ -1,12 +1,50 @@
...
+  VStack,
+} from '@metafam/ds';
 import { ProfileSection } from 'components/Profile/ProfileSection';
 import { GuildFragment } from 'graphql/autogen/types';
 import React from 'react';
 
-type Props = { guild: GuildFragment };
+type Props = {
+  guild: GuildFragment;
+  canEdit: boolean;
```
dan13ram, line 17: @dysbulic on his PR has changed all occurences of `canEdit` with `editing`. So maybe good to rename this as well.

packages/web/pages/guild/[guildname].tsx
```diff
@@ -5,24 +5,49 @@ import { GuildLinks } from 'components/Guild/GuildLinks';
...
+  useEffect(() => {
+    if (
+      user?.id ===
+      getGuildsResult?.data?.guild_metadata.some(
+        (gm) => gm.guildId === guild?.id,
+      )
+    ) {
+      setCanEdit(true);
+    }
+  }, [user, getGuildsResult, guild]);
```
Comment (dan13ram) [inline packages/web/pages/guild/[guildname].tsx:50]: perhaps we can repace this useEffect and `canEdit` useState with a single useMemo hook.

packages/web/graphql/queries/guild.ts
```diff
@@ -63,6 +65,35 @@ export const getGuildMetadata = async (id: string) => {
...
+    guild_metadata(where: { creatorId: { _eq: $id } }) {
+      guildId
+      guild {
+        guildname
+      }
+    }
+  }
+`;
+
+export const getAdministeredGuilds = async (playerId: string) => {
```
dan13ram, line 79: Not used anywhere?

packages/web/components/Player/PlayerGuild.tsx
```diff
@@ -35,11 +36,21 @@ export const InternalGuildLink: React.FC<InternalGuildLinkProps> = ({
 
 type DaoHausLinkProps = {
   daoUrl: string | null;
+  linkProps?: Maybe<LinkProps>;
 };
```
dan13ram, line 40: imo it would be cleaner to send linkProps as part of the props itself. 
```
type DaoHausLinkProps = { daoUrl: string | null } & LinkProps;
```
and we can destrucute the remaining props as already done below.

packages/web/components/Guild/GuildLinks.tsx
```diff
@@ -1,68 +1,103 @@
...
-        </WrapItem>
-      ) : null}
-    </Wrap>
-  </ProfileSection>
-);
+export const GuildLinks: React.FC<Props> = ({ guild }) => {
+  const daoHrefs = useMemo(
+    () =>
+      guild.daos.map(
+        (dao) => dao.url || getDaoLink(dao.network, dao.contractAddress),
```
dan13ram, line 18: Are all guilds by default DaoHaus DAOs ?

packages/web/components/Guild/GuildForm.tsx
```diff
@@ -293,6 +314,124 @@ export const GuildForm: React.FC<Props> = ({
...
+                  </Box>
+                  <Box flex="1">
+                    <Field label="URL">
+                      <Input
+                        placeholder="https://"
+                        {...register(`daos.${index}.url`)}
+                        background="dark"
+                      />
+                      <FieldDescription>
+                        An optional URL to link this address to
```
dan13ram, line 370: We may want to make dao URL mandatory if this is not a DaoHaus dao. Perhaps a checkbox for the user to select if guild is DaoHaus dao or not?

dan13ram (commented):
Most everything looks good to me. Lets wait for #1127 to be merged first though.

hasura/migrations/1645304381138_move_guild_dao_addresses/up.sql
alalonde: > Let's also remember to double check these addresses are the only gnosis chain daos that we have onboarded so far while all of the rest are on mainnet, before we merged this to master.

I believe I pulled this data from the prod env, but will double check.

packages/web/graphql/queries/guild.ts
alalonde: removed

packages/web/pages/guild/[guildname].tsx
Comment (alalonde) [reply packages/web/pages/guild/[guildname].tsx:50]: good call

packages/web/components/Player/PlayerGuild.tsx
alalonde: Taken care of during the rebase

packages/web/components/Guild/GuildLinks.tsx
alalonde: No, but they can be associated with one or more

dan13ram (approved):

PR #1175, opened 2022-03-01
Fixed styles in Dashboard + other minor fixes

dan13ram:
[fix: fixed minor bugs in profile page](https://github.com/MetaFam/TheGame/commit/9cfbaf66525488e882f7e4f9bfa397dee57e7043)
Before: 
Screenshot from 2022-03-01 20:58:44
Screenshot from 2022-03-01 21:27:22
After:
Screenshot from 2022-03-01 20:59:50
Screenshot from 2022-03-01 21:27:10

[fix: fixed menu spacing](https://github.com/MetaFam/TheGame/commit/2ff7f63737aeda7a6f141a07749b6ad95966662a)
Before the menu would move after connection, now it does not: 
mega-menu

[feat: updated dashboard layout styles](https://github.com/MetaFam/TheGame/commit/8b6906bb63221c8e0fe340c462bbb3ca14156c09)
Dashboard has been updated with new styles. Do check it out! :)
Before: 
Screenshot from 2022-03-01 21:28:30
Screenshot from 2022-03-01 21:29:11


vidvidvid:
thank you for tending to my ocd and fixing the centering/repositioning of the menu categories. <3

vidvidvid (commented):
I really like the animation of the dashboard when it loads / resizes as well as opacity changes of the widgets!
The increase in paddings on top is not bad, but it got me thinking - what if we moved the buttons to edit the layout to the side of the dashboard like this:?
Screenshot 2022-03-01 at 19 04 25
We could keep them at the top for smaller screen sizes.
We could also think about adding some kind of a header to the dashboard - what do you think about that?

dan13ram:
> what if we moved the buttons to edit the layout to the side of the dashboard like this:? 
> We could keep them at the top for smaller screen sizes. We could also think about adding some kind of a header to the dashboard - what do you think about that?

The only thing about moving the buttons on the dashboard is that currently it looks exactly the same as the profile page. And it should give the user the same UX. But if we switch it up then imo we should switch it up on the profile page as well. 

What would the header contain? There's already the MegaMenu header, and i wouldn't wanna cramp it up there. But I'm open to your ideas here.

vidvidvid:
Yep, I see your point. let's keep the UI's consistent.

vidvidvid (approved):

Issue #1180, opened 2022-03-02
Move vercel deployments over to google cloud on test and prod


alalonde:
### What would you like to be added?

We already have a Dockerfile for building frontend containers, as well as a working staging instance deployed in Google Cloud. We just have to spin up another instance linked to the `master` branch and point the DNS.

### Why is this needed?

Vercel isn't providing much value for us atm, and is actually blocking some functionality due to not supporting node.js v16 in their build containers.

dysbulic:
From [Discord](//discord.com/channels/629411177947987986/950436965461950555/950546228238774332):

> I was thinking one option is to do it with `if` statements on the existing Cloud Run deployment.
> 
> If `branch == master` skip the backend and Hasura builds & change the endpoint environment variables.
> 
> Also, set the `--min-instances` option to `1` when master.
> 
> There might not be enough overlap to make it worth it, but it'd be nice to not have to port improvements to another Action. *(I'm still planning to get Docker BuildKit or Cloud Native BuildPacks working to reduce build times…)*

PR #1184, opened 2022-03-06
feat: upgraded dependencies

dan13ram:
### Upgrades all dependencies along with cleaning up some old/unnecessary dependencies

- The following dependencies are NOT upgraded
  1. `discord.js` - leaving it as  #1070 addresses it
  2. `@typeit/discord` - same as above
  3. `next` - tried to do it, but it is much more complicated than anticipated due to some issues with importing ESM modules
  4. `react-timezone-select` - same as above. The upgrades of `next` and `react-timezone-select` need to be done together and well tested in a future PR.
  5. `sourcecred` - there are breaking changes, so left it for a future PR.
  6. `hasura-cli` - should ideally upgrade server and client together, so leaving this as is for now.

- Except for the above packages, all other dependencies have been upgraded along with Typescript v4.6.2
- There is no logic change anywhere.
- Please clone and test locally in different environments to ensure this works for all builders.
- Tested with `yarn v1.19.2` and `node v16.8.0`


dan13ram:
Looks like this is also blocked by #1180 due to vercel node version issue.

dysbulic:
@dan13ram, I pushed a couple changes to the deployment Action & the codegen config to try & get the Cloud Run deployment to complete.

dysbulic:
Unfortunately, it is now floundering on the code that's missing because [the DAUHaus “Supergraph” is currently offline](//api.thegraph.com/subgraphs/name/odyssy-automaton/daohaus/graphql).

I noticed another graph that seemed to be pulling in the same info, but I'll not fix it on this branch because that's getting a bit off topic from the dependency upgrades.

dysbulic:
> @dysbulic what if we set it in `resolutions` of the root package.json ? would that help ?

Unfortunately not. It will install `v8.3.2` in the root `node_modules/` then, but libraries which were built against `v3.4` or less will fail because "v4 is not listed in the exports in package.json". In short, it errors out.

vidvidvid:
Hello, I'm testing various things that were things in the app:
1. I'm not sure if that's an error which was introduced in this PR, but I recently renamed my account from `vid` -> `was-vid`, however https://frontend-pr-1184-mjhnbmqqna-uk.a.run.app/player/vid leads to the old profile and https://frontend-pr-1184-mjhnbmqqna-uk.a.run.app/player/was-vid throws an error.

2. I'm getting an error in the NFT section:
Screenshot 2022-03-08 at 12 55 44
tested on following accounts: [acc1](https://frontend-pr-1184-mjhnbmqqna-uk.a.run.app/player/dysbulic), [acc2](https://frontend-pr-1184-mjhnbmqqna-uk.a.run.app/player/vid)

dan13ram:
@vidvidvid I'm not entirely sure if the frontend on these preview deployments are pointing to the prod ceramic. That might be what's causing this issue for you, where the username is not updated.

What do you think @dysbulic  ?

PR #1192, opened 2022-03-13
Fix two major UI issues

alalonde:
## Overview

Fixes two major issues:
- the edit button appeared on the guild page if you were not logged in. You can't actually view any non-public guild info (or update anything), but it's awkward to say the least.
- the landing page had a major visual bug on Firefox.

- The guild edit icon doesn't actually render if you are the guild owner.  After a couple hours of debugging I tracked it down to the graphql client it is using. Its requests are not even authenticated for some reason.

## Assets

Screen Shot 2022-03-08 at 2 23 07 PM


vidvidvid (approved):
looks 👍

Issue #1196, opened 2022-03-15
[MyMeta] Profile Creation Flow Fixes & Updates


peth-yursick:
I was showcasing MyMeta to the Seed Fund people yesterday & noticed some things off...
This issue includes things that should obviously be fixed but also some changes that are my personal suggestions which can be disregarded.

As usual, actually writing these things down turned out a lot more small issues than expected.
Should this issue be split into multiple?

Either way, I'm marking with "🚨" those that I think are crucial & urgent for the MyMeta product launch that was supposed to be this week but imo should be postponed at least a few days until some of these are solved.
But obviously, the more we can finish, the better.

- [x] 🚨 The progress bar is using the old logo, should be updated to [the new](https://discord.com/channels/629411177947987986/760181556765130752/952971204317294632)
- [x] 🚨 The progress bar indicator is already at 30% on step 1
- [x] I'd change "connect to progress" to "connect to continue" which sounds smoother imo.
- [x] I think it should say "Name" or at least "Nickname". Imo "Username" makes no sense as its not used to log in & also sounds kind of lame & we're trying to move away from this paradigm of people being just "users". For the sentence below, I'd write something more friendly-sounding like "Hey! What's your name? 🙃". And the field name should be changed accordingly.
- [x] When using uppercase, which isn't allowed, [the error isn't always clear](https://user-images.githubusercontent.com/55857222/158350508-e0284680-b334-49c0-9699-fed53198dd28.png)
- [x] Also, could we allow people use uppercase letters in their name?
- [x] 🚨🚨🚨 There is no "bio" step at all, for people to describe themselves.
- [x] Color Disposition
    - [x] The color disposition screen... Is called "color disposition" but the boxes aren't about colors. There's an issue about #1135 revisiting the quiz & whether it should remain about these descriptive words or actually go to colors, but until its resolved maybe it would be best to name the screen just "Personality Type".
    - [x] 🚨 On the colors disposition screen, the boxes [aren't all the same size](https://user-images.githubusercontent.com/55857222/158352504-4480b7e2-a8d4-4e27-ac42-84d2946bbce0.png)..
    - [x] 🎨 Maybe its just me, but I also think the drop shadow on icons takes away from the overall design & makes icons appear to be floating. Thoughts @davort?
[...]

PR #1197, opened 2022-03-16
Fix/UI content improvements

vidvidvid:
## Overview

Fixed a bunch of issues from #1196 

+ aligned the edit profile form:
Screenshot 2022-03-16 at 17 22 40

paired-with: @luxumbra


luxumbra:
Also noticed this in the MegaMenu today - Both MetaGame Wiki and Welcome to MetaGame have the same description text. 
image

vidvidvid:
@luxumbra yup, when i was making all the iframe embeds i fcked this up, will fix it!

packages/web/components/Setup/SetupName.tsx
```diff
@@ -5,23 +5,23 @@ import React from 'react';
...
-  const field = 'username';
+export const SetupName: React.FC = () => {
+  const field = 'name';
 
   return (
     <ProfileWizardPane
       {...{ field }}
-      title="Username"
-      prompt="What name would you like to use in your MyMeta profile URL?"
+      title="Name"
```
dan13ram, line 14: I've a feeling just "name" might be somewhat misleading since we do not allow spaces or even uppercase letters to my knowledge. Something like "User Handle" or "MetaGame Handle" ?

packages/web/next.config.js
```diff
@@ -25,6 +25,11 @@ module.exports = withTM(
           destination: '/profile/setup/username',
           permanent: false,
         },
+        {
+          source: '/profile/setup/username',
```
dan13ram, line 29: may wanna replace all occurences in the codebase of this url with the other one, and also change the redirect just above this to the new one.

packages/web/pages/profile/setup/description.tsx
```diff
@@ -12,12 +12,12 @@ export const getStaticProps = async () => ({
 
 export type DefaultSetupProps = InferGetStaticPropsType<typeof getStaticProps>;
 
-const PronounsSetup: React.FC<DefaultSetupProps> = () => (
+const UsernameSetup: React.FC<DefaultSetupProps> = () => (
```
dan13ram, line 15:
```suggestion
const DescriptionSetup: React.FC<DefaultSetupProps> = () => (
```

packages/web/pages/profile/setup/description.tsx
```diff
@@ -12,12 +12,12 @@ export const getStaticProps = async () => ({
...
   <SetupContextProvider>
     <SetupProfile>
-      <SetupPronouns />
+      <SetupDescription />
     </SetupProfile>
   </SetupContextProvider>
 );
 
-export default PronounsSetup;
+export default UsernameSetup;
```
dan13ram, line 23:
```suggestion
export default DescriptionSetup;
```

packages/web/pages/index.tsx
```diff
@@ -94,6 +96,12 @@ const Landing: React.FC = () => {
 
   return (
     <>
+      <HeadComponent
+        title="MetaGame"
+        description="To find your metagame means to play life in the optimal way. Coordinating with others on building a better world; doing things that create a a positive impact make you happy AND earn you money."
+        url="https://metagame.wtf"
+        img="https://metagame.wtf/_next/image?url=%2Fassets%2Fmanifesto.jpg&amp;w=1080&amp;q=75"
```
dan13ram, line 103: How sure are we that this will work ? 
IMO it would be much cleaner to move the manifesto.jpg to the public folder and make it simply 
`https://metagame.wtf/manifesto.jpg` ?

dan13ram (approved):
LGTM, requested some minor changes but shouldn't block a merge

packages/web/pages/index.tsx
luxumbra: > How sure are we that this will work ? IMO it would be much cleaner to move the manifesto.jpg to the public folder and make it simply `https://metagame.wtf/manifesto.jpg` ?

I assumed it would as there are other instances of that component using the same format but happy to change it if you think would be better. Is there a way to do it without hard-coding the hostname?

packages/web/next.config.js
vidvidvid: good catch, thank you. also i'll leave
```

        {
          source: '/profile/setup/username',
          destination: '/profile/setup/name',
          permanent: false,
        },
```
in case someone accesses an old link. afaik i have already changed all the other occurances of `/setup/username`

luxumbra:
Is this ready for merge? @vidvidvid

vidvidvid:
@luxumbra made a couple of additional improvements, but yes, i think this should be ready very soon!

PR #1220, opened 2022-03-29
Move players to /players

### LINKED ISSUE #1219: Move the leaderboard back to `/players`

A while back @vidvidvid put the leaderboard at `/community/players` to align with the categories of the megamenu.

There is no simple way (that I know of) to go straight from [metagame.wtf](//metagame.wtf) to the leaderboard, so I end up typing it in.

In general, you add to a path to disambiguate, but there's no other `players` collections on our site, so it isn't necessary.

I vote we move it back to just `/players` & also add some jumping off points to the initial homepage, so return visitors can more easily get to where they want to be.

### Why is this needed?

Unnecessary typing and more to remember.

m9p909:
## Overview

Moves the players page to /players

Closes #1219 

## Implementation
move players file up a directory, reverse the redirect so community/players redirects to /players

## Assets
Screenshot from 2022-03-29 10-21-09


dysbulic:
If you do a `git grep community/players`, you'll see the path is present a couple of other places in the codebase.

m9p909:
> If you do a `git grep community/players`, you'll see the path is present a couple of other places in the codebase.

Should be good now. Thanks for the tip!

dysbulic (commented):
Looks good. @peth-yursick had two URL changes that need to be made.

Could you throw those in this as well? They're in [Discord](https://discord.gg/KSU8FneP6K).

dysbulic (approved):
Parfait.

PR #1266, opened 2022-05-10
Fix XP sync for players with multiple linked eth addresses in SC

### LINKED ISSUE #1243: vid's xp is stuck
### What happened?

Both total and seasonal xp have been stuck for the past 3 weeks:
Screenshot 2022-04-13 at 16 30 24

xp should keep on changing as it does here: https://xp.metagame.wtf/

Go to https://metagame.wtf/players or https://metagame.wtf/player/vid 

My xp is also stuck in my header:
Screenshot 2022-04-13 at 16 32 11
 

_No response_

alalonde:
## Overview

This fixes an issue where players that have multiple eth addresses in the SC Ledger aren't getting their XP updated properly during the sync job.

Closes #1243 

Paired with @dysbulic 

I tested it locally by manually setting vid's XP to lower numbers and then running the sync job. Seems to work.

### DIFF
--- a/packages/utils/src/sourceCredHelpers.ts
+++ b/packages/utils/src/sourceCredHelpers.ts
@@ -1,19 +1,17 @@
-import { ethers } from 'ethers';
+import { utils } from 'ethers';
 import { SCIdentity, sourcecred as sc } from 'sourcecred';
 
 export const getLatestEthAddress = (identity: SCIdentity): string | null => {
   const ethAddresses = identity.aliases.filter((alias) => {
     const parts = sc.core.graph.NodeAddress.toParts(alias.address);
-    return (
-      parts.indexOf('ethereum') > 0 && ethers.utils.isAddress(alias.description)
-    );
+    return parts.indexOf('ethereum') > 0 && utils.isAddress(alias.description);
   });
   if (ethAddresses.length > 0) {
     // this assumes that the fetched data is in chronological order, which should
     // be the case for the SC Ledger
     const latestAddress = ethAddresses.pop();
     if (latestAddress != null) {
-      return latestAddress?.description.toLowerCase();
+      return latestAddress.description.toLowerCase();
     }
   }
 
--- a/packages/utils/src/sourceCredHelpers.ts
+++ b/packages/utils/src/sourceCredHelpers.ts
@@ -2,13 +2,19 @@ import { ethers } from 'ethers';
 import { SCIdentity, sourcecred as sc } from 'sourcecred';
 
 export const getLatestEthAddress = (identity: SCIdentity): string | null => {
-  const ethAddress = identity.aliases.find((alias) => {
+  const ethAddresses = identity.aliases.filter((alias) => {
     const parts = sc.core.graph.NodeAddress.toParts(alias.address);
-    return parts.indexOf('ethereum') > 0;
-  })?.description;
-
-  if (ethAddress && ethers.utils.isAddress(ethAddress)) {
-    return ethAddress.toLowerCase();
+    return (
+      parts.indexOf('ethereum') > 0 && ethers.utils.isAddress(alias.description)
+    );
+  });
+  if (ethAddresses.length > 0) {
+    // this assumes that the fetched data is in chronological order, which should
+    // be the case for the SC Ledger
+    const latestAddress = ethAddresses.pop();
+    if (latestAddress != null) {
+      return latestAddress?.description.toLowerCase();
+    }
   }
 
   return null;
--- a/packages/web/lib/hooks/useUserXP.ts
+++ b/packages/web/lib/hooks/useUserXP.ts
@@ -1,5 +1,5 @@
 import { Constants } from '@metafam/utils';
-import { ethers } from 'ethers';
+import { utils } from 'ethers';
 import { useEffect, useState } from 'react';
 import { SCAccount, SCAccountsData } from 'sourcecred';
 
@@ -15,7 +15,7 @@ interface XPProps {
 export const useUserXP = (userAddress: string): XPProps | null => {
   const [xpDataObj, setXPDataObj] = useState<XPProps | null>(null);
   useEffect(() => {
-    if (ethers.utils.isAddress(userAddress))
+    if (utils.isAddress(userAddress))
       getXP(userAddress).then((res) => {
         if (res) setXPDataObj(res);
       });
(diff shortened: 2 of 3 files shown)


alalonde:
Note that discovery of this bug was a side effect of my manually updating vid's eth address in the test/prod DBs...

packages/utils/src/sourceCredHelpers.ts
```diff
@@ -2,13 +2,19 @@ import { ethers } from 'ethers';
...
+      parts.indexOf('ethereum') > 0 && ethers.utils.isAddress(alias.description)
+    );
+  });
+  if (ethAddresses.length > 0) {
+    // this assumes that the fetched data is in chronological order, which should
+    // be the case for the SC Ledger
+    const latestAddress = ethAddresses.pop();
+    if (latestAddress != null) {
+      return latestAddress?.description.toLowerCase();
+    }
```
dysbulic, line 17: Since you're using the nullish coalescing operator, you can let `latestAddress` be `undefined`.

If, for whatever reason you wanted to always return `null` on a failure, you could do `return latestAddress?.description.toLowerCase() ?? null`.

Since the filter only passes aliases with valid ETH addresses as descriptions, if we make it past the empty array check, I don't see how the `pop` could be `undefined` anyhow.

```suggestion
    return latestAddress?.description.toLowerCase();
```

dysbulic (approved):
Looks good to me. I'd remove one of your `if`s since it's not needed on two fronts.

dysbulic:
Also, I'm expecting this to break on my MyMeta account. I've currently got two accounts: one that was compromised long ago, but that has continued to show as active because the `getLatestEthAddress` function was returning the first registered address, and one for my actual active account.

Since we don't zero everyone's XP before updating from SourceCred, I'm expecting the new address will have a bunch of XP and the old one will maintain the XP it already has putting me on the list twice.

We should likely add a zeroing step to the SourceCred import or zero an account when it's ETH address is force set.

dysbulic:
Well, I approved too soon. I have no idea why, but visiting an address not in the system, like [0xd8acbe34b498c8717ccc95559536ed2b7fa77fbc](https://frontend-pr-1266-mjhnbmqqna-uk.a.run.app/player/0xd8acbe34b498c8717ccc95559536ed2b7fa77fbc) causes the server to keel over with a 500 error:

```
TypeError: ethers__WEBPACK_IMPORTED_MODULE_5__.ethers.utils.isAddress is not a function
at formatIfAddress (/app/packages/web/.next/server/chunks/8039.js:909:93)
at getPlayerName (/app/packages/web/.next/server/chunks/8039.js:891:172)
at PlayerPage (/app/packages/web/.next/server/chunks/4538.js:3661:74)
at d (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:33:498)
at bb (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:36:16)
at a.b.render (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:42:43)
at a.b.read (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:41:83)
at exports.renderToString (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:52:138)
at Object.renderPage (/app/node_modules/next/dist/next-server/server/render.js:54:854)
at Function.getInitialProps (/app/packages/web/.next/server/chunks/331.js:511:19)
at Function.getInitialProps (/app/packages/web/.next/server/pages/_document.js:28:83)
at loadGetInitialProps (/app/node_modules/next/dist/next-server/lib/utils.js:5:101)
at renderToHTML (/app/node_modules/next/dist/next-server/server/render.js:54:1145)
at async /app/node_modules/next/dist/next-server/server/next-server.js:112:97
at async __wrapper (/app/node_modules/next/dist/lib/coalesced-function.js:1:330)
at async Server.renderToHTMLWithComponents (/app/node_modules/next/dist/next-server/server/next-server.js:137:387)
```

dysbulic (changes requested):
I approved too soon. Connecting with an account not in the system causes an error that "`ethers.utils.isAddress is not a function`". Why is beyond me.

packages/utils/src/sourceCredHelpers.ts
dysbulic: All you'd have to do is hit the "Commit Suggestion" button. The `if` is unarguably superfluous.

As always, it's just a suggestion.

alalonde:
> Well, I approved too soon. I have no idea why, but visiting an address not in the system, like [0xd8acbe34b498c8717ccc95559536ed2b7fa77fbc](https://frontend-pr-1266-mjhnbmqqna-uk.a.run.app/player/0xd8acbe34b498c8717ccc95559536ed2b7fa77fbc), causes the server to keel over with a 500 error:
> 
> ```
> TypeError: ethers__WEBPACK_IMPORTED_MODULE_5__.ethers.utils.isAddress is not a function
> at formatIfAddress (/app/packages/web/.next/server/chunks/8039.js:909:93)
> at getPlayerName (/app/packages/web/.next/server/chunks/8039.js:891:172)
> at PlayerPage (/app/packages/web/.next/server/chunks/4538.js:3661:74)
> at d (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:33:498)
> at bb (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:36:16)
> at a.b.render (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:42:43)
> at a.b.read (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:41:83)
> at exports.renderToString (/app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:52:138)
> at Object.renderPage (/app/node_modules/next/dist/next-server/server/render.js:54:854)
> at Function.getInitialProps (/app/packages/web/.next/server/chunks/331.js:511:19)
> at Function.getInitialProps (/app/packages/web/.next/server/pages/_document.js:28:83)
> at loadGetInitialProps (/app/node_modules/next/dist/next-server/lib/utils.js:5:101)
> at renderToHTML (/app/node_modules/next/dist/next-server/server/render.js:54:1145)
> at async /app/node_modules/next/dist/next-server/server/next-server.js:112:97
> at async __wrapper (/app/node_modules/next/dist/lib/coalesced-function.js:1:330)
> at async Server.renderToHTMLWithComponents (/app/node_modules/next/dist/next-server/server/next-server.js:137:387)
> ```
> 
> Why it only occurs with a new address is beyond me at the moment, but that seems to be the case.

Yeah no idea why this is happening. I don't see it on staging. I pushed an update to see if it will fix it.

dysbulic:
> Since we don't allow updating your eth address in mymeta (which we might consider) I don't see this problem surfacing again without manual DB intervention.

It would happen when someone force updates their ETH address. That will leave their previous address with whatever XP it had and also give XP to the new address.

dysbulic:
You could always forget the checksumming & do `/^0x[\da-f]{40}/g.test()`. Or, put a `isAddresslike` in `utils` with that as the definition.

Honestly, I've never had the checksum catch an error because absolutely no one is typing these things in.

dysbulic (approved):

PR #1275, opened 2022-05-24
Changing location of `isAddress` function

### LINKED ISSUE #1274: Getting 500 errors when calling `ethers.utils.isAddress()`
### What happened?

When connecting to [metagame.wtf](//metagame.wtf) with an account that hasn't been connected before, also in the `/api/opensea` endpoint, calls to `ethers.utils.isAddress()` are failing.

In the Cloud Run logs, the error is `TypeError: ethers__WEBPACK_IMPORTED_MODULE_5__.utils.isAddress is not a function`.

The method to execute correctly.

Visit the [OpenSea API endpoint](https://metagame.wtf/api/opensea?owner=0x70a92ca61f410453b6c4c1ad0338c2a782422b9d&offset=0&limit=50) and it will fail.

_No response_

_No response_

dysbulic:
## Overview

Version 4 of the `ethers` library, which is where the implementation of `isAddress` is being imported from, doesn't include `isAddress`. This is causing `/api/opensea` to fail as well as `/me` for new addresses.

This PR changes the import location of `isAddress` to `@ethersproject/address` which pulls from version 5.6 of the library where the function is defined.

Closes #1274.

Alternatively, this could have been accomplished by setting the `resolutions` in `package.json`, but that might have had unknown side-effects.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -15,14 +15,15 @@
   },
   "dependencies": {
     "@datamodels/identity-profile-basic": "0.1.2",
+    "@ethersproject/address": "^5.6.0",
     "bignumber.js": "9.0.2",
     "cids": "1.1.9",
     "ethers": "5.5.4",
     "imgix-core-js": "2.3.2",
+    "js-base64": "3.7.2",
     "key-did-provider-ed25519": "1.1.0",
     "key-did-resolver": "1.4.4",
     "node-fetch": "3.2.1",
-    "js-base64": "3.7.2",
     "sourcecred": "0.11.0",
     "uuid": "8.3.2"
   },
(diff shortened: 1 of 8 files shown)


alalonde (approved):
Nice! Can we get rid of `ethers` then? Is @ethersproject the same thing but split up for better modularity?

dysbulic:
> Can we get rid of `ethers` then? Is @ethersproject the same thing but split up for better modularity?

`ethers` includes all the `@ethersproject` subpackages from a single source. Traditionally, you'd use `ethers` for imports, but the version that's getting hoisted by `yarn` is `v4` rather than the needed `v5`. Using the `@ethersproject` subpackage is just a way to coerce the usage of the `v5` package.

PR #1285, opened 2022-06-04
Fix the XP command on the discord bot

### LINKED ISSUE #1284: XP discord command is broken
### What happened?

image

It should tell me my XP

`!mg xp` in the proper channel

alalonde:
## Overview

Fixes #1284 

## Implementation

I fixed parsing of `discordId` from a sourcecred address. I suspect this broken during our last sourcecred upgrade. These homemade utilities to parse SC internals are brittle; best would be to contribute them back to SC to be used by everybody.


dysbulic (approved):
Looks reasonable to me, but if it is the last element in the array, I'd write it as:

```javascript
const [discordId] = sc.core.graph.NodeAddress.toParts(discordAccount.address).slice(-1)
```

That would make it less likely to fail due to changes in the path.

PR #1290, opened 2022-06-08
switched pSEED to Polygon

### LINKED ISSUE #1021: Move pSEED balances from Mainnet to Polygon + Cache

Update the subgraph to use polygon network. 

### Why is this needed?

We have migrated from mainnet balancer pool to polygon, so we need this.

tenfinney:
## Overview

Changes TheGraph used to find pSEED balances from @dan13ram's to a new Graph. Updates the check in Hasura for quest creation to use Polygon.

Closes #1021 

@paired-with: @dysbulic

### DIFF
--- a/packages/backend/src/handlers/actions/quests/createQuest/permissions.ts
+++ b/packages/backend/src/handlers/actions/quests/createQuest/permissions.ts
@@ -1,7 +1,7 @@
 import { numbers } from '@metafam/utils';
 
 import { CONFIG } from '../../../../config';
-import { getERC20Contract } from '../../../../lib/ethereum';
+import { getERC20Contract, polygonProvider } from '../../../../lib/ethereum';
 
 const { BN, amountToDecimal } = numbers;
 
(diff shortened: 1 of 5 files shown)


packages/backend/src/handlers/actions/quests/createQuest/permissions.ts
```diff
@@ -12,8 +12,9 @@ const { BN, amountToDecimal } = numbers;
 export async function isAllowedToCreateQuest(
   playerAddress: string,
 ): Promise<boolean> {
+  console.log(CONFIG.pSEEDAddress);
```
alalonde, line 15: Left a bit of cruft.  FYI, at the bottom of our README I recently added instructions for debugging backend code using VS code. It has sped up my workflow quite a bit

packages/backend/src/lib/ethereum.ts
```diff
@@ -5,9 +5,13 @@ import ERC20_ABI from './abis/ERC20.json';
 
 const { infuraId } = CONFIG;
 
-export const defaultProvider = new ethers.providers.InfuraProvider(1, infuraId);
+export const mainnetProvider = new ethers.providers.InfuraProvider(1, infuraId);
+export const polygonProvider = new ethers.providers.InfuraProvider(
+  137,
```
alalonde, line 10: Is there a constant we can use rather than a hardcoded int ?

alalonde (commented):
Did you test the frontend for quests also? I seem to remember there being separate code to validate pSeed balance on the frontend vs backend. You can just hover over the "create quest" button to test

packages/backend/src/lib/ethereum.ts
dysbulic: > Did you test the frontend for quests also?

Yeah, that is why we needed to create the Polygon provider. @tenfinney & I both successfully created quests.

dysbulic: > Is there a constant we can use rather than a hardcoded int ?

We could make some… Something like:
```javascript
const chain = {
  polygon: 0x89,
  mainnet: 0x1,
}
```

alalonde (approved):

PR #1295, opened 2022-06-21
Fix graphQL error in quest filter "Created by Me"

alalonde:
Also improved the look and feel.


.github/workflows/gcp-deploy.yaml
```diff
@@ -496,6 +496,7 @@ jobs:
             --set-env-vars OPENSEA_API_KEY=${{secrets.OPENSEA_API_KEY}} \
             --set-env-vars NEXT_PUBLIC_IMGIX_TOKEN=${{secrets.IMGIX_TOKEN}} \
             --set-env-vars NEXT_PUBLIC_GRAPHQL_URL=${{needs.env.outputs.GRAPHQL_URL}}
```
dysbulic, line 498: The newline needs to be escaped.

```suggestion
            --set-env-vars NEXT_PUBLIC_GRAPHQL_URL=${{needs.env.outputs.GRAPHQL_URL}} \
```

dysbulic (changes requested):
Everything looks reasonable, but the test deployment didn't work because of an unescaped newline. I'd like to try it out in test before approving…

alalonde:
Once #1299 goes through I'll rebase this and it should be gtg as well

PR #1301, opened 2022-07-06
Dependency Upgrades

alalonde:


dysbulic (changes requested):
It looks like the backend won't start…

```
The user-provided container failed to start and listen on the port defined provided by the PORT=4000 environment variable. Logs for this revision might contain more information.
```

alalonde:
I looked into importing these upgraded Ceramic ESM modules, and it seems that the only path forward is to convert our own backend code (and potentially the entire monorepo) into ESM as well. I didn't want to bite that off as part of this PR, so I just downgraded those few packages.

I found [this](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) which should help if and when we need to upgrade our own repo(s).

dysbulic (approved):

PR #1318, opened 2022-07-19
Cache Dao Memberships

### LINKED ISSUE #997: Cache correctly-sourced DAOHaus membership information

Currently, DAOHaus memberships are associated with the player record by way of a remote schema.

This means that every time a player record is loaded, [The Graph](//thegraph.com) is queried three times for the Polygon, xDAI, and mainnet DAO memberships.

This is not only time consuming, as of a release several months ago, DAOHaus is no longer putting the name of the DAO on chain. It is stored in an extensive metadata file available [elsewhere](//github.com/MetaFam/TheGame/issues/646).

A stale-while-revalidating cache to be triggered with the same logic as the Ceramic profile cache (dirty a user when their profile details page is loaded, or every four days at the longest) would likely work.

### Why is this needed?

Page loads are currently abysmally slow.

alalonde:
## Overview

Improves the performance on the /players page, as documented in #997.

Closes #997 

- [ ] We should probably update the guild form to allow guilds to select one of the daohaus DAOs, since we will be pulling in a ton of them. Or, at the least, associate an existing DAO if the inputted contract address matches a known one in our DB
- [ ] During player setup (and in the edit profile screen), allow a user to specify the DAOs they want to show up on their profile. This should be relatively easy now that we have a `dao_player` table and `visible` column.


alalonde:
Implemented as planned [https://github.com/MetaFam/TheGame/issues/997#issuecomment-1188594379](here).  Just pushed a fix for the tests as well

dysbulic (approved):
All this looks really good. I'll poke at the test instance some, but I'm pretty sure it's solid.

PR #1324, opened 2022-08-08
Feat/onboarding game

luxumbra:
## Overview
Work in progress on the onboarding game.

Issue number to follow

- [ ] Add in either new 3d render or a 3d scene instead of the poor quality image

## Implementation

**Side effects**

## Assets


dysbulic:
I don't think you can put the `paired-with` inline. I think it has to be on it's own line at the end.

luxumbra:
> @luxumbra, I don't think you can put the `paired-with` inline. I think it has to be on its own line at the end.

moved to the 'Implementation' section

packages/web/utils/stringHelpers.ts
```diff
@@ -7,3 +11,57 @@ export const hashCode = (str: string): string => {
...
+        }
+      }
+    },
+  };
+
+  const parsed = parse(clean, options);
+  return parsed;
+};
+
+export const shortenAddress = (address: string, sliceSize?: number): string => {
```
dan13ram, line 59: there's a formatAddress in `web3/utils/playerHelpers.ts` that does something similar to this and perhaps we could reuse that here, or modify that method to support this extra paramer "sliceSize".

packages/web/components/Landing/LandingConnectButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+        toast({
+          title: 'Wallet Connection',
+          description: `Successfully connected to ${networkName} with ${shortAddress}. 🥳`,
+          status: 'success',
+          isClosable: true,
+          duration: 5000,
+        });
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dan13ram, line 77: I see a few places where you're using `eslint-disable-next-line react-hooks/exhaustive-deps`. In my experience you're doing something wrong if you need this rule disabled. 

In any case you should be able to add whatever state variables inside the hook into the dependency array and this warning should go away :)

packages/web/components/Landing/LandingConnectButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+  } = useWeb3();
+  const [wrongNetwork, setWrongNetwork] = useState(false);
+  const spinnerSize = useBreakpointValue({ base: 'sm', '2xl': 'md' });
+  const toast = useToast();
+
+  useEffect(() => {
+    if (connected && address && provider) {
+      const networkName = provider.network.name;
+      const shortAddress = shortenAddress(address);
+      console.log('Connected to', networkName, 'with address', shortAddress);
```
dan13ram, line 37: Also we may want to remove all these console logs. If you want them during testing we can wrap them all into a helper function which logs based on an env variable.

packages/web/components/Landing/LandingConnectButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+                <Box as="strong" textTransform="capitalize">
+                  {networkName ?? ''} ({chainId})
+                </Box>{' '}
+                network is not supported. This area of the website requires{' '}
+                <strong>Polygon (id: 0x89)</strong> network. Please switch
+                networks.
+              </Text>
+              <Text>
+                You can visit{' '}
+                <MetaLink href="https://chainlist.org/chain/137" isExternal>
```
dan13ram, line 55: There is a `web3/components/SwitchNetworkButton.tsx` that you can use to help users automatically add this network to their metamask.

We could modify that component to have a link to `chainlist.org` as you've done here.

packages/web/components/Landing/LandingConnectButton.tsx
```diff
@@ -0,0 +1,176 @@
+/* eslint-disable no-nested-ternary */
```
dan13ram, line 1: You're disabling this in multiple places. We might as well disable the rule in our eslint config 🗡️

packages/web/components/Landing/JoinUs.tsx
```diff
@@ -1,25 +1,57 @@
...
 export const JoinUs: React.FC = () => {
   const ref = useRef<HTMLDivElement>(null);
   const onScreen = useOnScreen(ref);
+  const root = typeof window !== 'undefined' ? document.body : null;
+  const [noMotion, setNoMotion] = useState(true);
+  const displayElement = noMotion ? true : !!onScreen;
+  const buttonSize = useBreakpointValue({ base: 'sm', xl: 'lg' });
   const section = 'join-us';
 
+  useEffect(() => {
```
dan13ram, line 31: this useEffect and its content is being used in a lot of places. Worth wrapping it into a separate hook that can be called in all these places.

We could also store `no-motion` in localStorage. Its my assumption that if someone is opting for "no motion" then it should be persisted the next time they visit our site.

packages/web/components/Landing/LandingHeader.tsx
```diff
@@ -1,23 +1,114 @@
...
+  }, [effectsToggle, noMotion]);
+
+  /** Initially set the motion pref if it's not already set */
+  useEffect(() => {
+    if (typeof window !== 'undefined' && window.matchMedia !== undefined) {
+      if (savedMotionPreference === null) {
+        set('MotionPreference', prefersReducedMotion ? 'off' : 'on');
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dan13ram, line 78: lets remove all occurences of `eslint-disable-next-line react-hooks/exhaustive-deps` and add the necessary dependencies into it.

packages/web/components/Landing/LandingHeader.tsx
```diff
@@ -1,23 +1,114 @@
...
 export const LandingHeader: React.FC = () => {
   const [toggle, setToggle] = useState(false);
+  const prefersReducedMotion = usePrefersReducedMotion();
+  const savedMotionPreference = get('MotionPreference');
+  const [noMotion, setNoMotion] = useState(
+    savedMotionPreference === 'off'
+      ? true
+      : savedMotionPreference === 'on'
+      ? false
+      : prefersReducedMotion,
```
dan13ram, line 39: Elsewhere I had commented that we store the `no-motion` in localStorage but looks like you're already doing it! 🥳

Lets `setNoMotion` here in a useEffect. This `useState` will be run on the server during static build and won't be set correctly. (This may explain the TODO you've added below)

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+        console.log('receipt', tx);
+      }
+      setClaiming(false);
+    } catch (error: any) {
+      console.log('handleMintingError', error);
+    }
+  }, [chievId, mintChiev]);
+
+  /** function to get the token image url from the nft metadata */
+  const getTokenImage =
```
dan13ram, line 104: We can move this function "outside" the component else this function will be recreated on every render of this component. 

I also noticed that all the information for this token (`alchemyApiKey`, `contractAddress`, `ABI` and the `cheivId`) are constant. In this case we can probably get the metadata when the server is statitically generated this page. 

What say @dysbulic ?

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+  const prefersReducedMotion = usePrefersReducedMotion();
+  const [noMotion, setNoMotion] = useState(false);
+  const root = typeof window !== 'undefined' ? document.body : null;
+  const { mintChiev, txLoading } = useGame();
+  const [claiming, setClaiming] = useState(false);
+  const { address: account, chainId, connected } = useWeb3();
+  const [wrongNetwork, setWrongNetwork] = useState(false);
+  const { alchemyApiKey, chievContractAddress: contractAddress } = CONFIG;
+  const claimed = get('ChievClaimed') === 'true' ?? false;
+  const chievId = BigInt(
```
dan13ram, line 55: Lets capitalize constants as a naming convention. 
And lets move them outside the component.

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+      }
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    if (chievData) {
+      console.log('chievData', chievData);
+    }
+  }, [chievData]);
```
dan13ram, line 144: unnecessary useEffect hook. can be removed.

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+              },
+            },
+          }}
+        >
+          <Text
+            as="h3"
+            fontSize={{ base: 'md', '2xl': 'xl' }}
+            fontWeight="black"
+            textTransform="uppercase"
+            color="var(--chakra-colors-landing600)"
```
dan13ram, line 246:
```suggestion
            color="landing.600"
```

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+                <Button
+                  onClick={handleMinting}
+                  isDisabled={txLoading || claiming || claimed || wrongNetwork}
+                  variant="white"
+                  color={
+                    account
+                      ? wrongNetwork
+                        ? 'brightIdOrange.600'
+                        : 'green'
+                      : 'var(--chakra-colors-landing550)'
```
dan13ram, line 291:
```suggestion
                      : 'landing.550'
```

packages/web/pages/onboarding.tsx
```diff
@@ -0,0 +1,184 @@
...
+  },
+});
+
+const Onboarding: React.FC = () => {
+  const scrollContainer =
+    typeof document !== 'undefined'
+      ? document.getElementById('scroll-container')
+      : null;
+  const [section, setSection] = useState(0);
+  const hostName = useRef('https://metagame.wtf');
```
dan13ram, line 33: Can `hostName` be stored as a constant outside the component ?

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+          status: 'info',
+          isClosable: true,
+          duration: 5000,
+        });
+
+        const receipt = await tx.wait(confirmationsWait);
+        const receiptUrl = `${CONFIG.polygonscanBaseURL}/tx/${tx.hash}`;
+        const nftURL = httpLink(metadata.external_url);
+
+        set('ChievClaimed', 'true');
```
dan13ram, line 257: We may want to store this w.r.t the user's address, since i use multiple wallets/addresses in the same browser. Other's may do so too.

packages/web/config.ts
```diff
@@ -31,4 +31,12 @@ export const CONFIG = {
   googleDataApiKey: process.env.NEXT_PUBLIC_YOUTUBE_API_KEY,
   web3StorageToken: process.env.WEB3_STORAGE_TOKEN || '',
   openseaApiKey: process.env.OPENSEA_API_KEY || '',
+  onboardingGameDataURL:
+    'https://ipfs.fleek.co/ipfs/bafybeig5n6md7kemfjgdbguiepzisrt2y6icxhsj65kyrirs3kqgpv56eu',
+  openseaBaseURL: 'https://opensea.io',
+  polygonscanBaseURL: 'https://polygonscan.com',
```
dan13ram, line 37: `polygonBaseURL` is already there `utils/networks` as "explorer" for polygon network.

packages/web/config.ts
```diff
@@ -31,4 +31,12 @@ export const CONFIG = {
   googleDataApiKey: process.env.NEXT_PUBLIC_YOUTUBE_API_KEY,
   web3StorageToken: process.env.WEB3_STORAGE_TOKEN || '',
   openseaApiKey: process.env.OPENSEA_API_KEY || '',
+  onboardingGameDataURL:
```
dan13ram, line 34: All these constants are better saved in `utils/constants.ts` in uppercase.

packages/web/config.ts
```diff
@@ -31,4 +31,12 @@ export const CONFIG = {
...
+    'https://ipfs.fleek.co/ipfs/bafybeig5n6md7kemfjgdbguiepzisrt2y6icxhsj65kyrirs3kqgpv56eu',
+  openseaBaseURL: 'https://opensea.io',
+  polygonscanBaseURL: 'https://polygonscan.com',
+  chievContractAddress: '0x85fCaAFc0dA050FCE685DcB8965F0C1Aa1Ba466b',
+  alchemyRpcUrl: process.env.NEXT_ALCHEMY_ENDPOINT || '',
+  alchemyApiKey:
+    process.env.NEXT_ALCHEMY_API_KEY || 'eoEduI_LCCxfMPe9MKrpvly-qWtBBY7I',
```
dan13ram, line 41: We are already using an infura id, can we not reuse infura for Cheivs? We don't need to use multiple RPC providers.

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+   * this function
+   */
+  const fetchGameData = useCallback(async () => {
+    try {
+      setIsLoading(true);
+      console.log('Fetchng GameData...');
+
+      // const response = await fetch(gameJson);
+      // const data = (await response.json()) as GameProperties;
+      const data = gameJson as GameProperties;
```
dan13ram, line 81: There is nothing async happening here. We can use `gameJson` AS IS whereever it is needed, and can safely remove this `useCallback`.

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+      console.error(error);
+      setIsLoading(false);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (address) {
+      setAccount(address);
+    }
+  }, [address]);
```
dan13ram, line 98: Why do we have an extra state variable `account` and this useEffect. Can we not use `address` itself ?

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+        return undefined;
+      }
+    },
+    [],
+  );
+
+  /** Increment number of elements a user has visited & store in localStorage
+   * This is used to trigger an achievement after so many elements have been visited
+   */
+  const visitedElements = (increment?: boolean): string => {
```
dan13ram, line 188: Can we wrap this in a useCallback ?

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+    }
+  }, [gameDataState, isLoading]);
+
+  /** Function to get, set or clear game state
+   *
+   * @param currentPlace - the current place of the player
+   * @param reset - if true, clear the game state
+   * @returns the game state or undefined if reset is true or the game state is not set
+   */
+  const gameState = (
```
dan13ram, line 147: Can we wrap this in a useCallback ?

packages/web/components/Landing/OnboardingGame/index.tsx
```diff
@@ -0,0 +1,869 @@
...
+  const [chievFound, setChievFound] = useState(false);
+
+  /** Function to async fetch `CONFIG.onboardingGameDataURL` as json and return the data */
+  const fetchGameData = useCallback(async () => {
+    try {
+      setHasError(false);
+
+      // const response = await fetch(CONFIG.onboardingGameDataURL);
+      // console.log('fetchGameData RES', response.status);
+      const data = gameJson as GameProperties;
```
dan13ram, line 92: Since we're not fetching (same as in `GameContext.tsx`) we don't need this useCallback at all.

packages/web/components/Landing/OnboardingGame/Chiev.tsx
```diff
@@ -0,0 +1,431 @@
...
+                  }
+                  textShadow="var(--chakra-colors-landing500)"
```
dan13ram, line 293:
```suggestion
                  textShadow="landing500"
```

packages/web/utils/stringHelpers.ts
luxumbra: > there's a formatAddress in `web3/utils/playerHelpers.ts` that does something similar to this and perhaps we could reuse that here, or modify that method to support this extra paramer "sliceSize".

I missed that. Modifying the method sounds like the most sensible thing to do. thanks.

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: > I see a few places where you're using `eslint-disable-next-line react-hooks/exhaustive-deps`. In my experience you're doing something wrong if you need this rule disabled.
> 
> In any case you should be able to add whatever state variables inside the hook into the dependency array and this warning should go away :)

this is interesting. i must be missing some key knowledge of correct use of useEffect (probably hooks in general) as oftentimes, when I include all deps, i just get endless re-renders. if I am doing it wrong,I need to address that here and in myself asap. 🙏🏻

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: > Also we may want to remove all these console logs. If you want them during testing we can wrap them all into a helper function which logs based on an env variable.

they're definitely helping during testing so yeah, a helper function is a good idea. 👍🏻

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: OK, I'll look at modifying the `SwitchNetworkButton` component to include the Polygon network and a chainlist link (it's a very useful site for folks to know about).

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: If there's no issue with turning it off in eslint config then yeah, cool but if i should be using a different method to get the same result, I'd like to do it the right way.

packages/web/contexts/GameContext.tsx
luxumbra: > We may want to store this w.r.t the user's address, since i use multiple wallets/addresses in the same browser. Other's may do so too.

Aye, I have quite a few addresses also. I wonder if making the chiev claimable once per wallet is the right thing to do anyway. there are multiple ways to get to the end, so maybe it should be upped to 3 or 4. 

On a side note, would you change anything WRT the design of the chiev art itself?

packages/web/pages/onboarding.tsx
luxumbra: sure. will do.

packages/web/components/Landing/LandingHeader.tsx
luxumbra: > Elsewhere I had commented that we store the `no-motion` in localStorage but looks like you're already doing it! partying_face

😄 I replied to that and now see you've seen that I handled it. 🤣 

> Lets `setNoMotion` here in a useEffect. This `useState` will be run on the server during static build and won't be set correctly. (This may explain the TODO you've added below)

Ah, thanks. I'll address that

packages/web/components/Landing/JoinUs.tsx
luxumbra: > this useEffect and its content is being used in a lot of places. Worth wrapping it into a separate hook that can be called in all these places.
> 
> We could also store `no-motion` in localStorage. Its my assumption that if someone is opting for "no motion" then it should be persisted the next time they visit our site.

I am already getting the users `prefers-reduced-motion` preference on loading the page which sets `MotionPreference` in `localStorage`. The effects toggle also updates the value. I take your point about moving it to a separate hook, so will add that to my list. 👍🏻

packages/web/config.ts
luxumbra: > We are already using an infura id, can we not reuse infura for Cheivs? We don't need to use multiple RPC providers.

Infura doesn't support Polygon does it? Chieves is on Polygon and I was testing using a ThirdWeb contract on Mumbai, so used my Alchemy account.

packages/web/contexts/GameContext.tsx
luxumbra: i think this has happened because when i was building initially with ThirdWeb (chiev.es roles weren't quite ready), they use `account` and it stuck in places. Agreed, it is not good and I'll get it sorted. 👍🏻

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    if (address) {
+      setAccount(address);
+    }
+  }, [address]);
+
+  const game = useMemo((): GameProperties => {
```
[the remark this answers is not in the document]
luxumbra: > What's the difference between `game` and `gameDataState`. My understanding is that both the variables are storing the exact same information. And once stored its not changed.

You're right, there isn't a difference.  Needs cleaning up. See [this comment](https://github.com/MetaFam/TheGame/pull/1324#discussion_r965239373)

packages/web/components/Landing/OnboardingGame/Chiev.tsx
luxumbra: > I gave it a quick look and left a bunch of comments.

Amazing! Thank you for the time taken to check this over @dan13ram. I've responded or emoji'd all comments and will address them over the next couple of days.  

> Overall this is just amazing!!! Fantastic work @luxumbra pray

I appreciate that, thanks. 😊 

> I'll take another look once you fix these comments :)

I appreciate you. 🐙

luxumbra:
> I gave it a quick look and left a bunch of comments.

Amazing! Thank you for the time taken to check this over @dan13ram. I've responded or emoji'd all comments and will address them over the next couple of days.  

> Overall this is just amazing!!! Fantastic work @luxumbra pray

I appreciate that, thanks. 😊 

> I'll take another look once you fix these comments :)

I appreciate you. 🐙

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * @param currentPlace - the current place of the player
+   * @param reset - if true, clear the game state
+   * @returns the game state or undefined if reset is true or the game state is not set
+   */
+  const gameState = (
+    currentPlace?: string,
+    reset?: boolean,
+  ): IGameState['state'] => {
+    if (currentPlace !== undefined) {
```
dysbulic, line 151: It is a string, so this is likely safe.

```suggestion
    if (currentPlace) {
```

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+      }
+    },
+    [],
+  );
+
+  /** Increment number of elements a user has visited & store in localStorage
+   * This is used to trigger an achievement after so many elements have been visited
+   */
+  const visitedElements = (increment?: boolean): string => {
+    const visited: string | null = get('OnboardingGameVisitedElements');
```
dysbulic, line 189: Having this as a string representing a number is a kludge & I'd convert it from a string as soon as you pull it out of storage.

```suggestion
    const visited: number | null = Number(store.get('OnboardingGameVisitedElements'));
```

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+  };
+
+  const resetGame = useCallback((): boolean => {
+    gameState(undefined, true);
+    const success = gameState() === null;
+
+    if (success) {
+      return true;
+    }
+    return false;
```
dysbulic, line 210:
```suggestion
    return success;
```

packages/web/contexts/GameContext.tsx
```diff
@@ -0,0 +1,328 @@
...
+        });
+
+        const receipt = await tx.wait(confirmationsWait);
+        const receiptUrl = `${CONFIG.polygonscanBaseURL}/tx/${tx.hash}`;
+        const nftURL = httpLink(metadata.external_url);
+
+        set('ChievClaimed', 'true');
+
+        toast({
+          title: 'Chiev claimed 🎉',
```
dysbulic, line 260: I'd make the `title`s title case.

```suggestion
          title: '’Chiev Claimed 🎉',
```

packages/web/pages/onboarding.tsx
```diff
@@ -0,0 +1,184 @@
+import { Box, Tooltip, VStack } from '@metafam/ds';
+import { PageContainer } from 'components/Container';
+import { LandingHeader } from 'components/Landing/LandingHeader';
+import { Onboard } from 'components/Landing/Onboard';
+import { MetaLink } from 'components/Link';
+import { HeadComponent } from 'components/Seo';
+// import { gsap } from "gsap";
+// import { ScrollToPlugin } from 'gsap/dist/ScrollToPlugin';
+// import { ScrollTrigger } from 'gsap/dist/ScrollTrigger';
+// import { useRouter } from 'next/router';
```
dysbulic, line 10:
```suggestion
```

packages/web/utils/stringHelpers.ts
```diff
@@ -7,3 +11,57 @@ export const hashCode = (str: string): string => {
...
+  const sliceNumber = sliceSize ?? 4;
+  const start: string = address.toLowerCase().slice(0, sliceNumber);
+  const end: string = address
+    .toLowerCase()
+    .slice(Math.max(0, address.length - sliceNumber));
```
dysbulic, line 64: Why do you lower case it? I'd not.

```suggestion
  const start: string = address.slice(0, sliceNumber);
  const end: string = address
    .slice(Math.max(0, address.length - sliceNumber));
```

packages/web/utils/stringHelpers.ts
```diff
@@ -7,3 +11,57 @@ export const hashCode = (str: string): string => {
...
+  const shortAddress = `${start}...${end}`;
```
dysbulic, line 65:
```suggestion
  const shortAddress = `${start}…${end}`;
```

dysbulic (approved):
My eyes sorta glazed over at times, but I didn't see anything that looked like it'd blow up in the near term. 😹

The most common one is something @dan13ram said as well: `colorname.num` rather than `var(--chakra-color-colornamenum)`. I'm uncertain about the dots. I would expect the format to be the same as with system props.

packages/web/config.ts
dan13ram: it does support polygon :)

luxumbra:
> My eyes sorta glazed over at times, but I didn't see anything that looked like it'd blow up in the near term. joy_cat
> 
> The most common one is something @dan13ram said as well: `colorname.num` rather than `var(--chakra-color-colornamenum)`. I'm uncertain about the dots. I would expect the format to be the same as with system props.

Thanks for taking the time on this @dysbulic. Now I am back in Chakra mode, yes, I think it can just be referenced as `colornamenum` in Chakra attributes. Will address feedback and push again in a cpl days. 🙏🏻

packages/web/utils/stringHelpers.ts
luxumbra: ✅

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: I have been trying to add the `<SwitchNetworkButton chainId="0x89" />` to the toast here but it doesn't render anything. Logging the values to the console shows `isMetaMask = false`  when I am using MM to connect.

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: ✔️

packages/web/components/Landing/LandingConnectButton.tsx
luxumbra: ✔️

packages/web/components/Landing/OnboardingGame/Chiev.tsx
luxumbra: ✔️

PR #1349, opened 2022-09-29
fix(login): fixed token verification

dan13ram:
## Overview

In the PR https://github.com/MetaFam/TheGame/pull/1324 we added a welcome message to the message being signed. But this broke the login flow since we aren't doing the same thing during the verification flow.

For already logged in users the old token works and you stay logged in but new users are NOT able to login right now.

This can also be verified by disconnecting and trying to connect again.

So this fix is urgent and needs to be moved to prod asap.

## Implementation

I'm adding the same welcome message to the message sent for verification just as we added it to the signing method.


alalonde (approved):
Quick catch! Just tested, looks good

Issue #1311, opened 2022-07-13
[Dashboard] "add new block" to the dashboard


peth-yursick:
### What would you like to be added?

Some block ideas
- My first suggestions is to do the same thing as in #1309 except linking to [this quest](https://metagame.wtf/quest/99047a66-533a-43a0-abc2-c9a9de85de46) instead.
- Other than that, I think the main one needed right now is "quests" #1310
- Quest Chains started get auto-pinned on the dashboard
- After that, it will probably be "your DAOhaus DAO proposals" as we're most likely to move to DAOhaus
Either way, no need to wait for 2) & 3) or anything else besides 1) really 
- Proposals from your Snapshot DAOs?
- Tasks assigned to your address on Dework
- Tasks assigned through whatever other bounty platform
- Issues assigned to your account on Github (I think we have github account verification?)
- Custom token stats block
- Chart block
- Lens feed
- Farcaster feed
- RSS feed
- Atlantis World block
- Carbon offset
- Txs queued in the multisig you're a member of?
- Tasks assigned to you on Dework?
- Maybe integrate something like RSS3 or even normal RSS feed? Maybe Lens or other web3 social media protocol posts?
- Maybe Dune Analytics?
- Feel free to be creative; it can be anything a DAO / world citizen might need or want.

### Why is this needed?

Making it obvious there should be an option to add more blocks to the dashboard seems like the obvious next step, the other part of it is making the feedback loop tighter like in #1309.

peth-yursick:
Migrating to #1512

PR #1350, opened 2022-09-30
[Dashboard] Persisting dashboard layout to hasura

### LINKED ISSUE #1347: [Dashboard] Persist dashboard layout changes to hasura

We need to store the json object that represents the current user's dashboard in their player row on our database. 

This can be implemented in a similar manner to how the player's profile layout is persisted in the database.

### Why is this needed?

Currently the dashboard layout is stored in the local storage, and is lost if user deletes local storage or tries to access their dashboard on another browser. Thus, we should persist it so the user sees the same dashboard when logging in from any device.

dan13ram:
## Overview

- I've implemented this in the same way as the player profile layout is persisted to hasura:
  - A serialised json string with the boxType and metadata is stored in the database
  - This json is used to recreate the dashboard layout when the user logs in.
- Since we need the user to login, the dashboard can only be viewed when logged in.

Closes #1347

- Support more block templates (https://github.com/MetaFam/TheGame/issues/1311)


vidvidvid (approved):
Really nice changes with the `<EditableGridLayout />`.
Tested moving stuff around and everything worked fine.
⚡️⚡️⚡️

PR #1364, opened 2022-10-04
Implemented graphql query for fetching PSeed price on polygon

### LINKED ISSUE #1361: Calculate $ amounts & add to the patrons leaderboard

Besides the total number of pSeeds, patron cards should also be displaying the $ value of those pSeeds.

The dollar value of a pSeed is derived by summing up total value of all tokens [inside the pool](https://app.balancer.fi/#/pool/0x0b09dea16768f0799065c475be02919503cb2a3500020000000000000000001a) & dividing it by the total number of [pSeeds](https://polygonscan.com/token/0x8a8fcd351ed553fc75aecbc566a32f94471f302e).

### Why is this needed?

To make it understandable how much $ money is required for each ranked league & allow the patrons conversion page (#1341) to pull in that data for league/perk unlock requirements.

alalonde:
## Overview

e.g.
```
query MyQuery {
  getPSeedInfo {
    priceUsd
  }
}
```
returns 
```
{
  "data": {
    "getPSeedInfo": {
      "priceUsd": "9.773548133805933"
    }
  }
}
```
Closes #1361

For some reason, this computes the value of the SEEDs in our pool to be about $52k, where [the pool](https://polygon.balancer.fi/#/pool/0x8a8fcd351ed553fc75aecbc566a32f94471f302e000100000000000000000081) values them at about $66k. Not sure where this discrepancy is coming from

### DIFF
--- a/packages/backend/src/handlers/remote-schemas/resolvers/balancerPolygon/resolver.ts
+++ b/packages/backend/src/handlers/remote-schemas/resolvers/balancerPolygon/resolver.ts
@@ -1,0 +1,36 @@
+import { Constants } from '@metafam/utils';
+import {
+  PSeedInfo,
+  QueryResolvers,
+} from 'handlers/remote-schemas/autogen/types';
+import { GetPoolTokenDataQuery } from 'lib/autogen/balancerpolygon-sdk';
+import { balancerPolygonGraphClient } from 'lib/balancerPolygonClient';
+
+export const getPSeedInfo: QueryResolvers['getPSeedInfo'] = async () => {
+  const poolData = await balancerPolygonGraphClient.GetPoolTokenData({
+    address: Constants.PSEED_ADDRESS,
+  });
+  const pseedPrice = computePSeedPrice(poolData);
+
+  return {
+    priceUsd: pseedPrice,
+  } as PSeedInfo;
+};
+
+const computePSeedPrice = (
+  poolData: GetPoolTokenDataQuery,
+): number | undefined => {
+  if (poolData?.pools?.length === 1) {
+    const poolTotalValueUSD = poolData.pools[0].tokens?.reduce(
+      (sum, poolToken) =>
+        sum +
+        parseFloat(poolToken.token.latestUSDPrice) *
+          parseFloat(poolToken.balance),
+      0,
+    );
+    if (poolTotalValueUSD != null) {
+      return poolTotalValueUSD / parseFloat(poolData.pools[0].totalShares);
+    }
+  }
+  return undefined;
+};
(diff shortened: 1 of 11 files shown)


packages/backend/src/handlers/remote-schemas/resolvers/balancerPolygon/resolver.ts
```diff
@@ -0,0 +1,36 @@
...
+  const pseedPrice = computePSeedPrice(poolData);
+
+  return {
+    priceUsd: pseedPrice,
+  } as PSeedInfo;
+};
+
+const computePSeedPrice = (
+  poolData: GetPoolTokenDataQuery,
+): number | undefined => {
```
dan13ram, line 22:
```suggestion
): number | null => {
```

packages/backend/src/handlers/remote-schemas/resolvers/balancerPolygon/resolver.ts
```diff
@@ -0,0 +1,36 @@
...
+        sum +
+        parseFloat(poolToken.token.latestUSDPrice) *
+          parseFloat(poolToken.balance),
+      0,
+    );
+    if (poolTotalValueUSD != null) {
+      return poolTotalValueUSD / parseFloat(poolData.pools[0].totalShares);
+    }
+  }
+  return undefined;
```
dan13ram, line 35:
```suggestion
  return null;
```

dan13ram (approved):
LGTM. Just a small change, afaik you need to return null from graphql apis when something is not found.

dan13ram:
Also with regards to the price, it is possible that the subgraph only updates the price when there is some activity on the pool and since there is no activity since 2 weeks it might be 2 weeks old data.

packages/backend/src/lib/balancerPolygonClient.ts
```diff
@@ -0,0 +1,10 @@
+import { GraphQLClient } from 'graphql-request';
+
+import { getSdk } from './autogen/balancerpolygon-sdk';
+
+const balancerPolygonGraphqlURL =
+  'https://api.thegraph.com/subgraphs/name/balancer-labs/balancer-polygon-v2';
```
dysbulic, line 6: Should this be in constants?

dysbulic (approved):
LGTM.

packages/backend/src/lib/balancerPolygonClient.ts
alalonde: I left this here because presumably it would only be used here

dysbulic: > I left this here because presumably it would only be used here

For me, the constants file is the place I'd go if wanting to customize the behavior of the application. If I were looking to find where that value is defined, I'd look in constants.

Issue #1287, opened 2022-06-04
[MyMeta] Guild profile images not queried & page format not aligned with player profiles


peth-yursick:
Guild profiles don't seem to be pulling in guild avatars. Not sure if this is true only of discord-based guilds or affecting some others as well but seems that for discord-based ones it doesn't work for sure.

Other than that, it would be good if guild profiles were formatted in the same way as player profiles.
Right now, guild page boxes may be hardcoded like player ones were before.

Player profiles:
![image](https://user-images.githubusercontent.com/55857222/171993075-5cb6a761-720a-4d11-bdcc-80a3ca955d7a.png)
Guild profiles:
![image](https://user-images.githubusercontent.com/55857222/171993098-d2b48d79-a7f3-4430-a702-93b1e8169b1a.png)

alalonde:
I think the issue here is that we are just hot-linking to guilds' logo images, so we have no recourse if e.g. the filename changes, the URL changes, etc. The Discord ones mostly do work, e.g. Opolis, Bloom

dan13ram:
Reopening since there is still the issue of Guild Logos, which is fixed in #1367

peth-yursick:
> Reopening since there is still the issue of Guild Logos, which is fixed in #1367

Didn't get this last bit "its still the issue, which is fixed"? 😂
I planned on manually uploading all guild logos to hasura today along with adding more links & info to each guild page and create a few of the missing guild pages (Atlantis World, Ark, Safary, rnDAO, YapDAO, Liminal Village, DAOhaus, Ethernaut DAO)..
Would be awesome to not have to do each of these steps manually

PR #1366, opened 2022-10-06
Support change guild profile layout

### LINKED ISSUE #1365: Changing default layouts (MyMeta & the dashboard)
This is an issue about changing the default block layout on MyMeta profiles & the dashboard.

Currently on MyMeta:
image
I think this should be the default:
image

The reason is that it looks more balanced by default + I think DAO memberships are more relevant than they appear like this on the bottom right block.

On the dashboard I'm not sure what the current default layout is because I can't check other people's, but I think it should be centered on the news/content & imo this looks most balanced as its symmetrical with the biggest block in the middle.
image

dan13ram:
## Overview

- Support changing guild profile layout
- Support add new block to guild profile
- Fix default layout for player profile

Closes #1365
Partially Closes https://github.com/MetaFam/TheGame/issues/1287

- [ ] Support new guild profile sections

## Implementation

- Re-used `EditableGridLayout` and it is implmented in the same manner as player profile as well as dashboard.


alalonde (approved):
🔥

Issue #1310, opened 2022-07-13
[Dashboard] Quests appearing on the dashboard


peth-yursick:
### What would you like to be added?

Right now, I have 15 quests live. If I want to check if anyone applied for quest completions, I need to go under quests & open each of them separately to see..
- If a user has posted a quest, there should be a block on the dashboard displaying applicants for quest completions.
- If a user has accepted or completed a quest, there should be a block on the dashboard displaying quests.
Longer term, there should probably be some kind of a notifications aka "journal" system but I think adding a block to the already existing dashboard is a good enough solution for now.

### Why is this needed?

So it stops being cumbersome & it becomes more likely that people will notice whats going on with their quests.

alalonde:
Not sure if the dashboard is the right spot for this. It is a pretty squished layout that needs to show a fair amount of data.

I am thinking of adding a separate page  `/quests/activity` with the following potential two blocks of content:

1. (if you have created at least one quest)
### Your quests
- <quest name> <proposal-by> 

2. (if you have submitted at least one proposal
### Your quest proposals
- <quest name>

luxumbra:
> I am thinking of adding a separate page /quests/activity with the following potential two blocks of content:

I agree it deserves a page to itself but can see the usefulness of having a block on the dashboard to give a summary of that info.  For some, i'd say it is more helpful to a user to see quests than the current leaderboard and prolly needs to be a block you can add if you want it. 

@peth-yursick what are your thoughts?

alalonde:
OK, turns out we already have a "Completed Quests" block that a player can add to their profile page. 

Really, here we want two new types of Dashboard blocks:
- Created Quests
- Completed Quests

We could pretty easily repurpose the current "Completed Quests" block to go on the dashboard (without the current filtering for only accepted quests).

peth-yursick:
Hey @alalonde, this included "quests I proposed to" not just "quests I posted", right?

alalonde:
Yes, there are two different blocks

PR #1368, opened 2022-10-07
New page for viewing quest activity

alalonde:

This solves @peth-yursick's main pain point with quests, though not in the way he suggested. This is a WIP, but it adds a new page `/quest/activity` which displays in a table all the quests you have created as well as any proposals for those quests:

image

I also made some minor UI improvements to some of the links in the quests pages.

See #1310 for the original proposal.

This is a WIP, again, which needs:
1. Either a link to this new page from `/quests` OR
  - Move the logic into a new dashboard block type "Created Quests". This would require finagling the UI since the dashboard columns are so narrow
2. Create a new dashboard block type "Completed quests" Which could be repurposed from the existing player page "Completed quests" block (which is poorly named as it only shows _accepted quests_). The only difference is that the dashboard "completed quests" should only show your own pending quest submissions. 

There are some assumptions here per requirements, and I am out of time for now, but we could certainly just ship this with just a link to the new "quest activity" page.


peth-yursick:
😱😱😱
Thank you very much, I know I've said it many times already but this will come in *very* handy & allow me to add more quests on there 😄

peth-yursick:
I played around with it for a bit. Def too narrow to have all 3 fields but maybe "applicant name" not necessary 🤷‍♂️
Made it look like the leaderboard widget:
image
If @davort approves, I'll update the original issue & put this design there. (unless you think that issue should be closed with this & a new one open?)

davort:
> I played around with it for a bit. Def too narrow to have all 3 fields but maybe "applicant name" not necessary 🤷‍♂️ Made it look like the leaderboard widget: image If @davort approves, I'll update the original issue & put this design there. (unless you think that issue should be closed with this & a new one open?)

Great job with reusing the leaderboard design!
Can't say I immediately understand how this should work and what it should do—e.g. what do the icons mean and why don't we use letters/numbers instead—but it definitely fits in the overall look & feel. I'm happy to help with future iterations of this design 😊

alalonde:
I took the path of least resistance and just added a link to this new "Quest Activity" page. This should be gtg

dan13ram (approved):
I like this. Can create a dashboard box in a separate PR imo

packages/web/components/Dashboard/config.ts
```diff
@@ -40,6 +40,7 @@ export const ALL_BOXES = [
   BoxTypes.DASHBOARD_SEEDS_INFO,
   BoxTypes.DASHBOARD_CALENDER,
   BoxTypes.DASHBOARD_LEADERBOARD,
+  BoxTypes.DASHBOARD_QUESTS,
```
dan13ram, line 43: this is not needed for now can be commented ?

PR #1373, opened 2022-10-08
fixed dashboard height issue

### LINKED ISSUE #1347: [Dashboard] Persist dashboard layout changes to hasura

dan13ram:
Closes #1347

### DIFF
--- a/packages/web/components/Dashboard/DashboardSection.tsx
+++ b/packages/web/components/Dashboard/DashboardSection.tsx
@@ -84,7 +84,8 @@ export const DashboardSection = forwardRef<HTMLDivElement, Props>(
       >
         <Flex
           w="100%"
-          h="100%"
+          minH="100%"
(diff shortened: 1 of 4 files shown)


luxumbra (approved):
I can confirm that dashboard is now usable on mobile. I'd say this fixes the issue.

I also noticed the blocks are not removable on mobile. I think the `IconButton` needs the `onTouchStart` props added? That fixed it on my local build anyway.  Could we fix that also while we are here?

PR #1374, opened 2022-10-08
temporary fix for guilds sequence

### LINKED ISSUE #1369: Changing the guilds sequence
At some point we'll have to set a filtering system but for now it would be enough to just rearrange some of them..

- MetaCartel Ventures shouldn't be no.1 as they haven't invested nor contributed in any way
- CRE8R DAO should be first first, since they are the only ones who actually bought the epic metamanifesto
- MetaCartel (not ventures), as they gave us a grant
- then Panvala for giving us PAN 
- then it could be DAOhaus, MetaFactory & Bloom Network
- Quest Chains, MetaFam & Raid Guild
- after that, it doesnt really matter as none of them bought MMs or watered Seeds 🤷‍♂️

dan13ram:
This is a very temporary fix. Not too happy about it but not sure whats the best way to do this anyway. 

Also the guildnames vary between production and staging so this won't work properly against staging backend.

Closes #1369

### DIFF
--- a/packages/web/pages/community/guilds.tsx
+++ b/packages/web/pages/community/guilds.tsx
@@ -1,14 +1,62 @@
 import { PageContainer } from 'components/Container';
 import { GuildList } from 'components/Guild/GuildList';
 import { HeadComponent } from 'components/Seo';
+import { GuildFragment } from 'graphql/autogen/types';
 import { getGuilds } from 'graphql/queries/guild';
 import { InferGetStaticPropsType } from 'next';
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


alalonde:
I probably would have just added an `order` column to the guild table that auto-increments. Then we could just set the ordering manually in each environment's hasura

dan13ram:
> I probably would have just added an `order` column to the guild table that auto-increments. Then we could just set the ordering manually in each environment's hasura

If you have time then please make that fix. 🙏🏽 
Can we close this PR then ?

alalonde (approved):
Let's push it through for now.

PR #1388, opened 2022-10-09
remove undeploy step

### LINKED ISSUE #1258: Tweak staging and production frontend deployments so there is less downtime

Currently, when pushing to `develop` or `master`, there is this automated github actions flow:

- Undeploy frontend
- Delete frontend container
- Build frontend
- Deploy frontend

Building the frontend takes over 20m, which means at least 20m of downtime. Seems like we should build the new container before undeploying the old one.

### Why is this needed?

We have a bunch of unnecessary downtime when updating staging and prod.

polimyl:
## Overview
fixes #1258 by removing the undeploy steps. this works because deploying to an existing instance just updates and and automatically reroutes traffic to the new revision with zero downtime

as it is, unused container images would start piling up since we cant just remove the old one now since it wont have a tag anymore. this would require some extra logic to find out the hash of the old image before building the new one and and deleting the old image by that hash after the new one got deployed

### DIFF
--- a/.github/workflows/gcp-deploy-frontend.yaml
+++ b/.github/workflows/gcp-deploy-frontend.yaml
@@ -106,42 +106,6 @@ jobs:
           echo "::set-output name=FRONTEND_TAG::\
           ${{steps.second.outputs.DOCKER_PATH}}/frontend:${{github.ref_name}}"
 
-  undeploy-frontend:
-    name: Undeploy Frontend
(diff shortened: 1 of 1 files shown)


luxumbra (approved):
Very happy this is done and would be good to see your improvement implemented in a future PR. :) Nothing is broken. LGTM 🚀

PR #1389, opened 2022-10-09
Fix deployment downtime (hotfix)

polimyl:
hotfix for issue caused by https://github.com/MetaFam/TheGame/pull/1388

another job depended on the removed ones, which makes the runs fail now. this pr removes those dependencies


luxumbra (approved):
LGTM.

PR #1392, opened 2022-10-10
1341 build the patrons conversion page update prices

kon-rad:
## Overview

This PR updates prices by 30% on the Patron League page.


luxumbra (approved):
Leaving my tuppence worth - just some minor, non-blocking improvements.All the links work as they should.  Would be good to get this up on the live site in the next push. 🚀  

Out of curiosity, should this content be centered on the page? 👇🏻 
image


Nothing from me that should block this from getting merged. Thanks for more updates @kon-rad 🙏🏻

dan13ram (commented):
Looks good to me. Only thing is if the URL should be `community/patrons/join` rather than `community/join/patrons`...
what do you think @peth-yursick @luxumbra

dan13ram:
Screenshot_2022-10-11-09-50-11-62_e4424258c8b8649f6e67d283a50a2cbc

The arrows could be smaller on mobile.

and the left side padding can be removed here

PR #1406, opened 2022-10-12
No more dependabot for npm packages

alalonde:


dan13ram (approved):

dan13ram:
Nice find. I didn't know that the config was in our own repo

Issue #1414, opened 2022-10-14
Add error tracking and ability to capture user feedback


luxumbra:
### What would you like to be added?

I would like error & exception tracking added to the MetaOS application. Coupled with this, I would also like to see a feedback widget added to the frontend to capture feedback from any users who are having issues without them needing to find Discord or another comms channel.

I would also like to see MS Clarity working again on the app to gather usability data like heatmaps, rageclicks, etc.

I have used [Honeybadger](https://honeybadger.io) and [Sentry](https://sentry.io/) in the past on smaller projects and found them incredibly helpful in squishing bugs that users encounter in the wild.

### Why is this needed?

I feel we should have had this earlier than v1.0 but imho this is essential now we are in v1.0 and if we are to take our product seriously. 

- Tracking errors and getting detailed reports in a dashboard/email/discord channel will give us way more insight into issues and the problems our users are experiencing.
- A feedback widget will give our users an easy way to report their issues and us a better overview of issues with the the app.
- Re-activating Clarity will give us additional data we can use to improve the product.

alalonde:
I 100% agree with the need for professional tooling to help with quality control. All of these things (with the exception of Clarity) I would say are mandatory.  Before implementing a tool, though, can we do some basic research around privacy and data harvesting?  I have used hotjar in the past but was never really comfortable with it TBH. It's basically pseudonymous spyware.  If we claim to be building a new web, then we need to abide by a higher standard.  (I'm skeptical that we would actually do anything with the data anyway. )

Let's definitely proceed with the error tracking and feedback widgets at least, provided they are not too much of a tradeoff with user privacy.

luxumbra:
> All of these things (with the exception of Clarity) I would say are mandatory. Before implementing a tool, though, can we do some basic research around privacy and data harvesting? I have used hotjar in the past but was never really comfortable with it TBH. It's basically pseudonymous spyware. If we claim to be building a new web, then IMO we need to abide by a higher standard. (I'm skeptical that we would actually do anything with the data anyway. )

Totally agree with the privacy aspect and holding higher standards than our web2 counterparts and will do some digging on the data harvesting aspect. 

I've pushed a branch which has Honeybadger and UserBack added and I've removed HotJar & Clarity. Userback is  is on a free trial but isn't too pricey and lets you customise the feedback forms and info gathered as you like.

PR #1419, opened 2022-10-16
Feat/1414 add error tracking & feedback

luxumbra:
## Overview

- Adds Honeybadger error tracking to the frontend and is also using a hook to post updates to the [#-🦡-honeybadger-reports](https://discord.com/channels/629411177947987986/1030538751279104110) in Discord. 
- Also adds [UserBack](https://userback.io) for handling user feedback and bug reports. WRT privacy concerns, while we look into where we stand with and how to handle user data, we can turn off features/fields that use/request this info. I do think that giving a user the option to add their email is important as it allows us to follow up if needed.
- HotJar & Clarity are not included after @alalonde comments in #1414.

 

#1414 

- [ ] Add to the backend?

## Implementation
Currently tested on local using env vars for setting `debug` &/or `reportData` `true` to send data to Honeybadger. 
[Honeybadger Docs for React](https://docs.honeybadger.io/lib/javascript/integration/react/)

All pretty self explanatory. I have gone through the frontend package and added `Honeybadger.notify(error)` to most places and enabled `enableUncaught` to capture uncaught errors/exceptions.
The public dashboard for errors (that have been shared) is here: [Public Dashboard](https://app.honeybadger.io/project/LWn2d4Bj1b?q=)

Ping me for any API keys,etc.

The UserBack config is done via their dashboard.

**Side effects**
Better visibility of what our frontend is doing in the wild. 

## Assets
image


packages/web/components/ConnectedPage.tsx
```diff
@@ -39,6 +40,7 @@ export const ConnectedPage: React.FC<{
   }
 
   if (error) {
+    Honeybadger.notify(error);
```
alalonde, line 43: I think the right way to do this would be to create a generic errorHandler function in web/utils. Then import Honeybadger in that function (_in one place_). So if we want to switch to another error handling service we only have to change the code in one place.

packages/web/config.ts
```diff
@@ -17,7 +17,13 @@ export const CONFIG = {
...
   publicURL: process.env.NEXT_PUBLIC_FRONTEND_URL,
   gaId: process.env.NEXT_PUBLIC_GA4_ID,
-  clarityId: process.env.NEXT_CLARITY_ID,
+  honeybadgerApiKey: process.env.NEXT_PUBLIC_HONEYBADGER_API_KEY,
+  honeybadgerReportData:
+    process.env.NEXT_PUBLIC_HONEYBADGER_REPORT_DATA === 'true',
+  honeybadgerDebug: process.env.NEXT_PUBLIC_HONEYBADGER_DEBUG === 'true',
+  userbackToken:
+    process.env.NEXT_PUBLIC_USERBACK_TOKEN ||
+    '37882|74240|mjn1zXGiYaozZR6qmqFfjl9Sk',
```
alalonde, line 26: These will need to be added to the dockerfile and github action as well I believe

packages/web/next-env.d.ts
```diff
@@ -4,3 +4,5 @@ declare module '*.png';
 declare module '*.jpg';
 declare module '*.jpeg';
 declare module '*.svg';
+declare module '@honeybadger-io/webpack';
+declare module '@honeybadger-io/js';
```
alalonde, line 8: Why is this necessary?

packages/web/config.ts
luxumbra: Correct. I did post in builders guild but a better place to mention it would have been in this PR. 😅

packages/web/next-env.d.ts
luxumbra: It seemed to fix an issue i was having when running my local build but I've removed them now and all is fine - I think the issue was my environment node version had changed from 16.13.1  to stable.

packages/web/components/ConnectedPage.tsx
luxumbra: > I think the right way to do this would be to create a generic errorHandler function in web/utils. Then import Honeybadger in that function (_in one place_). So if we want to switch to another error handling service we only have to change the code in one place.

just checking - leaving the `HoneybadgerErrorBoundary` in `_app.tsx` is fine?

alalonde: > just checking - leaving the `HoneybadgerErrorBoundary` in `_app.tsx` is fine?

Yup!

packages/web/config.ts
alalonde: I think adding these to the `docker/frontend/Dockerfile` should be part of this PR. You could then build the container locally and see if the variables make it in properly. LMK if you need help with that, some trial and error is probably involved. If it works in the PR deployment we should be good

alalonde: no, these would go in the frontend container which is not included in the docker-compose. So you'd have to manually `docker build` it. That only gets you part of the way anyway, probably the best thing to do would be to just push what you have. The test is if those env variables end up in the PR deployment on Google cloud

luxumbra: > probably the best thing to do would be to just push what you have. 

Pushing now.

packages/web/config.ts
```diff
@@ -17,7 +17,15 @@ export const CONFIG = {
     process.env.NEXT_BRIGHTID_NODE_URL || 'http:%2f%2fnode.brightid.org',
   publicURL: process.env.NEXT_PUBLIC_FRONTEND_URL,
   gaId: process.env.NEXT_PUBLIC_GA4_ID,
-  clarityId: process.env.NEXT_CLARITY_ID,
+  honeybadgerApiKey:
+    process.env.NEXT_PUBLIC_HONEYBADGER_API_KEY ||
+    'hbp_Z4027J7qYTktQkqvUHcOHJqIbpjpWD30A1kd',
```
alalonde, line 22: Should these be secrets?

alalonde:
I submitted feedback through the userback widget, did it go anywhere? Any ideas on how to trigger a honeybader error/report on that PR env?

luxumbra:
> I submitted feedback through the userback widget, did it go anywhere? Any ideas on how to trigger a honeybader error/report on that PR env?

Feedback received. Wrt triggering honeybadger, there's errors in the profile screen when editing & saving contents of blocks on a new account. They're handled already with toasts but should trigger a HB notify event.

packages/web/config.ts
luxumbra: Probably should be. I'll change the keys and remove from config.

luxumbra: > Should these be secrets?

Do you have access to the cloud account?

alalonde: no

alalonde (approved):

Issue #1412, opened 2022-10-14
Finishing the Patrons signup page


peth-yursick:
As these changes were only requested after @kon-rad was already mid-way through, there wasn't enough time to implement them before the launch.

The remaining 2 things:
- Automating ranked league $ requirements
   - With the help of #1361
- Updating the tiles to the new design
   - Old & new design side by side:
   - ![image](https://user-images.githubusercontent.com/55857222/195819380-81165b6e-67d3-4fe8-89cf-5ec69f25f376.png)
   - [Link to the figma file](https://www.figma.com/file/n36ggiRBQEGzRUbGxWfkMX/Meta-Game-(Copy)?node-id=5836%3A3265)

kon-rad:
I will take this one on, will be starting Monday, if that's alright.

PR #1428, opened 2022-10-22
1412 finishing the patrons signup page

kon-rad:
## Overview

Style changes on Patrons Join page.

https://github.com/MetaFam/TheGame/issues/1412

Closes #
https://github.com/MetaFam/TheGame/issues/1412

n/a

## Implementation

n/a

**Side effects**

n/a

## Assets

n/a


peth-yursick:
Hey, thank you!
Doesn't seem pSeed requirements are working, its displaying as "2,268.8 pSeeds" required for being no.1 but the top patron is actually only holding 633 pSeeds..

kon-rad:
> e

It looks like it's pulling the data in from the API. This value hasn't changed from what is currently on the live site. Let me investigate the value difference.

alalonde:
That’s because not all the patron eth addresses have created profiles

kon-rad:
> 

Does that mean, it is indeed displaying the correct value?

alalonde (changes requested):
I looked at the Figma and see a couple minor discrepancies: 

- The font size for the section headers seem too large 
- The "current req:" label is missing for each patron tier
- ..and for the one and only... block width should be the same as above

Also, unrelated, but the patron card time zone display has an erroneous display (e.g. for dan13ram it displays a random 5.5 to the right of the time zone). Not worth its own PR IMO and would be a super quick fix

kon-rad:
> I looked at the Figma and see a couple minor discrepancies:
> 
> * The font size for the section headers seem too large
> * The "current req:" label is missing for each patron tier
> * ..and for the one and only... block width should be the same as above
> 
> Also, unrelated, but the patron card time zone display has an erroneous display (e.g. for dan13ram it displays a random 5.5 to the right of the time zone). Not worth its own PR IMO and would be a super quick fix

• The font size of the section headers matches what is on the ['join guild' page](https://metagame.wtf/join/patron) - it is a 'Heading' element as 'h2' I think this is the correct element to use and that the figma be updated to this.

• I improved the styling on PatronTile time zone section and removed the time offset.

alalonde (approved):
Looks good! thanks

peth-yursick:
still not able to view deployment for some reason

luxumbra:
It's been merged to develop so the preview deployment will most likely have been deleted. You'll want to check it on the test URL I reckon. 

lux
---
Signed by luxumbra
Front-end developer, community manager & dweller of the Meta
***@***.***
twitter: @bdgrdev
web: metagame.wtf

Sent with Proton Mail secure email.

------- Original Message -------

peth-yursick:
The main issue I see is the text inside the boxes being bigger than box headings..
image

Here's everything in a list:
- Box content font looks bigger than box heading, shouldn't be
- The headings should probably be bigger rather than bold & "total of" shouldn't be bold either
- The box headings aren't capital letters as they are across the platform
- Box content looks like its either using a different font or a very thin version?
- Margins are much thinner than the rest of the platform or the design
  - image
- This last box was intentionally made wider so it can accommodate all text without breaking into a new line but looks like its breaking into a new line anyway..
  - image
- There's all this extra empty space in the bottom call-to-action cards, above button should be the same as below
  - image

Don't mean to be nitpicking but this is probably the most important page to get right as we really need to convert some new patrons & get on with fundraising..

But given this is already merged to test, should I open this as a new issue or how does it work?

alalonde:
@peth-yursick  Yes please create a new issue linking to this [comment](https://github.com/MetaFam/TheGame/pull/1428#issuecomment-1304459324)

PR #1431, opened 2022-10-26
Refactor: proper guild sorting

alalonde:
## Overview

As part of the 1.0 push, @dan13ram quickly and effectively implemented guild sorting, but admitted it was a bit of a hack where the guild order was hardcoded into a React component.

This PR adds a `guildPosition` column to the guild table. We can then set the guild positions directly in the database once this goes to prod.

I decided not to include a migration that would actually set the guild ordering, since only our prod environment needs it.

### DIFF
--- a/hasura/metadata/tables.yaml
+++ b/hasura/metadata/tables.yaml
@@ -62,10 +62,6 @@
       filter: {}
 - table:
     schema: public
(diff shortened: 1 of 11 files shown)


dan13ram (approved):
LGTM. 🔥

PR #1434, opened 2022-11-03
Disable arrows when submit proof modal is open

### LINKED ISSUE #1424: [Quest Chains] Pressing back arrow in the text box moves to the previous quests & loses the text
### What happened?

A user reported the interface switching the whole quest upon pressing back arrow instead of just going back a single letter in the input box - the text also disappears as the quest is changed.

Move back a letter.

Probably by going to any of the paths on the website https://test.metagame.wtf/roles

_No response_

_No response_

vidvidvid:
the arrows will still move the cursor correctly

## Overview

When the submit proof modal is open, we disabled arrow button functionality which navigates the user between quests.

Closes #1424 

## Implementation

I made a flag which is set to `true` if the modal is open and `false` if it's closed & added it to the CarouselContext.

### DIFF
--- a/packages/web/components/QuestChain/UploadProof.tsx
+++ b/packages/web/components/QuestChain/UploadProof.tsx
@@ -19,6 +19,7 @@ import {
   UseToastOptions,
 } from '@metafam/ds';
 import { contracts, graphql, helpers } from '@quest-chains/sdk';
+import { useCarouselContext } from 'components/Carousel/CarouselContext';
 import { useWeb3 } from 'lib/hooks';
 import { useDropFiles, useDropImage } from 'lib/hooks/useDropFiles';
 import { useInputText } from 'lib/hooks/useInputText';
@@ -43,6 +44,8 @@ export const UploadProof: React.FC<{
 }> = ({ refresh, questId, name, questChain }) => {
   const { chainId, provider, address } = useWeb3();
   const { isOpen, onOpen, onClose } = useDisclosure();
+
+  const { setIsSubmittingProof } = useCarouselContext();
 
   const toast = useToast();
   const toastIdRef = useRef<ToastId | undefined>(undefined);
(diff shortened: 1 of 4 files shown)


luxumbra (commented):
That is much better. 👍 

> the text also disappears as the quest is changed.

Less of an issue now but it would be quite nice if inputs, especially the textfield persisted if the user moves / accidentally clicks away from the modal.

vidvidvid:
> Less of an issue now but it would be quite nice if inputs, especially the textfield persisted if the user moves / accidentally clicks away from the modal.

done

packages/web/components/QuestChain/UploadProof.tsx
```diff
@@ -70,11 +73,11 @@ export const UploadProof: React.FC<{
   const { imageFile, onResetImage } = dropImageProps;
 
   const onModalClose = useCallback(() => {
```
luxumbra, line 75: I know I said "especially the textfield persisted" but imho it should persist the files and images too. What you think?

vidvidvid: sure! :)

vidvidvid: done!

luxumbra: I don't know if it is just me but I am getting mixed results when adding images with drag n drop and clicking doesn't open a file brower. Conversely, clicking the files field opens a file browser and adds the file but dropping a file doesn't work for me.

vidvidvid: All works for me and the changes i introduced shouldn't have influenced this behaviour.

luxumbra: Well, LGTM except for that and I am perhaps a bit of an edge case being on Linux.

luxumbra (approved):
All good

PR #1436, opened 2022-11-03
QC mobile glitches

### LINKED ISSUE #1421: Quest Chains integration glitching on mobile
As reported by a few users, looks like the scrolling down the paths glitches on mobile, making it impossible to use.
Trying to scroll down skips the quests & its not possible to swipe or scroll them back..

vidvidvid:
## Overview

The point of this PR was to make Role paths usable on mobile.
Disable horizontal dragging on mobile. Add buttons for navigating back and forth between subsequent quests.

Closes #1421 

- [ ] Enable horizontal dragging

### DIFF
--- a/packages/web/components/Carousel/Track.tsx
+++ b/packages/web/components/Carousel/Track.tsx
@@ -3,6 +3,7 @@ import {
   Flex,
   motion,
   useAnimation,
+  useBreakpointValue,
   useMotionValue,
   VStack,
 } from '@metafam/ds';
@@ -37,6 +38,7 @@ export const Track: React.FC = ({ children }) => {
   const controls = useAnimation();
   const x = useMotionValue(0);
   const node = useRef<HTMLDivElement>(null);
+  const isMobile = useBreakpointValue({ base: true, lg: false });
 
   const handleDragStart: DragHandler = () => {
     setDragging(true);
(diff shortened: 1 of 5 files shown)


packages/web/components/Carousel/Track.tsx
```diff
@@ -139,8 +141,8 @@ export const Track: React.FC = ({ children }) => {
         <VStack ref={node} spacing={5} alignItems="stretch">
           <MotionFlex
             dragConstraints={node}
-            onDragStart={handleDragStart}
-            onDragEnd={handleDragEnd}
+            onDragStart={isMobile ? undefined : handleDragStart}
```
luxumbra, line 144: I it is an improvement but it is still super easy to skip to the next step by accident. The impact of that is less so now after you've fixed the persistence in the inputs but it's still pretty disorientating. Is it possible to make it so the active drag area is the same as the visible portion of the next step?

Also, in my changes I didn't add a check for the prev/next buttons when at the start/end of the track.

vidvidvid: how is it super easy to skip to the next step by accident? isn't the only way one could do that by pressing the tiny edge of the next quest?

vidvidvid: >  Is it possible to make it so the active drag area is the same as the visible portion of the next step?

but it is like that atm, isn't it? one can't drag the current quest.

> Also, in my changes I didn't add a check for the prev/next buttons when at the start/end of the track. 

will add this 👌

luxumbra: > how is it super easy to skip to the next step by accident? 

Actually, giving this another go today it seems fine.

vidvidvid: :D pls give it another look, i hid the relevant buttons on first/last step

luxumbra (approved):
LGTM. 🥳

PR #1437, opened 2022-11-07
Add completed and created quests to the dashboard

### LINKED ISSUE #1310: [Dashboard] Quests appearing on the dashboard

alalonde:
Continues the work done in #1368 to add a similar quest view to the dashboard.

Closes #1310

## Assets

Dashboard box:
image

"View all" modal:
image


alalonde:
@peth-yursick @davort I finished the first pass at this. The second box probably needs a little work at least; I'm open to input

alalonde:
Todo still:

- [ ] Remove old quest activity page
- [ ] Add UI tip to quests explorer for people to be able to find these new dashboard boxes
- [ ] Refactor the `MetaFilterSelectSearch`

peth-yursick:
Nice! I was gonna say that it should only be displaying pending submissions rather than a full list but then I saw there's a toggle.
Should be set to "pending only" by default but might even be unnecessary altogether only adding to clutter, I am only ever interested in pending submissions & cant imagine why would anyone want to have a full list of submissions for all their quests displayed - but thats just me 🤷‍♂️

Great seeing this happen, really looking forward to having it in the dashboard, thank you! 🙏

Issue #752, opened 2021-06-28
Playbooks 


peth-yursick:
Playbooks are short onboarding & skill-building guides.
Right now they are hosted [on the wiki](https://wiki.metagame.wtf/docs/playbooks/browse) but we need to build them a proper interface to make it easier for people to submit them as well as upvote/downvote them, filter them, etc.

The design should be based on the [existing MetaGame branding](https://skinny-estimate-0fd.notion.site/Design-MetaGame-1f9977cfaf3e4c29aeaa225734152f22) & some of this [design already exists for Playbooks as well](https://skinny-estimate-0fd.notion.site/Playbooks-UI-5ae958cfe1e945658deecd3a78764a0a).

The design still needs to be finished & help is required but we mainly need help with building. You can find the links to everything you'll need to [get familiar with the codebase here](https://skinny-estimate-0fd.notion.site/Builder-Path-aed890599bc743d8896f6f3249693ca2), you can [ask for help here](https://discord.gg/6JFXC9T) or contact me @petheth on twitter or telegram.

Useful things to get built:
- Playbook posting itself
- Categories & filtering
- Upvoting/downvoting playbooks
- Time to read
- Reacting to playbooks
- Comment
- Share
- Fork
- Bookmark (pin to your "dashboard" or "goals" both of which are nonexistant features)

We realize this is far from being specced out & requires more than 2 people, but if you're interested in jumping on this, do know there are others ready to jump in & help as well.
If you're interested in working on this, just comment here or [ask about it on discord](https://discord.gg/6JFXC9T).

More links:
[Playbooks Spec WIP](https://docs.google.com/document/d/1OZlidtGFbhNFb6aKnEJBR7ycnduXamChaW2s8pchdzE/edit#)
[More info on the whole Playbooks project](https://meta-game.notion.site/Playbooks-UI-5ae958cfe1e945658deecd3a78764a0a)
[Playbook WIPs on Notion](https://www.notion.so/Write-a-Playbook-3d8b963053f5457091abfe74f06248c9)
[📕-playbooks discussion on Discord](https://discord.gg/TCFdHCr)
[Playbooks UI WIP](https://www.figma.com/file/EDj3DFVrV2ms6QA7Q8gRdp/Playbooks-UI---Quick-Questions?node-id=0%3A1)

alalonde:
At Peth's request, I'm attempting to lay out a path towards an MVP for adding playbooks to the existing codebase (commonly known as MyMeta).

[...]

PR #1438, opened 2022-11-07
New cards

vidvidvid:
## Overview

Updated cards design for:
- [x] players
  - [x] fix badges with XP
  - [x] ~~generated octo avatars (dall e API or stable diffusion (probably an idea for the future))~~
- [x] guilds
  - [x] funding guild / social guild tags?
  - [x] barrier of entry
  - [x] skills
- [x] role paths
  - [x] Design
  - [x] Pictures
  - [x] Difficulty of the Role Paths
  - [x] Time needed for Role Paths
- [ ] #752 
- [x] patrons
  - [x] reuse player cards
  - [x] update the xp flag with $
- [ ] quests
  - [x] design
  - [x] disable the js tilt in the quest view 
  - [ ] #1446 

Issues:
- [x] Fix the `Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.` in cards. This happens because the whole card is wrapped into a link & there are more links inside of it.

Closes #

## Implementation

Used [vanilla-tilt](https://www.npmjs.com/package/vanilla-tilt) package for the hover effect.

#### Design:
https://www.figma.com/file/n36ggiRBQEGzRUbGxWfkMX/Meta-Game-(Copy)?node-id=6701%3A6415

#### Preview:
https://user-images.githubusercontent.com/35112939/200346493-db98eeed-9d79-4c90-98a6-c03d9d43515d.mov


peth-yursick:
As requested, some issues:
image
- [ ] All profile photos should be the same size square, cropped when necessary
- [ ] The photo part & text part should all be same size on all cards
- [ ] "Contact" & "member of" should always be displayed at the bottom of the card & "skills" should also always float on the same part of the card regardless of how big the bio is, whether it exists or not etc.
- [ ] Should be using emojis instead of the "diamond"
- [ ] All card starting in row 4 have this weird white glow when hovering the bottom part of the card
   - image
- [ ] If you hover the top part of the bottom row cards & scroll while hovering, the effect stays "locked" on the top part hover even after moving the mouse to the bottom part of the card
- [ ] The whole cards should be clickable, currently the outer margins of a card aren't (actually on players they aren't but on patrons looks like they are wholly clickable - patrons have a different problem tho)
- [ ] Patrons page is marked as complete but looks more like the old version to me
   - image
- [ ] Guilds are only displaying MCV & MetaFam for some reason with MetaFam image broken
   - image
- [ ] The join button should be removed so people open the pages before going to their website
- [ ] Should also all be same size & proportions regardless of image dimensions, amounts of text etc.
- [ ] Also outer margins of the card not clickable

That's all I'm seeing right now but also probably more than enough 😂

luxumbra:
> Should be using emojis instead of the "diamond"

I'd also add that in the Figma, these bits are square aspect.

From a design perspective, I kinda feel the radius on corners of elements *inside* the cards (cut out, title, etc) should match - 10px feels too heavy for the card title. 
image

vidvidvid:
> Overall, all cards be same size & proportions regardless of image dimensions, amounts of text etc.

I can make all of the cards in the same row the same size - that's the simplest solution. Also the differences between rows won't be noticeable, i'd prefer not to make the height and width static.

peth-yursick:
> From a design perspective, I kinda feel the radius on corners of elements _inside_ the cards (cut out, title, etc) should match - 10px feels too heavy for the card title.

Agreed, all edges are rounded at 10, now rounded the inner square & membership/contact squares to 7, looks a bit better
image

> I can make all of the cards in the same row the same size - that's the simplest solution. Also the differences between rows won't be noticeable, i'd prefer not to make the height and width static.

Yeah, I don't think the size should be static in terms of actual pixels, just the proportions of each part in relation to others, if that makes any sense 😂
What do you mean the difference between rows won't be noticeable?

vidvidvid:
@peth-yursick 
> Guilds are only displaying MCV & MetaFam for some reason with MetaFam image broken

also in the testing env the img is broken, but it should be visible in production.
https://test.metagame.wtf/guilds

luxumbra (commented):
Looking sweet @vidvidvid but I still find the contrast on player names is less than ideal where folks have lighter avatars. I think a darker color...perhaps `landingGlassDark` would work better for the bg of the name. 
image

I also feel the text in the insets should be white or at least lighter to improve contrast now that the bg is darker.
image

vidvidvid:
@luxumbra oukidouki, i will fix that. btw what is that addon which shows you the contrast?

dysbulic (approved):
Looks reasonable to me given the changes that are in #1448 to some stylistic issues.

vidvidvid:
@luxumbra @dysbulic i fixed your suggestions! :)

Issue #1267, opened 2022-05-12
Metagame.wtf: Players - After I came back from a player's profile, my filter was gone.


vance4243:
### What happened?

On the https://metagame.wtf/players page: I sorted the players by Username (A-Z), scrolled down and clicked on a profile. When I went back to the list of players, my filter was gone and I was put to the beginning of the page.

### What did you expect to happen?

Continue scrolling through the sorted Players from where I stopped.

### How can we reproduce the problem (as minimally as possible)?

1. Got to https://metagame.wtf/players.
2. Sort by Username (A-Z).
3. Scroll down and click on a profile.
4. Go back to list of Players

### Browser and operating system used (if applicable)

Firefox, Win 11

vance4243:
When I open the Timezone filter and try to click on some Timezone in the lower part, the Player underneath the menu gets opened.

alalonde:
Yes, the solution to this is to update the URL when filters are applied; essentially storing the filter state in the the URL. This way back/forward functionality works as expected

PR #1439, opened 2022-11-07
Show all the patrons on the patrons page (1417)

### LINKED ISSUE #1417: [MyMeta] Patrons leaderboard only loading top 48 patrons
When you scroll down to the bottom of the patrons leaderboard, its not loading any more than the first 48.
There's a person holding 8pSeeds who reported that issue.

Also, same issue as in quests - patron cards go all the way to the edge of the screen instead of there being a margin..

HHH-GH:
## Overview

**1. only 48 patrons shown**  
> When you scroll down to the bottom of the patrons leaderboard, its not loading any more than the first 48.  

Now loads all the patrons (max 150) and displays them in sets of 30, with a Load More button to show the next 30 patrons.

**2. Cards go all the way to the bottom of the screen**  
> patron cards go all the way to the edge of the screen instead of there being a margin.  

Added a back to top link at the bottom to provide a margin between cards and edge of the screen

Closes #1417 

- [ ] If clicking in to view player details, and then clicking the back button to return to the list, it should go back to where they were on the list before. I *think* it does this, will test again when it deploys.

## Implementation

**Showing all the patrons part**  
- getPatrons query set to retrieve 150 patrons
- Standard (?) ‘load more’ button based on various Google results
- Load more button is disabled when there are no more patrons to display
- Below the Load More button it shows a count of patrons, so people can see how many times to click before they'll be at the end
- More comments (maybe too many) in the code

**Back to top link**  
- Copied it from other pages
- The patrons page doesn't have a heading to use as a ref for the back-to-top target, so I added a heading that is not shown on the screen (Chakra's VisuallyHidden component)

- Added the VisuallyHidden component to the design system import

**Side effects**
n/a

## Assets

Load more button and etc  👇
image

Load more button disabled, with tool tip 👇
image


HHH-GH:
Seems like this doesn't work - the position in the list of patrons is lost when clicking the back button after looking at a patron.

Try it
1. Load the patrons page
2. Click the load more button to ... load more
3. Click into a patron
4. Then click the back button in the browser to go back to the list
5. It's reset to the first 30 patrons

Any hints on what to look at here? (or is it not super important)

> ## Follow up Improvement Ideas
> 
>     * [ ]  If clicking in to view ~~player~~ patron details, and then clicking the back button to return to the list, it should go back to where they were on the list before. I _think_ it does this, will test again when it deploys.

alalonde:
> Any hints on what to look at here? (or is it not super important)

Yes, notice also how when you change a filter and then click the back button, it doesn't go back to your last filter state.  This is because we're not persisting the filter / sort state in the URL. Same deal with the "load more" button. When you go back, the browser doesn't just remember that you had clicked "load more" before, since we didn't persist something like `display=60` in the URL.

I don't think fixing this is in the scope of this ticket.  There is already [a ticket](https://github.com/MetaFam/TheGame/issues/1267) for this general problem.

HHH-GH:
Okay, I'll leave that, then.

@peth-yursick    If the patrons’ page now looks like you wanted it to, I'll open it for a code review.

HHH-GH:
Ready for review.

Couple of notes:

**1. A weird thing with the `getPatrons()` query**  
With no limit set for the query, it was returning 33 patrons. (Using the staging DB I guess)  
With a limit of 50 e.g. `getPatrons(50)`, it was returning 42 patrons.

**2. Not using spread operator when passing `patrons` to the `PatronsList` component**  
Does that matter? It looks to be working fine. 

*Original version:*
```
<PatronList {...{patrons, pSeedPrice }} />
```

*New version:*  
```
<PatronList
  patrons={patrons.slice(0, visible)}
  pSeedPrice={pSeedPrice}
/>
```

**3. Errors on my local dev**  
`Error: Text content does not match server-rendered HTML.`  
I tracked it down to the `PatronList` component. Maybe because the `a` tag is wrapping a `div` block, which isn't valid HTML? I'm not seeing that error on the current patrons page, so will wait and see what happens when this latest version is deployed.

alalonde:
> **1. A weird thing with the `getPatrons()` query** With no limit set for the query, it was returning 33 patrons. (Using the staging DB I guess) With a limit of 50 e.g. `getPatrons(50)`, it was returning 42 patrons.

Yes, I believe that we end up filtering out patrons that don't have mymeta profiles linked. We should really do that filtering before applying the count, though.

> **2. Not using spread operator when passing `patrons` to the `PatronsList` component** Does that matter? It looks to be working fine.

Nope, they transpile to be the same thing :)

alalonde (approved):

HHH-GH:
I'm closing this one because I messed up the commit history.

Made a clean version -> [Show all the patrons on the patrons page (1417) (v2) #1464](https://github.com/MetaFam/TheGame/pull/1464)

PR #1449, opened 2022-11-15
Fix dev environment

alalonde:
Nobody can run the backend locally in Docker; this fixes that.

It uses `nodemon` since `ts-node-dev` doesn't support esm and `ts-node` doesn't support hot reloading.


vidvidvid (approved):
gg, the app starts for me now! :)

PR #1470, opened 2022-12-01
Use consistent underlying images for all docker containers

alalonde:
We have been building all our docker containers using the image 
`FROM node:16-slim as base`
but when we deploy we use
`FROM node:16-alpine as app`
and then just copy over the files.  This can cause runtime errors due to the underlying environment being different.


polimyl (approved):
lgtm

Issue #1468, opened 2022-11-29
The filters for quest selection are not visible


dysbulic:
### What happened?

The filter bar for selecting quests appears to have `overflow: hidden` set so the dropdowns don't display:

![Screenshot from 2022-11-28 16-26-03](https://user-images.githubusercontent.com/181523/204385125-c47a54a0-f3ac-4476-9460-f6d4ccd2bb2f.png)

When `overflow: hidden` is disabled, on the `div` near the options, it becomes apparent that `overflow: hidden` is also set on another `div` higher in the tree.

![Screenshot from 2022-11-28 21-36-21](https://user-images.githubusercontent.com/181523/204425464-34dbd09a-ed3d-4a40-8480-8bfc71ea8d29.png)

### What did you expect to happen?

You should be able to see the filter options.

### How can we reproduce the problem (as minimally as possible)?

Go to [test.metagame.wtf/quests](//test.metagame.wtf/quests) and try to filter the results.

peth-yursick:
test.metagame won't load for me for some reason but seems to be working normally for me in prod
![image](https://user-images.githubusercontent.com/55857222/205033457-6bd00012-f0a4-4b36-a769-9bb0952c1ec5.png)

luxumbra:
> ### What happened?
> The filter bar for selecting quests appears to have `overflow: hidden` set so the dropdowns don't display:
> 
> ![Screenshot from 2022-11-28 16-26-03](https://user-images.githubusercontent.com/181523/204385125-c47a54a0-f3ac-4476-9460-f6d4ccd2bb2f.png)
> 
> When `overflow: hidden` is disabled, on the `div` near the options, it becomes apparent that `overflow: hidden` is also set on another `div` higher in the tree.
> 
> ![Screenshot from 2022-11-28 21-36-21](https://user-images.githubusercontent.com/181523/204425464-34dbd09a-ed3d-4a40-8480-8bfc71ea8d29.png)
> 
> ### What did you expect to happen?
> You should be able to see the filter options.
> 
> ### How can we reproduce the problem (as minimally as possible)?
> Go to [test.metagame.wtf/quests](//test.metagame.wtf/quests) and try to filter the results.

Which browser/os is this? I am not experiencing it on Chrome/Linux

PR #1461, opened 2022-11-25
Squashed Adding of an Image to Quests

dysbulic:
This request is a squashed version of #1448 by @thebeyondr. The original request was:

## Overview

This adds:
- The ability to add images when creating a quest
- Image preview on the form (wasn't in the issue but it just made sense 🤷‍♂️)
- Updated spacing of form items
- Quest image upload to `web3.storage`

Closes #1446, #1465, #1466, & #1467.

## Follow-Up Improvement Ideas

- Create greater contrast between form and background. Form labels are sometimes hard to read.

## Side Effects

- The `image` property that is stored on the quest is returned as `ipfs://` so retrievers will need to use the `httpLink` helper when accessing `quest.image` to receive a URL to be used in `<Image src=*/>`

- Fixing a hydration error currently present in `develop` as well as a misconfiguration of `NODE_ENV` for the test instances was necessary to properly test this PR.

## Attributions

* Paired-With: @thebeyondr

PR #1475, opened 2022-12-06
Fixing Cropping of Filter Options

### LINKED ISSUE #1474: `PlayerTile` Width Is ~0 With < 3 Entries
### What happened?

I'm working with an unseeded database with a single entry in it & the players' leaderboard looks like:

Screenshot from 2022-12-06 03-41-45

For it not to look like that. 😜

Create a new instance, add an address, view the leaderboard.

_No response_

_No response_

dysbulic:
## Overview

`Wrap` components, by default, set `overflow` to `hidden`. This PR sets `overflow` to `visible` for the containers of the filters so they aren't cropped.

Also, it switches the outer `Wrap` in the `QuestFilter` to a `Flex`, and puts the children of the inner `Wrap` in `WrapItem`s since `Wrap` generates a `ul` & it was full of `div`s.

Closes #1474 & #1468.

## Follow Up Improvement Ideas

This is a fairly targeted PR. I specifically avoided syntactic fixes because I wanted to avoid conflicts with #1461.

### DIFF
--- a/packages/web/components/Player/Filter/DesktopFilters.tsx
+++ b/packages/web/components/Player/Filter/DesktopFilters.tsx
@@ -68,10 +68,10 @@ export const DesktopFilters: React.FC<Props> = ({
 
   return (
     <Wrap
+      ref={filterRef}
       transition="all 0.25s"
       py={6}
-      style={{ backdropFilter: 'blur(7px)' }}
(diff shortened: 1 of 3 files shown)


vidvidvid (approved):
o👍

Issue #1472, opened 2022-12-04
Update guild join flow to be more useful


alalonde:
### What would you like to be added?

- After connecting, button should just say "Join"
- More copy on the intro page explaining why going through Discord is necessary (from peth "better explain why its useful & what we'll be pulling in, currently only mentions announcements"). Currently we only describe why one might want to leave the permission box checked in the Discord flow
- Update final toast message to say something like "Thanks, your Guild will go live shortly!" 
- [Optional] On submission, post a message to champion's ring (or wherever appropriate) so someone can go into Hasura to activate the guild

### Why is this needed?

People will be manually sent a link to this page after they have already filled out a form and manually approved, so at this point they are pre-vetted and should be able to freely proceed.

However, anyone with the link could still input their data, so a manual approval process still seems useful

cmaxwe11:
What worked:
1. Discord linking worked, Metagame bot is now a member of the server (what commands should I test with, what is expected to work on a non-Metagame Discord?)
2. Pulled in roles for me to select Admin/Member roles. 

I got stuck trying to submit guild info, a contract is required (which I believe should not be, as not all partner orgs will have on-chain DAOs). I tried just submitting a wallet address, and got this error:
![image](https://user-images.githubusercontent.com/116289711/205688459-df5f7077-1c42-498c-975f-bd3e31c1eade.png)

cmaxwe11:
Per feedback from Alec, the default DAO can be removed, and I was able to submit. Consider defaulting to not including the entry?

Seroxdesign:
Guild flow has been improved VIA multiple PRs

PR #1479, opened 2022-12-08
Some minor updates to guild join flow

alalonde:
## Overview

See #1472 for context. 

Our new guild onboarding flow requires potential guilds to first fill out a form (externally) to gauge fit.  Once they get over that hurdle, they are sent to the "Join as a Guild" page to submit their information to our database.  This PR:

- adds additional copy about why we require linking with their Discord server
- Removes the default placeholder DAO from the form, as suggested by maxwell 
- On form submission, sends a message to the champions league channel so we don't miss it

Note that I had to explicitly give MetaGameBot the "send message" permission to the champion's league channel within Discord.

`/join/guild/start` is the path for testing

### DIFF
--- a/packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
+++ b/packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
@@ -1,11 +1,19 @@
-import { GuildDiscordMetadata } from '@metafam/discord-bot';
+import {
+  createDiscordClient,
+  GuildDiscordMetadata,
+} from '@metafam/discord-bot';
+import { Constants } from '@metafam/utils';
+import { TextChannel } from 'discord.js';
 import { Request, Response } from 'express';
 
+import { CONFIG } from '../../../config.js';
 import {
   Dao_Player,
   Guild_Set_Input,
   GuildDaoInput,
+  GuildFragment,
   GuildInfoInput,
+  GuildStatus_Enum,
   GuildType_Enum,
 } from '../../../lib/autogen/hasura-sdk.js';
 import { client } from '../../../lib/hasuraClient.js';
@@ -23,17 +31,37 @@ export const saveGuildHandler = async (
   const { input, session_variables: sessionVariables } = req.body;
   const playerId = sessionVariables['x-hasura-user-id'];
 
+  const { guildInformation } = input;
+
+  let updatedGuild: GuildFragment;
   try {
-    const { guildInformation } = input;
-    await saveGuild(playerId, guildInformation as GuildInfoInput);
-    res.json({ success: true });
+    updatedGuild = await saveGuild(
+      playerId,
+      guildInformation as GuildInfoInput,
+    );
   } catch (error) {
     console.error(error);
     res.json({
       success: false,
       error: (error as Error).message,
     });
+    return;
   }
+
+  const isNew = updatedGuild.status === GuildStatus_Enum.Pending;
+
+  if (isNew) {
+    try {
+      await sendDiscordNotification(guildInformation);
+    } catch (error) {
+      console.error(
+        "Error sending notification to Champion's League channel",
+        error,
+      );
+    }
+  }
+
+  res.json({ success: true });
 };
 
 const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
@@ -149,4 +177,21 @@ const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
     guildId: guildInfo.uuid,
     discordMetadata: updatedMetadata,
   });
+
+  const { guild } = await client.GetGuild({ id: guildInfo.uuid });
+  return guild[0];
 };
+
+async function sendDiscordNotification(
+  guildInfo: GuildInfoInput,
+): Promise<void> {
+  const discordClient = await createDiscordClient();
+
+  const targetChannel = (await discordClient.channels.fetch(
+    Constants.METAFAM_DISCORD_CHAMPS_RING_CHANNEL_ID,
+  )) as TextChannel;
+  const link = `${CONFIG.hasuraAdminURL}/data/schema/public/tables/guild/browse?filter=guildname%3B%24eq%3B${guildInfo.guildname}`;
+  targetChannel.send(
+    `A new guild signed up! Name: ${guildInfo.name}  Someone with access to the Hasura instance can approve them here: ${link}`,
+  );
+}
--- a/packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
+++ b/packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
@@ -48,16 +48,18 @@ export const saveGuildHandler = async (
     return;
   }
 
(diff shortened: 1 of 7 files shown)


packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -149,4 +177,21 @@ const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
...
     discordMetadata: updatedMetadata,
   });
+
+  const { guild } = await client.GetGuild({ id: guildInfo.uuid });
+  return guild[0];
 };
+
+async function sendDiscordNotification(
+  guildInfo: GuildInfoInput,
+): Promise<void> {
```
dysbulic, line 187: Given the name of the function, I think it makes sense for the channel id to be a param.

```suggestion
async function sendDiscordNotification(
  guildInfo: GuildInfoInput,
  channelId = Constants.METAFAM_DISCORD_CHAMPS_RING_CHANNEL_ID,
): Promise<void> {
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -149,4 +177,21 @@ const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
...
+  const discordClient = await createDiscordClient();
+
+  const targetChannel = (await discordClient.channels.fetch(
+    Constants.METAFAM_DISCORD_CHAMPS_RING_CHANNEL_ID,
+  )) as TextChannel;
```
dysbulic, line 192:
```suggestion
  const targetChannel = (await discordClient.channels.fetch(
    channelId,
  )) as TextChannel;
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -149,4 +177,21 @@ const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
...
+  const link = `${CONFIG.hasuraAdminURL}/data/schema/public/tables/guild/browse?filter=guildname%3B%24eq%3B${guildInfo.guildname}`;
```
dysbulic, line 193: Just to be extra careful about the input *(an `&`, for example, I'm pretty sure would mess up the parse)*:

```suggestion
  const link = `${CONFIG.hasuraAdminURL}/data/schema/public/tables/guild/browse?filter=${encodeURIComponent(`guildname;$eq;${guildInfo.guildname}`)}`;
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -149,4 +177,21 @@ const saveGuild = async (playerId: string, guildInfo: GuildInfoInput) => {
     guildId: guildInfo.uuid,
     discordMetadata: updatedMetadata,
   });
+
+  const { guild } = await client.GetGuild({ id: guildInfo.uuid });
+  return guild[0];
```
dysbulic, line 182: Does this work:

```suggestion
  const { guild: [guild] } = await client.GetGuild({ id: guildInfo.uuid });
  return guild;
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -23,17 +31,37 @@ export const saveGuildHandler = async (
...
       error: (error as Error).message,
     });
+    return;
   }
+
+  const isNew = updatedGuild.status === GuildStatus_Enum.Pending;
+
+  if (isNew) {
+    try {
+      await sendDiscordNotification(guildInformation);
```
dysbulic, line 55: I'd name this function something more specific to what it is doing:

```suggestion
      await requestChampsApproval(guildInformation);
```

That or I'd not have the method know that it's dealing with a guild:

```suggestion
      const queryParam = encodeURIComponent(`guildname;$eq;${guildInfo.guildname}`)
      const link = (
        CONFIG.hasuraAdminURL
        + '/data/schema/public/tables/guild/browse'
        + `?filter=${queryParam}`
      );
      const message = (
        `A new guild signed up named: "${guildInfo.name}"!`
        + `\n\nSomeone with access to the Hasura instance can approve them here: ${link}`
      );
      await sendDiscordNotification(message);
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -23,17 +31,37 @@ export const saveGuildHandler = async (
...
+
+  let updatedGuild: GuildFragment;
   try {
-    const { guildInformation } = input;
-    await saveGuild(playerId, guildInformation as GuildInfoInput);
-    res.json({ success: true });
+    updatedGuild = await saveGuild(
+      playerId,
+      guildInformation as GuildInfoInput,
+    );
```
dysbulic, line 41: I'd move all the rest of this function into this `try` block. Since the error handler returns this is the same basic logic, but having `updatedGuild` outside the `try` makes it look as though it will be used afterwards which is only the case if there isn't an error.
```suggestion
    const updatedGuild = await saveGuild(
      playerId,
      guildInformation as GuildInfoInput,
    );
```

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
```diff
@@ -23,17 +31,37 @@ export const saveGuildHandler = async (
...
   } catch (error) {
     console.error(error);
     res.json({
       success: false,
       error: (error as Error).message,
     });
+    return;
```
dysbulic, line 48: If the rest of the method below is moved into the `try`, you don't need this `return`.

```suggestion
```

packages/web/components/Guild/GuildForm.tsx
```diff
@@ -529,7 +528,7 @@ export const GuildForm: React.FC<Props> = ({
             </Box>
           ))}
           <MetaButton size="sm" onClick={() => append(placeholderDaoInput)}>
-            Add another
+            Add {daoFields.length > 0 ? 'another' : 'DAO'}
```
dysbulic, line 531:
```suggestion
            Add {daoFields.length > 0 ? 'Another' : 'a DAO'}
```

dysbulic (approved):
Everything seems logically sound. I had a couple structural criticisms of things I'd have done slightly differently.

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
alalonde: ah yes, good call

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
alalonde: Yes but IMO is less readable

packages/backend/src/handlers/actions/guild/saveGuildHandler.ts
alalonde: Well, then we'd have nested try/catch blocks which I don't like. The catches need to be separate as one is fatal and the other can be ignored.

Issue #1492, opened 2023-01-08
Build a tool for users to be able to migrate their existing profile data into ComposeDB


alalonde:
Existing Ceramic data cannot be automatically ported over to ComposeDB. Thus, we'll need to build a backend flow for users to trigger exporting their own data from Ceramic and importing it into ComposeDB. Ideally this is just a click of a button and signing a few transactions.

This sort of thing is inherently error-prone, so we will need to est, troubleshoot and maintain the tool for ~6 months after launch.

dysbulic:
Do you want to just use the Hasura cache, or do you want to maintain a legacy Ceramic tiles server?

alalonde:
Great question. We should only support migrating for a certain amount of time, so we could in theory just keep the existing Tile server as the original "source of truth".   However it seems like our Hasura cache is probably good enough? @dysbulic  What do you think ?

dysbulic:
Yeah. So I think there should be a button that populates your profile form with the deprecated values, and you just submit it like normal.

dysbulic:
Especially since [self.id](https://self.id) has started breaking, so, other than our system, I don't know how one would make changes to their legacy Ceramic profile in any case.

PR #1493, opened 2023-01-09
ComposeDB Upgrade

### LINKED ISSUE #1491: Upgrade MyMeta to use the ComposeDB graphQL queries and mutations
- [ ] Frontend: Create a `useSaveProfileToComposeDB` hook, and have `EditProfileModal` use it
- [ ] Frontend: Add error handling to the save function and persist to our Hasura as a fallback (hopefully with minimal data transformation)
- [ ] Backend: Update the `updateSingle` cache to pull from ComposeDB instead

alalonde:
## Scope of PR

This PR changes the persistence of profile data to use Ceramic's ComposeDB.  We will use the PR deployment as a beta instance, pointing https://beta.metagame.wtf to it.  High level changes include:

- On the player setup and player profile pages, query from and persist to ComposeDB from the browser.
- Since we can now reliably and performantly (hopefully...) query from ComposeDB directly from the browser, we can get rid of the client-side state management library we have been using (Jotai).  This PR removes all usages.
- Selecting a profile picture is now part of the setup wizard (thanks @kon-rad)
- We have deployed a unified [ComposeDB model](https://github.com/MetaFam/TheGame/blob/733cf24feb916a1f73e9672ba417717cf0597f4c/packages/utils/schema/user-profile.graphql) for housing profile data
- There is a "Port your data to ComposeDB" modal that pops up on your own player page. Clicking this will load your profile from Hasura, create a new Ceramic stream with it, and save the returned stream ID to your `player` record in Hasura. This also occurs during the setup flow on any step that persists to ComposeDB.
- `useUser` queries ComposeDB, and doesn't return a player object until both the hasura and composeDB queries complete (if a `ceramicNodeId` is detected that is)

This resolves #1491, #1500, and #1492.

## Known Issues and Limitations

- It is currently not feasible to search profiles in ComposeDB, due to a lack of filtering and sorting capabilities.  This means that any part of the app that shows many players at once (e.g. the leaderboard) will not show data from ComposeDB, but from Hasura.  Player profile pages and the setup wizard DO query from ComposeDB. See the section below for potential workarounds for this.
- Deletion of profile fields is currently not possible. Setting a field to `null` is not supported, at least in version 0.3. This means that you cannot e.g. remove your background image, though you can replace it with another.
- When you change your avatar on the Edit Profile modal, it doesn't update on the MegaMenu. We need to move the PlayerHydrationContextProvider up to `_app.tsx` or otherwise communicate the update up to the root of the component hierarchy, not just the player page.

## Next steps

- [x] We should be able to leverage the current Profile caching mechanism on the backend still, but just read from ComposeDB rather than Ceramic tiles.  A daily job to update players would probably suffice, but ideally we also invalidate the cache on every mutation as well to keep data as up-to-date as possible. I think this is mandatory to finish before merging. 
- [x] Merge in the ENS support branch
- [x] Merge in Dysbulic's large omnibus fix branch

## Code improvements

- [ ] We could be doing a much better job of writing typesafe code around Hasura/ComposeDB profile objects.  
- [ ] I had the idea of creating a `class` for each profile field where each field class is responsible for its own conversion from hasura <-> composeDB, type handling etc using inheritance. I think this would be much cleaner, with all the logic for a particular field in one place
- [ ] We could probably ditch the `useGetOwnProfileFieldFromComposeDB` hook used in the setup flow pages in favor of a simple `useUser` call, since that now queries ComposeDB as well


packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
+type ProfileName
+  @createModel(accountRelation: SINGLE, description: "Your name") {
+  name: String! @string(maxLength: 150)
+}
+
+type ProfileAvatar
+  @createModel(
+    accountRelation: SINGLE
+    description: "An image and optional fallback images representing your avatar"
```
dysbulic, line 9:
```suggestion
    description: "A square image, and optional fallback images, representing your avatar."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  original: ImageMetadata!
+  alternatives: [ImageMetadata] @list(maxLength: 3)
```
dysbulic, line 12: The structure I would like to have is a set of `Dimension` descriptions with an associated list of `Files`.

The dimension description will include `width` & `height` and `filesize`. Also it will include an aspect ratio. Raster images would put in their width & height, and vectors would use `aspectRatio` to  represent they can be flawlessly resized to any dimensions.

```suggestion
  sizes: [Dimensions] @list(maxLength: 69420) → Files
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileDescription
+  @createModel(
+    accountRelation: SINGLE
+    description: "A short profile bio / blurb / description"
```
dysbulic, line 18:
```suggestion
    description: "A Markdown-based short profile bio / blurb / description."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  description: String! @string(maxLength: 420)
```
dysbulic, line 20: In general, I'd like to err on the side of extravagance when imposing size restrictions.

It's important to remember that we're dealing with text characters. The amount of space they take up is trivial compared to video or another heavyweight.

```suggestion
  description: String! @string(maxLength: 6942069)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileEmoji
+  @createModel(accountRelation: SINGLE, description: "Your spirit emoji") {
+  emoji: String! @string(maxLength: 2)
```
dysbulic, line 25: I think there are combining emoji sequences that consist of up to 5 characters.

Why not just let them specify a list of emoji. If the circumstances call for a single emoji we can just take the head of the list, or one at random.

I think the structure I like the most is an associative array where descriptors are linked to a string of emoji. Like "`happy`" linking to "`💆🏿‍♀️🤓👨🏿‍🌾🤪🧞‍♀️`".

Sites could pick a default descriptor to load the emoji set from, or allow users to choose.

Ultimately, I'd like to be supporting both emoji and vector images that can be used to represent the user accurately and concisely in a 2D interface.

```suggestion
  emoji: DescribedEmojiString @list(maxLength: 69420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileHeaderBackground
+  @createModel(
+    accountRelation: SINGLE
+    description: "An image (3:1 ratio) and optional fallback images to be used as a header background"
+  ) {
+  original: ImageMetadata!
+  alternatives: [ImageMetadata] @list(maxLength: 3)
```
dysbulic, line 34:
```suggestion
  sizes: [Dimensions] @list(maxLength: 69420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileURL
```
dysbulic, line 37:
```suggestion
type ProfileHomepage
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  @createModel(
+    accountRelation: SINGLE
+    description: "Your home on the internet"
```
dysbulic, line 40:
```suggestion
    description: "Your home on the internet."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  url: String! @string(maxLength: 240)
```
dysbulic, line 42:
```suggestion
  url: String! @string(maxLength: 420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileHomeLocation
+  @createModel(accountRelation: SINGLE, description: "Where you live") {
```
dysbulic, line 46:
```suggestion
  @createModel(accountRelation: SINGLE, description: "Where you live.") {
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  homeLocation: String! @string(maxLength: 140)
```
dysbulic, line 47: Does the `!` mean it can't be set to `null`?

I think we should reserve the ability to null-out a user's profile if they request it. Also the ability to request we unpin it.

The right to be forgotten is a really big deal, so if someone wants to burn an identity, they should have that capacity.

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileUsername
+  @createModel(
+    accountRelation: SINGLE
+    description: "A representation of your identity that can be used in an interface, typically as part of a URL"
```
dysbulic, line 53:
```suggestion
    description: "A representation of your identity that can be used in an interface, typically as part of a URL."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  # pattern: "^[a-z0-9-_]+$"
```
dysbulic, line 55: We'll still need to have pretty strict requirements for how this value is structured.

I like the idea of allowing capital letters, but doing the uniqueness check based on a purely lowercase string.

I think the character set should be restricted to whatever is permissible as part of a subdomain name. Limited to a single position though, so no `.`s.

If the name submitted contains a dot, it is derefenced using ENS.

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  username: String! @string(maxLength: 383)
```
dysbulic, line 56: I don't know what the limits on DNS name sizes are, but I'm pretty sure they exist and aren't huge.

```suggestion
  username: String! @string(maxLength: 128)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileTimeZone
+  @createModel(
+    accountRelation: SINGLE
+    description: "The time zone you are typically in. This can be represented in any many ways; e.g. `GMT-8:00`, `Pacific Time`, `PST`"
```
dysbulic, line 62:
```suggestion
    description: "The time zone you are typically in. This can be represented in any many ways; e.g. `GMT-8:00`, `Pacific Time`, `PST`, `U.S.A./New York`."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  timeZone: String! @string(maxLength: 127)
```
dysbulic, line 64:
```suggestion
  timeZone: String! @string(maxLength: 420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileSquareBackground
+  @createModel(
+    accountRelation: SINGLE
+    description: "An image (1:1 ratio) and optional fallback images to be used as a page background"
+  ) {
+  original: ImageMetadata!
+  alternatives: [ImageMetadata] @list(maxLength: 3)
```
dysbulic, line 73:
```suggestion
  sizes: [Image] @list(maxLength: 69420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+
+type ProfileTimeZone
+  @createModel(
+    accountRelation: SINGLE
+    description: "The time zone you are typically in. This can be represented in any many ways; e.g. `GMT-8:00`, `Pacific Time`, `PST`"
+  ) {
+  timeZone: String! @string(maxLength: 127)
+}
+
+type ProfileSquareBackground
```
dysbulic, line 67:
```suggestion
type ProfileBackground
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  @createModel(
+    accountRelation: SINGLE
+    description: "An image (1:1 ratio) and optional fallback images to be used as a page background"
```
dysbulic, line 70:
```suggestion
    description: "A square image (1:1 ratio) to be used as a webpage background."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  original: ImageMetadata!
+  alternatives: [ImageMetadata] @list(maxLength: 3)
+}
+
+type ProfileAvailability
+  @createModel(
+    accountRelation: SINGLE
+    description: "Some data indicating how available you are to take on additional work"
```
dysbulic, line 79:
```suggestion
    description: "Number of hours per week you have to take on additional work."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  weeklyHours: Int! @int(min: 0, max: 168)
+}
+
+type ProfileGenderIdentity
+  @createModel(
+    accountRelation: SINGLE
+    description: "Some data indicating your gender identity"
```
dysbulic, line 87:
```suggestion
    description: "Data indicating your gender identity."
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+  pronouns: String @string(maxLength: 255)
```
dysbulic, line 89:
```suggestion
  pronouns: String @string(maxLength: 420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  gender: String @string(maxLength: 42)
```
dysbulic, line 90:
```suggestion
  gender: String @string(maxLength: 420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type ProfileDisposition
+  @createModel(
+    accountRelation: SINGLE
+    description: "Some information about your personality or psychometric makeup. "
+  ) {
+  # pattern: "^([wW]?|[uU]?|[bB]?|[rR]?|[gG]?){0,5}$"
+  fiveColorDisposition: String @string(minLength: 1, maxLength: 5)
```
dysbulic, line 99:
```suggestion
  mtgColorDisposition: String @string(minLength: 1, maxLength: 5)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+  # pattern: "^(Pioneer|Settler|Town Planner)$"
+  explorerType: String @string(maxLength: 12)
+}
```
dysbulic, line 102:
```suggestion
  mbti: String @string(maxLength: 4)
  enneagram: String @string(maxLength: 12)
}
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
...
+
+type ImageMetadata {
+  # pattern: "^ipfs:(//)?.+$"
+  ipfsUrl: String! @string(maxLength: 150)
+  mimeType: String! @string(maxLength: 50)
+  width: Int! @int(min: 1)
+  height: Int! @int(min: 1)
+  size: Int @int(min: 1)
+}
```
dysbulic, line 111:
```suggestion
type Image {
  height: Int @int(min: 0)
  width: Int @int(min: 0)
  aspectRatio: Int @int(min: 0, max: 1)
  size: Int @int(min: 0)
  url: String! @string(maxLength: 420)
  mimeType: String! @string(maxLength: 420)
}
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
+type ProfileName
+  @createModel(accountRelation: SINGLE, description: "Your name") {
+  name: String! @string(maxLength: 150)
```
dysbulic, line 3:
```suggestion
  name: String! @string(maxLength: 69420)
```

packages/utils/schema/user-profile.graphql
```diff
@@ -0,0 +1,111 @@
+type ProfileName
+  @createModel(accountRelation: SINGLE, description: "Your name") {
```
dysbulic, line 2:
```suggestion
  @createModel(accountRelation: SINGLE, description: "Your name.") {
```

dysbulic (commented):
This is only a first pass. I don't know GraphQL's schema language, so I could only create structures based on your examples. I'd like to spend some time with the standard & have another pass.

3avi:
sorry that we don't support `pattern` yet - i captured it into our backlog and having a conversation with one of our engineers tomorrow about where we can squeeze it in. appreciate y'all figuring it out!

packages/utils/schema/user-profile.graphql
alalonde: I changed this to move away from the trademarked `mtg`

packages/utils/schema/user-profile.graphql
alalonde: why 420, generally-speaking?

packages/utils/schema/user-profile.graphql
alalonde: The description is for the entire model, not the field. I can easily see adding more fields to this model

packages/utils/schema/user-profile.graphql
alalonde: Yes, that's right. I was thinking that applications would just remove the model entirely, but maybe that's insufficient

packages/utils/schema/user-profile.graphql
alalonde: Patterns are unfortunately not supported yet, but they're on the Ceramic team's roadmp

packages/utils/schema/user-profile.graphql
alalonde: I don't think we want to make this max smaller since we will porting over old data that has a higher max

packages/utils/schema/user-profile.graphql
alalonde: This schema was just taken from the original basic profile data model. I think it's a bit more ergonomic than your proposal, as one can just ignore the alternative images if they don't care about them

packages/utils/schema/user-profile.graphql
alalonde: This is an interesting idea but IMO needlessly complicated. At least we should support the basic `string` representation.  Models should be as self-explanatory as possible if we want them to be used.

alalonde: We can always add this as an additional field here

dysbulic:
I opened the test instance in Chrome with MetaMask *(I can't authenticate with Frame)*, and I'm seeing the following in the profile editing pane:

Screenshot from 2023-07-12 11-40-23

dysbulic (approved):
Sure. Let's do this!

mzkrasner:
@alalonde I know filtering and sorting was called out as a ComposeDB limitation at the time this issue was authored - this has since been resolved with the feature release of filtering and ordering (described in this [blog post](https://blog.ceramic.network/tutorial-query-filtering-and-ordering-in-composedb/)). 

Do you know if there are eventual plans to upgrade your ceramic/composeDB packages and add indexes to the ComposeDB schema you're using?

Issue #1176, opened 2022-03-01
[MyMeta] ENS usernames


peth-yursick:
### What would you like to be added?

The ability for people to set their ENS address as their username.

### Why is this needed?

Given its popularity in the space, its kind of embarrassing not to allow people to set their username as so.
The problem seems to be at the ceramic level? You're simply not allows to have . in your username? @dys

alalonde:
What if we just added a "Use ENS" checkbox as an alternative to username? Since we already have their ETH address at that point.  I'm not sure that we want to store ENS names in our DB (I believe they can expire, be re-assigned, etc).

We could support ENS out of the box already by just looking up ENS names from players that have no "username" (name in the profile table I believe)

dysbulic:
I agree with not storing the names. Not only might they be transfered, but we'd also have to worry about people registering someone else's name with ill intent.

I like the idea of disallowing `.eth` as the terminal part of a username, resolving the ENS name of any profiles requested with a `.eth` extension, then serving up the profile of the associated address.

It's very similar to the situation with addresses. Usernames of the same format as an address (`/^0x[a-f\d]{40}$/i`) aren't allowed, and if a profile is requested using one, it is looked up using a method specifically for addresses.

I don't think it's necessary to ask them to do it.

Seroxdesign:
Hey, I found this task on Dework, I've worked on adding ENS support for multiple DAOs and I can see this task through if it's still available. @alalonde

alalonde:
@dysbulic I'd like to make this task's requirements much more explicit so that new devs can start working on it.  How does this sound?

- Update the backend regex filter to allow `.eth` as a suffix, but otherwise still disallow `.` in a username
- When a user types in a `.eth` username, query ENS from the frontend to see if the username is valid?

Outstanding questions:
- I assume we want the user to go through some validation process as part of the ENS query to ensure that the username is in fact theirs.
- We would still need to do some verification on the backend, since the hasura graphQL endpoint allows a user to update their own username. So do we only allow a user to use the ENS name associated with the eth address linked to their profile? And if that's the case, why not just pull in a linked ENS name automatically when they set their ETH address?

Seroxdesign:
### Milestones: 

- [ ]  Allow user to update their username with an ENS by updating RegEx
- [ ]  Resolve ENS from connected user wallet and validate input ENS is owned by user
- [ ]  If user's connected wallet does not resolve an ENS disallow them from saving/updating username's with dedected . chars
[...]

PR #1509, opened 2023-01-23
add ENS set up to web3State

Seroxdesign:
## Overview

Adding ENS support

https://github.com/MetaFam/TheGame/issues/1176

Closes #

## Implementation

**Side effects**

## Assets

### DIFF
--- a/packages/web/contexts/Web3Context.tsx
+++ b/packages/web/contexts/Web3Context.tsx
@@ -9,6 +9,7 @@ import { CONFIG } from 'config';
 import { DIDSession } from 'did-session';
 import { DID } from 'dids';
 import { getResolver as getKeyResolver } from 'key-did-resolver';
+import ethers from 'ethers';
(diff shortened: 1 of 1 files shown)


dysbulic:
So, three things:

1. Could you push your branch to the `MetaFam/TheGame` repo and make a PR from that? The automated test setup doesn't work for branches from remote repositories.
2. Variable names are camel-cased, with all elements of an acronym being the same case, so the variable would be named `ens`.
3. There is still ongoing discussion as to how this feature should be added. If we go [my route](https://github.com/MetaFam/TheGame/issues/1176#issuecomment-1399714113) you won't resolve the reverse address, rather you'll change the routing code for `/player/<name>` and if `name` contains a `.` you'll see if there's an address associated with it in ENS & load that profile.

Seroxdesign:
Closed: Understood, will follow your recommendations

PR #1513, opened 2023-01-25
ENS support set up - WIP

Seroxdesign:
## Overview

Add name to address ens resolver

https://github.com/MetaFam/TheGame/issues/1176

Closes #

## Implementation

**Set up resolver to not rely on user connection, when user inputs ens to player search, resolve the address, set filter params**

**using useEff and async, would be good to get someone's view on this, or if it needs to change**

## Assets

### DIFF
--- a/packages/web/components/Player/Filter/PlayerFilter.tsx
+++ b/packages/web/components/Player/Filter/PlayerFilter.tsx
@@ -273,7 +273,7 @@ export const PlayerFilter: React.FC<Props> = ({
                       setQueryVariable('search', `%%`);
                     }
                   }}
-                  placeholder="SEARCH NAME OR ETH ADDRESS"
+                  placeholder="SEARCH ENS, NAME, OR ETH ADDRESS"
                   _placeholder={{ color: 'whiteAlpha.500' }}
                   size="md"
                   type="text"
--- a/packages/web/components/Player/Filter/PlayerFilter.tsx
+++ b/packages/web/components/Player/Filter/PlayerFilter.tsx
@@ -32,7 +32,7 @@ import {
 } from 'lib/hooks/players';
 import { useIsSticky } from 'lib/hooks/useIsSticky';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
-import { getAddressFromName } from 'utils/ensHelpers';
+import { getAddressForENS } from 'utils/ensHelpers';
 import { SkillOption } from 'utils/skillHelpers';
 
 const Form = chakra.form;
@@ -71,7 +71,7 @@ export const PlayerFilter: React.FC<Props> = ({
   useEffect(() => {
     const resolveName = async () => {
       if (search.includes('.')) {
-        const address = await getAddressFromName(search);
+        const address = await getAddressForENS(search);
         setQueryVariable('search', address);
       }
     };
--- a/packages/web/pages/player/[username].tsx
+++ b/packages/web/pages/player/[username].tsx
@@ -242,7 +242,7 @@ export const getStaticProps = async (
   // If username in url includes a . attempt to resolve ENS
   if (username.includes('.')) {
     const address = await getAddressForENS(username);
-    user.address = address || username;
+    user.address = address?.toLowerCase() || username;
     user.ens = username;
   }
   if (ethers.utils.isAddress(username.toLowerCase())) {
--- a/packages/web/pages/player/[username].tsx
+++ b/packages/web/pages/player/[username].tsx
@@ -7,7 +7,6 @@ import {
   DEFAULT_PLAYER_LAYOUT_DATA,
 } from 'components/Player/Section/config';
 import { HeadComponent } from 'components/Seo';
-import { ethers } from 'ethers';
 import {
   Player,
   useInsertCacheInvalidationMutation as useInvalidateCache,
@@ -26,8 +25,9 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import useSWR from 'swr';
 import { LayoutData } from 'utils/boxTypes';
-import { getAddressForENS, getENSForAddress } from 'utils/ensHelpers';
+import { getENSForAddress, getPlayerData } from 'utils/ensHelpers';
 import {
   getPlayerBackgroundFull,
   getPlayerBannerFull,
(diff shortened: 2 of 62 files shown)


packages/web/components/Player/Filter/PlayerFilter.tsx
```diff
@@ -67,6 +68,18 @@ export const PlayerFilter: React.FC<Props> = ({
     sortOptionsMap[SortOption.SEASON_XP],
   );
 
+  // Get ADDRESS from ENS, temporary.
+
+  useEffect(() => {
+    const resolveName = async () => {
+      if (search.includes('.')) {
+        const address = await getAddressFromName(search);
+        setQueryVariable('search', `%${address!}%`);
```
dysbulic, line 77: Since the address should match exactly, I don't think you need the wildcards.

```suggestion
        if (address) {
          setQueryVariable('search', address);
        }
```

packages/web/utils/ensHelpers.ts
```diff
@@ -0,0 +1,10 @@
+import { ethers } from 'ethers';
+
+export const getAddressFromName = async (ens: string) => {
+  // getting a provider without user connecting, todo: use whatever rpc metagame wants
+  const mainnetProvider = new ethers.providers.JsonRpcProvider(
+    'https://eth.llamarpc.com',
```
dysbulic, line 6:
```suggestion
    CONFIG.mainnetRPCURL,
```

packages/web/components/Player/Filter/PlayerFilter.tsx
dysbulic: In fact, now that I think about it, including the wildcards on an address would allow an attacker to potentially create a name that is a superstring of an ETH address and have it inappropriately match.

packages/web/utils/ensHelpers.ts
Seroxdesign: Do we already have a mainnetRPC to use, or should I make a new one?

dysbulic: There is not one currently, you'll need to add it. It should pull from an environment variable `NEXT_PUBLIC_MAINNET_RPC` *(I think. Is is accessed from Next?)*.

packages/web/utils/ensHelpers.ts
```diff
@@ -0,0 +1,17 @@
+import { ethers } from 'ethers';
+
+const mainnetProvider = new ethers.providers.JsonRpcProvider(
+  'https://eth.llamarpc.com',
+);
+
+export const getAddressFromName = async (ens: string) => {
+  const address = await mainnetProvider.resolveName(ens);
+  const data = address?.length === 42 ? address : ens; 
+  return data;
```
dysbulic, line 10:
```suggestion
  return address?.length === 42 ? address : ens; 
```

packages/web/utils/ensHelpers.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const getNameFromAddress = async (address: string) => {
+  const name = await mainnetProvider.lookupAddress(address);
+  const data  = name ? name : address;
+  return data;
```
dysbulic, line 16:
```suggestion
  return name ?? address;
```

dysbulic (commented):
Couple of places you can save a line & a few characters.

packages/web/utils/ensHelpers.ts
Seroxdesign: Followed through with this, in latest commit

packages/web/utils/ensHelpers.ts
Seroxdesign: Followed through with this, in latest commit

Seroxdesign:
> I reckon that you've made changes to your code that resolve some of these issues, but haven't pushed them yet. You shouldn't resolve the conversations until you've uploaded how you resolved the conversation.

I commented changes, I changed the getAddressFromName function to always return a string, either the ENS, or address depending on whether or not name is resolved.

Seroxdesign:
## Remaining work:

------------------------

- [ ] Get ENS to work in /me
- [ ] Get ENS working in tiles and leaderboard
- [ ] only  get ens URL || ens Name when no username

------------------------

Clean up

packages/web/components/Player/Filter/PlayerFilter.tsx
Seroxdesign: I resolved this by following your suggestion, removed the *%

packages/web/utils/ensHelpers.ts
Seroxdesign: Created a new env for NEXT_PUBLIC_MAINNET_RPC and added in config.

dysbulic:
Are you able to run `yarn build` in `packages/web/`? The [build process for the test instance](//github.com/MetaFam/TheGame/actions/runs/4078160716/jobs/7028129578) failed completely.

Seroxdesign:
> Are you able to run `yarn build` in `packages/web/`? The [build process for the test instance](//github.com/MetaFam/TheGame/actions/runs/4078160716/jobs/7028129578) failed completely.

I've been trying to get build to run in my cli, but I keep getting webpack errors and type/import errors that I'm not sure where' they're coming from:

Example:

```
./components/Carousel/Item.tsx:1:22
Type error: Could not find a declaration file for module '@metafam/ds'. '/home/sero/TheGame/packages/design-system/dist/ds.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/metafam__ds` if it exists or add a new declaration (.d.ts) file containing `declare module '@metafam/ds';`

> 1 | import { Flex } from '@metafam/ds';
    |                      ^
  2 | import type { PropsWithChildren } from 'react';
  3 | import React, { KeyboardEventHandler, useState } from 'react';
  4 |
```

In the github actions build I get errors regarding the username and guild pages, I didn't work on the guild pages.

dysbulic:
> Type error: Could not find a declaration file for module '@metafam/ds'.

That's my bad. I should have had you run `yarn build` in the root *(with the backend running)*, since the dependencies have to be built before the site.

> In the github actions build I get errors regarding the username and guild pages, I didn't work on the guild pages.

There's two things going on if your most recent log. One is repeated reports of "Database not available" which is happening because [The Graph](//thegraph.com) is on the fritz and refusing connections. Rerunning the job will eventually get by that.

For the user pages, however, you're getting an error like: `Error: invalid address (argument="address", value="misanth", code=INVALID_ARGUMENT, version=address/5.7.0)`. So you've caused a method that is supposed to return an address to return the username instead.

Seroxdesign:
## Ready for review

Features added:

 - 1. Resolve ENS to address in URL, this will allow users to navigate to a URL ex: https://metagame.wtf/player/serox.eth, the ens will retrieve address then use that to retrieve my profile

 - 2. Display name based on hierarchy, example: when navigating to a profile, the following will be displayed in name component: 
1. Username if available, 2. ENS name if available, 3. Address, if neither of the above are available

 - 3. Generate URLs based on same heirarchy as above.

### Build errors: 

On my local machine build succeeds with no issue, all pages are generated and things work as expected

successful-build-local-machine

In the Github Actions check, build fails because some address pages cannot be generated, I took those addresses and tried to test them in my local version and things work fine

Example: 
image
image

Why this is happening, I'm not sure. Perhaps someone can take a look and let me know, 

@dysbulic

packages/web/pages/player/[username].tsx
```diff
@@ -199,13 +230,24 @@ export const getStaticProps = async (
     };
   }
 
-  const player = await getPlayer(username);
+  // If username in url includes a . attempt to resolve ENS
+  if (username.includes('.')) {
+    user.address = await getAddressFromName(username);
+    user.ens = username;
+  }
+  if (username.length === 42) {
```
Comment (alalonde) [inline packages/web/pages/player/[username].tsx:238]: There is a regex in the codebase you should probably be using instead

packages/web/utils/ensHelpers.ts
```diff
@@ -0,0 +1,16 @@
+//import { CONFIG } from 'config';
+import { ethers } from 'ethers';
+//const { mainnetRPC } = CONFIG;
+
+//replace this with the mainnetRPC from above after testing and adding env, uncomment both mainnetRPC and CONFIG
+const mainnetProvider = new ethers.providers.JsonRpcProvider('https://ethereum.publicnode.com');
```
alalonde, line 6: If you search the project for other frontend variables, you should be able to figure out how to get this to work. You may need to add it to the Dockerfile and/or github action template

alalonde (changes requested):
Definitely need to be restricting `.` in usernames, validated:
1. frontend in the setup name page
2. frontend in the edit profile modal
3. backend

dysbulic:
If you click on the first link on the leaderboard it goes to `metagame.wtf/player/dys.metagame.wtf`, not `frontend-pr-1513-mjhnbmqqna-uk.a.run.app/players` like it should. You've made it an absolute URL when it should be relative.

If you click on the third person on the leaderboard, `musashi`, the link is `frontend-pr-1513-mjhnbmqqna-uk.a.run.app/player/musashi13` correctly, but it 404s.

packages/web/utils/ensHelpers.ts
dysbulic: If you look in `packages/web/config.ts`, it should be apparent how to specify a default value for a configuration value.

Seroxdesign:
Notes on what to change:

Also, all usernames which have an associated player record should resolve. Every player has to have an ETH address in order to create their account, so you should always be able to load the profile for a username.

I thought you'd mentioned just adding ENS support to getPlayerURL. Your code doing substring comparisons to figure out what to render is both difficult to read, prone to break, and about 5⨯ longer than it needs to be.

You could change getPlayerURL to:

export const getPlayerURL = async (
  player?: Maybe<Player | GuildPlayer>,
  opts: { rel?: boolean; default?: string } = {},
): string | undefined => {
  let { username } = player?.profile ?? {};
  username ??= await getENSForAddress(player?.ethereumAddress);
  username ??= player?.ethereumAddress;
  const { rel: relative = true } = opts;
  if (username) {
    const path = `/player/${username}`;
    return `${relative ? '' : 'https://my.metagame.wtf'}${path}`;
  }
  return opts.default;
};


You'd need to stop hiding errors in getENSForAddress (you called it getNameFromAddress), but that would be a good idea in general. Almost always when a name or address fails to resolve the calling code should handle that situation rather than trying to mask it like you're doing

packages/web/utils/ensHelpers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const getAddressFromName = async (ens: string) => {
+  const address = await mainnetProvider.resolveName(ens);
+  return address?.length === 42 ? address : ens;
+};
+
+export const getNameFromAddress = async (address: string) => {
+  const name = await mainnetProvider.lookupAddress(address);
+  return name || address;
+};
```
dysbulic, line 16: Again, you're hiding the failure to resolve. Why?

dysbulic (changes requested):
@Seroxdesign, you're prepping to submit a bunch of changes, so I'll wait for those before doing the whole review.

Issue #1409, opened 2022-10-13
[MetaOS] Members-only features


peth-yursick:
### What would you like to be added?

Thought it would make sense to start discussing the gating & members-only features now that we are starting to have some that could be defined as such.

Here's two we could start with:
- Limiting the player list to top 56, indicate total number of players (#871) & total number of profiles. Gate the use of search & filtering. Afaik no freelancing or social media platform allows you to use search without registering. As we onboarding more players & guilds who will onboard their players, search & filtering becomes more useful.
- We already have custom backgrounds enabled with #443, I think this is a perfect example of a feature that could be members-only as its not crucial for the use of profiles.

Other ideas:
- Custom roles
- Custom skills
- More than 2 custom blocks on the profile
- Champions Path questchain
- High value quests
- Bridgebuilding raid board

### Why is this needed?

This would help convert people into players or patrons, would also help us get more patrons in general as it would be easier to answer the question of making money as a project.

Seroxdesign:
This will be included with the SEED issue and turned into a project in Github

PR #1517, opened 2023-02-03
Limiting background setting to players & patrons *(+ a shitton of bug fixes)*

### LINKED ISSUE #1502: Figure out where my profile goes when sorting by seasonal XP & why the account that receives SEED isn't getting credited XP in MyMeta
### What happened?

When the [leaderboard](https://metagame.wtf/players/) is sorted by Seasonal XP, none of my several accounts on the system are displayed.

Under All Time XP sorting, my original ETH account *(from two compromised seeds ago)* is that shows with 5,374XP while the Discord CredBot reports I have 7,473.

My active [MetaGame-related address](https://metagame.wtf/player/0xeb4e3e9fa819e69e5df4ea35b9c7973062c96de9) should display in the leaderboard with the correct XP.

Look for me in the leaderboard.

dysbulic:
## What features/fixes does this PR include?

First off, this PR fixes a few backend bugs:

* In the action to sync the SourceCred ledger, if a user changed their ETH address, the system would attempt to update their new address with their XP and Discord id. Setting the Discord id would violate a uniquness constraint on the field, throw an error, and skip the rest of the updates. I changed the `UpdatePlayer` mutation to be multipart and clear the Discord id if set on a different account.
* The action to sync guild membership from Discord role holders would throw an error on an unrecognized guild id, and that would cause all subsequent updates to be skipped. I switched the awaiting function from `Promise.all` to `Promise.allSettled` which will attempt to update them all.
* I added descriptive messages to most of the HTTP error returns.
* The Express `errorMiddleware` handler only took three arguments. Express recognizes an error handler by the fact it takes four arguments, so this was getting incorrectly called on every request and only because the arguments were mismatched was it not causing more problems. I added the fourth argument to fix it.
* The Meet With Wallet integration had a really awkward check whether to update the field that also switched the check to an `async` method that wasn't `await`ed creating a race condition that kept some values from being updated correctly. I refactored the `async` out and simplified the check.
* When syncing the SourceCred data, I clear all the existing data before each load. This will properly handle season starts, rank changes as well as updates to ETH addresses. Removing the `where` condition will also likely cause the statement to execute more quickly since it eliminates `n` comparisons.
* For unknown reasons, likely network changes, the Ceramic cache in Hasura stopped updating. I noticed a format difference in the models configuration file, removed the `ceramic://` from one set of values, and it started working again.
* To try and deal with the load times, I converted all PNG images to WebP, generally saving about 80% on the file size.
* DAOHaus's Graph changed & we're getting a duplicate entry error from the product of `codegen`. I added an `awk` script to the `generate` `script` to strip the duplicate line.
* Fixes an issue with a mismatch between the value passed to `onChange` by `react-timezone-select` & what is expected by the `react-hook-form` controlling the input that was preventing the field from saving.
* Removes an inappropriate constraint from the `name` entered in the profile wizard. It was treated like a username, but saved as the display name.
* The test instance didn't have the GraphQL endpoint set, so I switched the syntax for setting environment variables in the GitHub Action from a deprecated format scheduled to be sunset May 31 to a new one.

The primary point was to limit the ability to display a background image. This PR:

* Only uses the background image if the user is a MetaGame player or patron.
* Adds the text "Only available to MetaGame players and patrons." to the description for the background image.
* Also, creates Hasura actions for the SourceCred and Discord sync tasks so they can be triggered at will.

## GitHub issue number

Closes #1502, #1527, #1528, #1409, & #1532.

## Follow Up Improvement Ideas

Currently the fact that background images only work for MetaGamers is only present in the information popup for the field that comes up when you hover over the field name:

Screenshot from 2023-02-03 12-02-34

This could be made more visible in the redesign that's accompanying the ComposeDB migration.

---

There is an outstanding hydration error because the server-rendered content doesn't match what the client produced on the landing and profile details pages.

---

I removed the cropping of the player description on the profile details page and render it as Markdown. The player's colors should likely be moved to the adjacent column, so that the hero is the only thing in the first column by default.


alalonde:
[Here](https://github.com/MetaFam/TheGame/actions/runs/4213173027/jobs/7312731826) you can see that the seed-db script is expecting `ACCOUNT_MIGRATION_URL` but you are setting `ACCOUNT_SYNC_URL`

.github/workflows/gcp-deploy.yaml
```diff
@@ -80,62 +79,62 @@ jobs:
       HASURA_PORT: ${{env.HASURA_PORT}}
       FRONTEND_PORT: ${{env.FRONTEND_PORT}}
       HASURA_SECRET: ${{env.HASURA_SECRET}}
```
alalonde, line 81: Can you update `gcp-deploy-frontend.yaml` as well please?

packages/backend/package.json
```diff
@@ -10,10 +10,10 @@
   "scripts": {
     "start": "node --trace-warnings dist/index.js",
     "build": "yarn generate && tsc -b",
-    "dev": "tsc --build && concurrently \"tsc --watch --preserveWatchOutput\" \"nodemon\" \"yarn generate --watch\"",
+    "dev": "tsc --build && concurrently 'tsc --watch --preserveWatchOutput' nodemon 'yarn generate # --watch'",
     "typecheck": "yarn build",
     "precommit": "yarn lint-staged",
-    "generate": "DEBUG=1 graphql-codegen --config=codegen.yml",
+    "generate": "DEBUG=1 graphql-codegen --config=codegen.yml && export OUT=src/lib/autogen/daohaus-sdk.ts && awk '!/MolochVersion = .molochVersion/' $OUT > $OUT.filtered && mv $OUT.filtered $OUT",
```
alalonde, line 16: Is the proper fix to migrate to the newer daohaus graph?

packages/backend/src/lib/apiHelpers.ts
```diff
@@ -47,7 +47,11 @@ export const errorMiddleware: ErrorRequestHandler = (
   error: Error,
   _req: Request,
   res: Response,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  next: NextFunction, // 4 args required to id this as an error handler
```
alalonde, line 51: Thanks for the comment here..

packages/web/components/EditProfileModal.tsx
```diff
@@ -545,13 +552,21 @@ export const EditProfileModal: React.FC<ProfileEditorProps> = ({
                 key: 'backgroundImageURL',
                 title: 'Page Background',
                 description:
-                  'An image with an ~1:1 aspect ratio to be the page background. 1MiB maximum size.',
+                  // eslint-disable-next-line prefer-template
+                  'An image with an ~1:1 aspect ratio to be the page background. 1MiB maximum size.' +
+                  (metagamer
+                    ? ''
+                    : '\nOnly available to MetaGame players and patrons.'),
```
alalonde, line 559: I don't think restricting features will really serve us at this point in the product's lifespan, but I suppose the bar to become a player/patron is pretty low...

alalonde (changes requested):

packages/backend/package.json
alalonde: Just a question, we don't have to upgrade now

.github/workflows/gcp-deploy.yaml
dysbulic: Done. To the best of my ability. We won't know 100% that it's correct until it deploys `develop`.

packages/backend/package.json
dysbulic: Yeah. I did the `awk` script before DAOHaus got back to us.

There's a ticket, #1529, to replace the endpoint.

packages/backend/src/lib/apiHelpers.ts
dysbulic: NP.

packages/web/components/EditProfileModal.tsx
dysbulic: I agree.

What I'd personally like to see is an easy to set up easy to theme SourceCred leaderboard with profiles that organizations purchase a subscription for.

In exchange for their money, they don't have to run a Ceramic node or SourceCred.

What I think will be crucial for this is a switch to Credequate & the addition of a second order system that allows people to adjust the algorithmically determined weights.

So far as MetaOS, an operating system really needs flexible and powerful extensibility — like installing programs in a desktop OS, or apps in a mobile OS. Designing and implementing such a system is a huge engineering feat. Maybe someone is going to come up with an elegant and simple answer, but I don't think that what we have is it, nor do I think we have a groundbreaking vision for it.

Simply the fact that all new modules have to be merged into our codebase is a significantly limiting factor both from a security, but also autonomy perspective. To have a real ecosystem, you can't have such a stricture on the publishing lifecycle.

I still say we let MetaOS go for the time being and focus on making it as easy to onboard into SourceCred as filling out a web form with the sources you want to collect and paying a little ether. Add on a program for adjusting the fairness, and I think you have something engaging that people would spend time digging through and playing with.

alalonde (approved):

packages/web/components/Player/Filter/AdjacentTimezonePlayers.tsx
```diff
@@ -1,4 +1,9 @@
-import { Flex, Skeleton, Text, TimeZoneOptions, VStack } from '@metafam/ds';
+import {
+  Flex,
+  Skeleton,
+  Text,
+  /* TimeZoneOptions, */ VStack,
+} from '@metafam/ds';
```
alalonde, line 6: Why are there two of these files with slightly different names? Can we consolidate?

packages/web/utils/dateHelpers.ts
```diff
@@ -23,7 +23,7 @@ export const getPlayerTimeZoneDisplay = (
 ): TimeZoneDisplay => {
   let tzLabel;
   let offsetLabel;
-  const timeZone = TimeZoneOptions.find((t) => t.name === zone);
+  const timeZone = null; // TimeZoneOptions.find((t) => t.name === zone);
```
alalonde, line 26: Do all the timezone selectors work again? I assume we want to reinstate react-timezone since it didn't fix the FOUC?

alalonde (changes requested):

alalonde (approved):

PR #1520, opened 2023-02-09
Playbooks x Quest Chains integration

vidvidvid:


alalonde:
Looks like there are some build errors here @vidvidvid

vidvidvid:
@alalonde yeah, sorry this should be a draft for now

vidvidvid:
The PR is ready. There is, however, a bug which is messing up the css of the MarkdownEditor and looks like this: 
Screenshot 2023-04-03 at 23 02 38

It looks like there might be a problem in importing the css files in `web/pages/_app.tsx`:
```
import 'react-markdown-editor-lite/lib/index.css';
import 'assets/custom-markdown-editor.scss';
```
The classes from these two files aren't showing up in the console, so I'm assuming it has to do something with this. I don't know how to solve this problem, nor how it got messed up in the first place. On master branch everything works fine.
Help, please

packages/web/utils/questChains.ts
```diff
@@ -193,6 +231,144 @@ export const QuestChainGreatHousesDetails: Record<
   },
 };
 
+export const QuestChainPlaybooksDetails: Record<
+  QuestChainType,
+  QuestChainInfo
+> = {
+  [QuestChainsPlaybooks.MEME_DRIVEN_ORGANIZATIONS]: {
```
alalonde, line 238: Seems like these playbooks should be in a database table... Do you have a good reason why they shouldn't be?  I assume we will want to be modfiying them fairly often, which shouldn't require mucking in the code

vidvidvid: I can't set up the hasura console on my mac, been at it for hours and then I gave up. Also it's convenient and simpler to change the config if it's in the client.

alalonde: OK. Definitely don't agree with this 

> Also it's convenient and simpler to change the config if it's in the client.

when you can just update the data in Hasura and have it reflect in any environment in a matter of seconds...

alalonde: @dysbulic Any thoughts here? Should we get this stuff in the DB or merge as-is

vidvidvid: > when you can just update the data in Hasura and have it reflect in any environment in a matter of seconds...

don't you have to write a migration?

vidvidvid: Anyways, I can't even spin up a hasura console.

PR #1523, opened 2023-02-13
Add profile photo as a step in the profile wizard

### LINKED ISSUE #1230: [MyMeta] Profile photo as one of the steps in the creation flow

I think adding the profile picture is important enough to have it in the main profile creation flow together with "name" & "description".

### Why is this needed?

Its a fairly basic piece of any profile.

kon-rad:
## Overview

Adds upload profile image step to the profile wizard. 

Closes #1230

## Implementation

Uploads image to web3.storage, then hash into ceramic, when 'next' button is clicked. 

Designs:
https://www.figma.com/file/eeJgWe6KjuSnRiOF47D5wH/Add-profile-image?node-id=2%3A957&t=1ls2xq43BkcQGiOB-0

**Side effects**

## Assets

Screen Shot 2023-02-14 at 11 04 26 AM
Screen Shot 2023-02-14 at 11 04 18 AM

### DIFF
--- a/packages/web/components/Setup/SetupProfilePicture.tsx
+++ b/packages/web/components/Setup/SetupProfilePicture.tsx
@@ -1,0 +1,122 @@
+import { LogLevel } from '@ceramicnetwork/common';
+import { VStack, Button, Center, Image, Input, useToast } from '@metafam/ds';
+import OctoAvatar from 'assets/graphics/octo-avatar.svg';
+import React, { useState, useRef } from 'react';
+import { ProfileWizardPane } from './ProfileWizardPane';
+import { WizardPaneCallbackProps } from './WizardPane';
+// import { Controller, useForm } from 'react-hook-form';
+// import { optimizedImage } from 'utils/imageHelpers';
+
+// const ERROR_MESSAGE = 'Has to be a .jpg, .png, or .gif file; under 1MB.';
+
+export const SetupProfilePicture: React.FC = () => {
+  const field = 'profileImageURL';
+  const [image, setImage] = useState<any>({});
+  const [imageURL, setImageURL] = useState<any>({});
+  const [file, setFile] = useState<any>(null);
+  const toast = useToast();
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  return (
+    <ProfileWizardPane
+      {...{ field }}
+      title="Profile Picture"
+      prompt="Upload an image that will make you instantly recognizable."
+    >
+      {({ register, errored, setter, current }: WizardPaneCallbackProps) => {
+        const { ref: registerRef, ...props } = register(field, {
+          validate: {
+            // lessThan10MB: (files: any) => files[0]?.size < 30000 || 'Max 30kb',
+            // acceptedFormats: (files: any) =>
+            //   ['image/jpeg', 'image/png', 'image/gif'].includes(
+            //     files[0]?.type,
+            //   ) || 'Only PNG, JPEG e GIF',
+          },
+        });
+        console.log(
+          'register props: ',
+          JSON.stringify(props),
+          `current: ${current}`,
+        );
+
+        const handleUpload = async (e: any) => {
+          console.log('uploading image', e);
+          const imageFile = e.target.files?.[0];
+          if (!imageFile) return;
+
+          setFile(imageFile);
+          const reader = new FileReader();
+          reader.addEventListener('load', () => {
+            setImageURL(reader.result as string);
+          });
+          reader.addEventListener('error', ({ target }) => {
+            const { error } = target ?? {};
+            toast({
+              title: 'Image Loading Error',
+              description: `Loading Images Error: “${error?.message}”`,
+              status: 'error',
+              isClosable: true,
+              duration: 10000,
+            });
+          });
+
+          // upload to IPFS
+          setImage({
+            preview: URL.createObjectURL(imageFile),
+            raw: imageFile,
+          });
+          console.log('setImage triggered image', e);
+          // create ref to the file
+          // const file = e.target.files[0];
+          setter(imageFile);
+          // question: Do we need to upload to web3.storage here or
+          // do we upload to web3.storage when next button is clicked?
+          console.log('end of handleUpload');
+        };
+        const handleClick = () => {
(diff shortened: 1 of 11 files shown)


packages/web/components/Setup/SetupProfilePicture.tsx
```diff
@@ -0,0 +1,120 @@
...
+                objectFit="cover"
+                borderRadius="full"
+                src={preview || current || OctoAvatar.src}
+              />
+            </Center>
+            <Input
+              type="file"
+              onChange={handleUpload}
+              name="profileImageURL"
+              accept="image/png, image/jpeg, image/gif"
```
dysbulic, line 80:
```suggestion
              accept="image/*"
```

This excludes, SVGs, WEBPs, AVIFs, and other valid image types.

There's a small chance that someone will try and enter BMP or something bizarre, but the chance is small.

packages/web/components/Setup/SetupProfilePicture.tsx
```diff
@@ -0,0 +1,120 @@
...
+              px={8}
+              letterSpacing="0.1em"
+              size="md"
+              fontSize="sm"
+              bg="purple.400"
+              color="white"
+              mt={{ base: 2, md: 6 }}
+              onClick={handleClick}
+            >
+              Upload image
```
dysbulic, line 99:
```suggestion
              Upload Image
```

packages/web/components/Setup/SetupProfilePicture.tsx
```diff
@@ -0,0 +1,120 @@
...
+                  mt={{ base: 2 }}
+                  color="#EB5757"
+                  variant="ghost"
+                  size="md"
+                  fontSize="sm"
+                  onClick={handleRemove}
+                >
+                  Remove Image
+                </Button>
+              ) : null}
```
dysbulic, line 113:
```suggestion
              {current || preview && (
                <Button
                  mt={2}
                  color="#EB5757"
                  variant="ghost"
                  size="md"
                  fontSize="sm"
                  onClick={handleRemove}
                >
                  Remove Image
                </Button>
              )}
```

packages/web/components/Setup/WizardPane.tsx
```diff
@@ -108,7 +116,41 @@ export const WizardPane = <T,>({
...
+                width: values.profileImageURL.width,
+                height: values.profileImageURL.height,
+              },
+            } as ImageSources;
+            // eslint-disable-next-line no-param-reassign
+            delete values.profileImageURL;
+            await onSave({ values, images, setStatus });
+          } else {
+            await onSave({ values, images, setStatus });
+          }
```
dysbulic, line 153: In general if you're executing the same code at the end of both an `if` & `else`, I would think you could just execute it outside the conditional.

```suggestion
          }
          await onSave({ values, images, setStatus });
```

packages/web/components/Setup/WizardPane.tsx
```diff
@@ -108,7 +116,41 @@ export const WizardPane = <T,>({
           });
         } else if (onSave) {
           setStatus('Saving…');
-          await onSave({ values, setStatus });
+          const images: HasuraImageSourcedProps = {};
+          // handle image upload of profileImageURL to web3.storage
+          if (values.profileImageURL) {
+            const formData = new FormData();
+
+            // 'profile' is the key for ceramic, equiveleant to profileImageURL in hasura
```
dysbulic, line 124:
```suggestion
            // 'profile' is the key for ceramic, equivalent to profileImageURL in Hasura
```

packages/web/components/Setup/SetupProfilePicture.tsx
```diff
@@ -0,0 +1,120 @@
...
+        };
+        const handleClick = () => {
+          inputRef.current?.click();
+          inputRef.current?.focus();
+        };
+        const handleRemove = () => {
+          setImage(null);
+          setter('');
+          setPreview(null);
+        };
```
dysbulic, line 63: If an image is removed and then saved, the edit is discarded, right? Specifically, the image is not removed from Ceramic?

The current UI is a little misleading in that it looks like you can remove an image and save the lack of an image to Ceramic.

The ideal would be to have the removed image persist to Ceramic. I don't know how the schema constraints come into play on this operation.

Barring that, maybe the "Remove" button should reset the image back to its original state. So, a blank image if it's never been set, but the image from the user's profile if it already has a value so it's more obvious you can't remove the image entirely.

This would only be if removing the image can't be persisted to Ceramic.

dysbulic (commented):
My main question has to do with saving a removed image. It doesn't look like the saving code handles that.

packages/web/components/Setup/SetupProfilePicture.tsx
kon-rad: I tried this, but doesn't work as you expect, it will return the preview, because preview is truthy, and the second part of && statement doesn't get reached.

packages/web/components/Setup/SetupProfilePicture.tsx
kon-rad: Yes, good catch, setting it as null, and passing null to ceramic works.

packages/web/components/Setup/SetupProfilePicture.tsx
dysbulic: You probably just need parens, but it's not worth revisiting.

kon-rad: > 

Yes, probably, and my auto formatter was adding parens automatically. Agreed, not worth revisiting.

dysbulic:
I'm good with this and my testing didn't break anything. One last change I was hoping you'd make is when you click on the profile image in the step you added, could you have that open the file selection dialog?

It should just be creating a `useRef` ref to the file input and calling `ref.current.click()` when someone clicks the image. My impulse, particularly when surfing & not really paying attention is to click on the thing I want to affect — the image in this scenario.

dysbulic (approved):
I had one last ask for handling clicks on the image, but otherwise I'm satisfied.

kon-rad:
> ne last ask for

copy. I'm on it.

PR #1533, opened 2023-02-18
Remove "MetaGame" from other parties' pages within our site

### LINKED ISSUE #1526: Changing page titles (removing MetaGame prefix)
It would be good to change the default page labeling way if possible.

The reason:
Looks like all pages are prefixed with "MetaGame" which is kind of lame, especially in the case of guild profiles:
image

If we want projects to use their guild pages as Linktree/something to send to people to get an overview & links to their project - the pages should probably simply be titled eg. "Giveth", rather than "MetaGame Guild Profile: Giveth".

But same goes for other things, eg. paths:
MetaGame Engaged Octo's Path > Engaged Octo's Path

Can this be done at the project level so it automatically affects all sub-pages or will each have to be changed manually?
@dysbulic @alalonde

dysbulic:
## Overview

Pages like the Guild overviews include "MetaGame Guild:" at the start of the title. This PR removes that branding to use only the external party's identifier.

Closes #1526.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -10,10 +10,10 @@
   "scripts": {
     "start": "node --trace-warnings dist/index.js",
     "build": "yarn generate && tsc -b",
-    "dev": "tsc --build && concurrently \"tsc --watch --preserveWatchOutput\" \"nodemon\" \"yarn generate --watch\"",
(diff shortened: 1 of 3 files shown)


dysbulic:
Completed as part of #1552.

PR #1558, opened 2023-05-19
Use Plausible for data tracking

vidvidvid:
## Overview

Add Plausible analytics.

Screenshot 2023-05-19 at 17 14 10

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -27,5 +27,3 @@ FRONTEND_URL=http://localhost:3000
 # packages/backend
 IMGIX_TOKEN=
 # GITHUB_API_TOKEN=   # IMPORTANT! create one at https://github.com/settings/tokens
-
-THE_GRAPH_API_TOKEN=  # create one at https://thegraph.com/studio/apikeys/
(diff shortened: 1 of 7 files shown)


dysbulic:
@vidvidvid, should Plausible be deployed on the test and development instances?

Rather than wrapping `App` in `PlausibleProvider`, you could wrap `DeployedApp` and it would only affect `production`.

vidvidvid:
> @vidvidvid, should Plausible be deployed on the test and development instances?
> 
> Rather than wrapping `App` in `PlausibleProvider`, you could wrap `DeployedApp` and it would only affect `production`.

This library already makes sure that plausible activates only in production.

dysbulic:
> This library already makes sure that plausible activates only in production.

@vidvidvid, when `next build` is run, as it is for the test instances, `NODE_ENV` is set to `production` even though it's not really the production server. There's no way to change this.

Is `NODE_ENV` what Plausible is using to determine the environment?

vidvidvid:
@dysbulic i did as you suggested

dysbulic (approved):
Seems reasonable to me.

You should include a link to these statistics somewhere. If nothing else, start a new document in `guides/`.

PR #1584, opened 2023-07-24
Upgrade to TypeScript v5 and fix resultant issues

dysbulic:
## Overview

After some difficulty getting @luxumbra & my dependencies to sync to the same version *(including using `yarn --frozen-lockfile`)*, I decided to update to the most recent version of TypeScript *(v4.9.4 → v5.1.2)*.

`react-hook-form` had an error when used with the new version & needed to be upgraded as well. Upgrading it triggered a bunch of TS errors, but they were all of the same basic format & not too difficult to resolve.

What was a pain was the dashboard components where to get their typing right *(in part because all are forwarded references)* was a huge pain, but after several hours it seems to be compiling without complaint.

I also added the `typecheck` task to the `precommit` script to avoid the issue where someone is able to commit with TS errors, but then can't push.

## Implementation

One of the problems I was having is `forwardRef<BaseType, Props>()` takes two generic arguments, and to work as a `ref` for a Chakra `Flex`, the first of those needs to be `HTMLDivElement`.

`HTMLDivElement` isn't a `ReactElement`, but the layout calculations for the boxes on the dashboard expect to receive `ReactElement`s. So, I coerced them to `JSX.Element`s by surrounding them with a `Fragment`. At first glance this seems to still run correctly, but I want to test more.


dysbulic:
And looking at the test instance, my conversion from `HTMLDivElement` to `JSX.Element` is f'ing with the height calculations. More work is required, but I can't switch this to a Draft PR as best I can tell…

PR #1591, opened 2023-08-10
Fix guild membership sync

### LINKED ISSUE #1590: Players are not getting added to the MetaFam guild
### What happened?

When someone is given the Player role in Discord, they should be added to the MataFam guild in Hasura by the discord bot. This is not happening.

@Seroxdesign was made a Player a while ago, and should be in the MetaFam guild, but isn't.

Check @Sero's profile & note the lack of MetaFam in the guilds.

Apparently the Quest Chains guild is failing for lack of membership roles & MetaFam isn't getting processed.

alalonde:
## Overview

Resolves #1590

- [ ] Fix the actual issues in the guilds that the sync is complaining about as well

## Implementation

Changed sync logic to not short-circuit if there are errors

## Assets

New trigger response e.g.
```
{
  "data": {
      "message": "Error syncing Discord guild memberships: Failed to sync 2 guilds:\nError: Quest Chains: Guild Quest Chains has no metadata.\nError: CRΞ8R DAO: Guild CRΞ8R DAO has no metadata.\n\nSuccessfully synced 1 guilds:\nMetaFam"
    },
    "type": "client_error",
    "version": "1"
}
```

### DIFF
--- a/packages/backend/src/handlers/actions/guild/sync.ts
+++ b/packages/backend/src/handlers/actions/guild/sync.ts
@@ -47,18 +47,23 @@ export const syncAllGuildDiscordMembers = async (
     );
 
     const errors = responses.filter((r) => r.status === 'rejected');
-    if (errors.length > 0) {
-      throw new Error(
-        // eslint-disable-next-line prefer-template
-        `Failed to sync ${errors.length} guilds:\n` +
-          errors.map((e) => (e as PromiseRejectedResult).reason).join('\n'),
-      );
-    }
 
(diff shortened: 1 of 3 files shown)


Seroxdesign (approved):
I'm assuming it was failing because you weren't processing succeses if even 1 error exists. Although not sure. LGTM though. Gotta test in prod

alalonde:
Looking at the code again more closely, I'm not sure this actually fixes the issue, but at least there will be better logging...

PR #1592, opened 2023-08-11
Made a couple small tweaks to the ComposeDB profile model

### LINKED ISSUE #1589: Take ComposeDB from testing to production

The ComposeDB branch is going to undergo further testing is staging so that we can avoid having constant rebasing as other branches make it into `develop`.

There are a few outstanding changes to the model schema that need to happen before going to production.

### Why is this needed?

’Cause I say so. 😸

alalonde:
## Overview

Closes #1589

Once this gets merged, I'll need to deploy the model to our production node as well


Seroxdesign (approved):
Again not sure what this does, but LGTM

packages/utils/schema/user-profile.graphql
```diff
@@ -7,7 +7,7 @@ type Profile
   avatar: ImageMetadata
   backgroundImage: ImageMetadata
   description: String @string(maxLength: 42000)
-  emoji: String @string(maxLength: 2)
+  emoji: String @string(maxLength: 3)
```
dysbulic, line 10: OpenMoji names their SVGs of emojis based on the characters. There's a `F9D1-1F3FB-200D-2764-FE0F-200D-1F48B-200D-1F9D1-1F3FF.svg`, *(a light skinned man loves a dark skinned woman)*, meaning there's at least series of ten.

There's no indication the Unicode people are going to tire of coming up with these strings, so why not do something like 20?

packages/utils/schema/user-profile.graphql
```diff
@@ -31,6 +34,7 @@ type ImageMetadata {
   # Dimensions for raster images
   width: Int @int(min: 0)
   height: Int @int(min: 0)
+
   # For vector images
-  aspectRatio: Float @float(min: 0, max: 1)
+  aspectRatio: Float @float(min: 0, max: 2)
```
dysbulic, line 39: It seems like the aspect ratio can be of arbitrary maximum size. Discord banners are 800:320 which is 2.5.

dysbulic (changes requested):
These still don't look quite right to me. Also, can someone be colorless on the five-color disposition, so the `minLength` would be zero?

*(Sorry for taking so long. Clearing headspace to review properly has been a challenge.)*

alalonde:
> These still don't look quite right to me. Also, can someone be colorless on the five-color disposition, so the `minLength` would be zero?

I removed the `minLength` requirement

dysbulic (approved):
LGTM.

PR #1596, opened 2023-08-25
fix my math on calculating total earned

Seroxdesign:
## Overview

- Fixes some math issues while calculating total earned
- Adds a few more stable coins to calculations for more accurate earnings
- Adds SEEDs to displayed total earned, for MG players

Closes #

## Implementation

**Side effects**

## Assets

### DIFF
--- a/packages/web/components/Player/Section/PlayerDework.tsx
+++ b/packages/web/components/Player/Section/PlayerDework.tsx
@@ -215,8 +215,9 @@ const DeworkProfile: React.FC<{
             <DeworkSectionWrapper>
               <DeworkSectionHeading text="Earnings" />
               <Text fontSize="2xl">
-                $
-                {(processedData.totalEarnedInUSDC / (10 * 10 ** 18)).toFixed(2)}
+                {processedData.totalEarnedInUSDC.toFixed(2)} USD
+                <br />
+                {processedData.totalSEEDsEarned.toFixed(2)} SEED
               </Text>
             </DeworkSectionWrapper>
             <DeworkSectionWrapper>
--- a/packages/web/components/Player/Section/PlayerDework.tsx
+++ b/packages/web/components/Player/Section/PlayerDework.tsx
@@ -217,7 +217,9 @@ const DeworkProfile: React.FC<{
               <Text fontSize="2xl">
                 {processedData.totalEarnedInUSDC.toFixed(2)} USD
                 <br />
-                {processedData.totalSEEDsEarned.toFixed(2)} SEED
+                {Number(processedData.totalSEEDsEarned) > 0 && (
+                  <>{processedData.totalSEEDsEarned.toFixed(2)} SEED </>
+                )}
               </Text>
             </DeworkSectionWrapper>
             <DeworkSectionWrapper>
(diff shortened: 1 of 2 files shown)


packages/web/components/Player/Section/PlayerDework.tsx
```diff
@@ -215,8 +215,9 @@ const DeworkProfile: React.FC<{
             <DeworkSectionWrapper>
               <DeworkSectionHeading text="Earnings" />
               <Text fontSize="2xl">
-                $
-                {(processedData.totalEarnedInUSDC / (10 * 10 ** 18)).toFixed(2)}
+                {processedData.totalEarnedInUSDC.toFixed(2)} USD
+                <br />
+                {processedData.totalSEEDsEarned.toFixed(2)} SEED
```
alalonde, line 220: Should we just hide each of these if they are zero ?

dysbulic: At least the SEED, since theoretically we have users from other guilds in the system as well.

packages/web/utils/dework.ts
```diff
@@ -61,9 +77,23 @@ export const processDeworkData = (data: DeworkData) => {
...
+        usdcTokens6Decimals.includes(token.address?.toLowerCase() || 'empty') &&
+        amount
+      ) {
+        totalEarnedInUSDC += +amount / 10 ** 6;
+      }
+      if (
+        usdTokens18Decimals.includes(token.address?.toLowerCase() || 'empty') &&
+        amount
+      ) {
+        totalEarnedInUSDC += +amount / 10 ** 18;
```
dysbulic, line 90: This actually works? It seems like `amount` could overflow an `int` and then get divided by 10¹⁸.

Do we need to be using `ethers.toWei()`?

packages/web/utils/dework.ts
```diff
@@ -61,9 +77,23 @@ export const processDeworkData = (data: DeworkData) => {
     });
     // get usdc rewards
     element?.rewards?.forEach(({ token, amount }) => {
-      if (usdcTokens.includes(token.address || 'empty')) return;
-      if (amount) {
-        totalEarnedInUSDC += +amount;
+      if (
+        usdcTokens6Decimals.includes(token.address?.toLowerCase() || 'empty') &&
```
dysbulic, line 81: This *should* work, but it's not the most straightforward statement.

```suggestion
      !!token.address &&
       
usdcTokens6Decimals.includes(token.address.toLowerCase())
```
Says more explicitly what is being checked.

packages/web/utils/dework.ts
```diff
@@ -61,9 +77,23 @@ export const processDeworkData = (data: DeworkData) => {
...
+        amount
+      ) {
+        totalEarnedInUSDC += +amount / 10 ** 6;
```
dysbulic, line 84: I know `+var` will coerce a variable, but it's using a side effect that some people might not know whereas `Number(var)` shows it's a cast.

dysbulic (commented):
I've got some quibbles, but the logic seems generally sound.

alalonde (approved):

PR #1609, opened 2023-09-21
limit description by length and words

x5engine:
Hello

This PR fixes this

https://meta-game.notion.site/Read-more-based-on-text-lenght-instead-of-fixed-b632c488e63b4212b8e8acf999c7eaa4


dysbulic:
The request was to limit based on the number of lines rather than the number of characters.

It looks like what you did was prevent it from breaking within a word. This both wasn't what it asked for, and could cause the text to wrap to an additional line.

The correct solution is almost certainly use of the CSS `line-clamp` property.

PR #1631, opened 2023-10-16
Improvements to the XP calculations

dysbulic:
## Overview

* Streamlines the XP calculation process using token receipt
* Returns extensive information from the balance sync process
* Creates a script for resetting the XP system
* Adds a multiplier for the conversion of received tokens to XP
* Takes the chain id into consideration when searching for Safe airdrop data
* Allows for failure of the sync process and returns an appropriate error
* Fixes the inclusion of @dysbulic on the leaderboard

Eventually it would be nice to have other methods for determining token receipt than Safe airdrops. Also, there is no UI for a couple of the configuration values like the Safe address and multiplier.

## Implementation

**Describe technical (nontrivial / non-obvious) parts of your code.**

The scraping task pulls in data from the Safe's airdrops. The XP values are calculated from base figures each time, so changing something like the multiplier will work correctly.

The seasonal XP is not multliplied by the multiplier. It's the raw amount of SEED received.


packages/backend/src/handlers/actions/player/syncBalances.ts
```diff
@@ -50,86 +59,141 @@ const setBalances = async ({
...
-    Object.entries(uniqueDrops)
-      .map(([executionDate, drops]) =>
+  const added = await Promise.all(
+    Object.entries(uniqueDrops).map(async ([executedAt, drops]) => {
+      const dropsRet = await Promise.all(
         Object.entries(drops).map(async ([to, value]) => {
-          await client.AddBalance({
-            amount: value,
-            executedAt: new Date(executionDate),
+          const dat = {
```
Seroxdesign, line 66: dat? what dat?

This could be name better

packages/backend/src/handlers/actions/player/syncBalances.ts
```diff
@@ -50,86 +59,141 @@ const setBalances = async ({
...
+                    accumulated: balance,
+                    calculated,
+                    seasonal: seasonalBalance,
+                  },
+                };
+              },
+            ),
+          );
+          return {
+            ...balRet,
```
Seroxdesign, line 166: can we use full words for return, balanceReturned or something. I have context and still confusing

Seroxdesign (changes requested):

Seroxdesign:
image

PR #1634, opened 2023-10-24
Sero/bug playbook submit proof

Seroxdesign:
## Overview

Closes #

## Implementation

**Side effects**

## Assets


dysbulic:
@vidvidvid or @dan13ram, when submitting a quest proof we were getting a “no quest with that id" error.

It seems as though there's a separate contract for each quest chain and each chain contract contains a single chain, so we just fixed the quest id at `0`.

This is safe?

dysbulic (approved):
Assuming the id change is correct, LGTM.

vidvidvid:
ohhh no, questId needs to be the same as the id of the relevant quest in the quest chain. so if a quest chain has 5 quests and the user will want to submit proof for quest nr. 1, they will have to use the index 0, etc...

vidvidvid:
you will also need to pay attention to whether there are any in-active quests, because they have incremented the index, so if a quest chain has 4 quests and quests 2 and 3 are inactive, only quests 1 and 4 will be active, but the relevant indices will be 0 and 3 as opposed to 0 and 1.

dysbulic:
> `questId` needs to be the same as the id of the relevant quest in the quest chain

We checked a couple playbooks & they all had `questCount`s of `1`. Do you know if all playbooks are a single quest? @peth-yursick?

@Seroxdesign, we can dig into this some more and see if we can figure out how the quest id is getting corrupted.

*(@vidvidvid, before we changed it, the value for `questId` getting passed into `submitProofs` is something like "`0x1923847891-0x0`" where `0x0` is the value we actually want.)*

PR #1636, opened 2023-10-31
adding env vars for the calendar stuff which will hopefully fix test.

### LINKED ISSUE #932: Fetch google calendar data and rework the dashboard calendar design

The dashboard includes a calendar section, which uses the default embeddable google calendar, that looks terrible to say the least.
In order to implement our own design, we would need to fetch the data either with some sort of api or by getting and parsing the ics format of our calendar.

### Option 1: Service account

Google services require to send authentication headers with the request, even though the calendar is public so in order to bypass every single user having to authenticate, we could make use of a service account.
Here is how to set it up: https://developers.google.com/identity/protocols/oauth2/service-account
@polimyl already set one up, so you can ping me or him and i can send you the json with all the data.

### Option 2: Fetch ics

[polimyl  suggested we could fetch the ics](https://github.com/MetaFam/TheGame/pull/902#issuecomment-970062812) (`https://calendar.google.com/calendar/ical/nih59ktgafmm64ed4qk6ue8vv4@group.calendar.google.com/public/basic.ics`) on the backend in order to work around the restrictive cors policy.

@dysbulic [suggested this solution would be similar to the implementation of the IDX cache.](https://discord.com/channels/629411177947987986/629412800346849302/910368806017306655)

### Reference for the pretty design of the calendar:
https://github.com/MetaFam/TheGame/blob/d4f0d07fba8da857f898aebe76242484be784bf2/packages/web/components/Dashboard/Calendar.tsx
In this solution I created a google sheet and a script that gets the data from the MetaGame calendar and fills out the fields in the sheet. Then I published the sheet as csv, fetched it in the web app and parsed the data. The data was being updated hourly. But this felt pretty hacky and since there are apparently better solutions available we should probably choose one of them, but I spent way too much time on trying to implement this, so I'm leaving this to someone more capable. Good luck!

### Why is this needed?

Because of this:
Screenshot 2021-11-21 at 16 54 45

luxumbra:
## Overview
Introduces the required env vars for the calendar update in #932 and will hopefully fix the broken build

Updates to the github workflows and renaming where needed in the app. I have also added the secrets in the repo settings in GH.

Closes #932 

## Implementation

**Side effects**
Should fix the test site.


packages/web/pages/_app.tsx
```diff
@@ -18,15 +18,20 @@ import PlausibleProvider from 'next-plausible';
 import { WithUrqlProps } from 'next-urql';
 import React from 'react';
 
-const { userbackToken, honeybadgerAPIKey } = CONFIG;
+const { userbackToken, honeybadgerAPIKey, gaId } = CONFIG;
 
-const Analytics: React.FC = () =>
-  !CONFIG.gaId ? null : (
+const Analytics: React.FC = () => {
+  // console.log('CONFIG', gaId);
```
Seroxdesign, line 24: comment

packages/web/pages/api/events.ts
```diff
@@ -1,29 +1,97 @@
...
   if (req.method === 'GET') {
-    const calId = options.GOOGLE_CALENDAR_ID || '';
+    const calId = options.CALENDAR_ID || '';
     const calendarId = `${calId}@group.calendar.google.com`;
 
+    // strip out the +++cover+++ from the description
+
+    const groupEventsByDay = (items: GoogleCalEventType[]) => {
+      const groupedEvents = items.reduce((acc, event) => {
+        // console.log('event', {acc, event});
```
Seroxdesign, line 49: comment

Seroxdesign (approved):

PR #1646, opened 2023-11-30
WIP feat: background update

luxumbra:
- added r3f threejs and implemented a starfield
- need to add/fix the scroll parallax (see the metafest.wtf site) (is it needed?)
- the bg images seem to be low quality despite me exporting high res & multiple versions
- next/image used in preference to BoxedNextImage (uses legacy next/image) or a plain bg image so we get benefits of optimisation, etc

## Overview

The work adds a new bg image to the app with a React Three Fiber parallax (on mousemove) starfield overlay with effects toggle and fallback bg image if the user prefers reduced motion or chooses to toggle the effect. It is quite a basic set up in terms of threejs but could be enhanced in a few ways.

I've just noticed a bug with the mousemove parallax having jerky movement. Will look into that. 

I don't think there is an issue number.
Closes #

- [ ] Add some lighting at the bottom to replace the blurred circles
- [ ] In the MF version, there is a parallax effect on scroll where the camera shifts with the scroll position but it doesnt work in this version currently due to the way we handle the scrolling of content on pages. I can get it to work but is more involved and is maybe not wanted at this point.

## Implementation

**Side effects**
Three.js and it's accoutrements do add some heft to the bundle so I'll change the import to a dynamic import and add Suspense. 

I noticed that the toggle is visible on the landing page so i need to add some logic to hide it.

## Assets
image
image


dysbulic:
You ought to [add the `sharp` package](https://nextjs.org/docs/messages/install-sharp) so it'll optimize your images.

Issue #1673, opened 2023-12-14
Dashboard: coingecko API being rate-limited


alalonde:
### What happened?

We're getting frequent 429s when fetching SEED price

### What did you expect to happen?

200s

### How can we reproduce the problem (as minimally as possible)?

Go to your dashboard on a PR deployment or test or production 

### Is there anything else we need to know?

Some potential solutions:
- Pay coingecko for a higher rate limit
- switch token price API providers
- Proxy the query through Hasura and cache it daily

Seroxdesign:
Proxying the query is a good idea. Coingecko costs 100$ a month and all other API providers will eventually have us run into the same issues.

PR #1679, opened 2023-12-21
Only fetch data from coingecko once

alalonde:
Addresses #1673


Udit-takkar (approved):
LGTM. Tested and works fine

PR #1697, opened 2024-01-10
Add pinned quest chains dashboard card

vidvidvid:
Screenshot 2024-01-10 at 17 13 46

Screenshot 2024-01-10 at 17 14 03


Seroxdesign (commented):
Can we get the questchain link rows styles similarly to the player rows in leaderboard on bottom right?

Also can we have the pinned block to display  by default instead of having the player save the layout

vidvidvid:
adapted the style & pinned the block by default.

Screenshot 2024-01-15 at 14 45 37
Screenshot 2024-01-15 at 14 44 58

Seroxdesign (approved):

PR #1705, opened 2024-01-18
Seeds redesign

vidvidvid:
https://www.figma.com/file/n36ggiRBQEGzRUbGxWfkMX/Meta-Game-(Copy)?node-id=10747%3A9896&mode=dev

https://github.com/MetaFam/TheGame/assets/35112939/536170ca-a6e9-4611-b75c-33788ac6f6c5


vidvidvid:
- [ ] now that image fits by itself, this massive block of dark purple should be removed
- [ ] now that image is shorter, all of the content could be moved up a bit more so its more visible
- [ ] amount of droplets still doesnt change when you click the $ amount buttons
- [ ] would make the first Q also hidden by default
- [ ] "planting" & "trading" images are still just images > should be clickable with the link embedded
- [ ] the "selected tab" line shouldn't be white but purple (same purple as the underline in the nav bar #A48DF3)
- [ ] there's an apostrophe at the end of the "Seeds are lifeblood..." sentence
- [ ] 7b. super minor, but could you pls make it so that "how we" also breaks into new line along with "curate & how we reward each other", and change "&" to "and"
- [ ] pls remove the "Page" from the page title

vidvidvid:
@peth-yursick fixed!
Screenshot 2024-01-24 at 19 49 02

peth-yursick:
looks like a new issue appeared on test - for some reason there are now margins on each side of the image :S
image

peth-yursick:
maybe see if its an easy fix, otherwise we can disregard for now so as to not waste time

vidvidvid:
@peth-yursick thanks for spotting it. please open an issue just to keep track - it's an easy fix.

PR #1761, opened 2024-07-09
Switch Quest Chains from `ethers` to Viem for Contract Access

dysbulic:
## Overview

Replaces the Quest Chains SDK `contract` code generated by [TypeChain](https://github.com/dethcrypto/TypeChain/) with Viem's `getContract`.

## Follow-Up Improvement Ideas

A Viem client masquerading as an `ethers` provider is still used in several places. Fully transitioning would likely not be too difficult.

This is part of the process of transitioning `TheGame` back to `ethers` `v6` so we can use the Ethereum Attestation Service.

Also, this is part of migrating the wallet management infrastructure to WalletConnect’s ([Wallet](https://walletconnect.com/walletkit)|[App](https://walletconnect.com/appkit))Kit.

## Implementation

Updates bunches of libraries including the GraphQL generation, ESLint, Ceramic, & `yarn` from 1.19 to 4.3. The `yarn` update required quite a bit of finagling with the building & deployment processes.

It switches local paths from using the `paths` directive of TypeScript to the `imports` section of `package.json`.

I'm creating a Viem `WalletClient` from `window.ethereum` without any additional sign-in. That may be problematic, but all that should get overhauled when integrating [AppKit](https://walletconnect.com/appkit/).


Innkeeping (commented):
so much space saved with webp.  hope viem helps on questchains

packages/discord-bot/src/config.ts
Innkeeping, line 1: Do you need the comma before the parentheses on line 54 for Hasura Admin?

Innkeeping (commented):
Seems good.  Mentioned maybe an extra comma in another comment.

packages/discord-bot/src/config.ts
dysbulic: Lines will frequently include a trailing comma so that later when a line is added after it, you don't have to add a comma in that edit.

Otherwise, `git` will show both lines as changed even though the one that just got a comma isn't meaningfully different.

JesuscoinsIII (commented):
updated

Innkeeping (commented):
👍

docker/backend/Dockerfile
Innkeeping, line 1: @dysbulic Does line 41 change need an actual url link also?

Innkeeping: Or it's fine because line 44 refers to the env?  I'm still getting backend build fails on the test build for dAcademy PR. @dysbulic

PR #1784, opened 2024-09-19
Update questChains.ts - fix url of players-path

Innkeeping:
- changed the url for 'players path' from the mistake of 'rogues-path'


dysbulic (approved):
Seems reasonable…

PR #1795, opened 2026-04-11
fix: limit guild membership display to 3 on player cards

### LINKED ISSUE #1629: Guilds display on player cards not working (properly 😂)
Just went back to take a screenshot for this issue & looks like they _do_ work - just take ~2-3 min to load..
Should probably be cached in some way?

The issue I was originally going to display is that the cards display all guild memberships instead of displaying 3 & "+X".
In some cases, this means it breaks into the second line & looks ugly.
image

## 2 issues
- Guild membership display on cards takes a few minutes to load
- It displays all memberships instead of only 2 or 3 with a number indicator for how many more there are

Ai-chan-0411:
Fixes #1629

## Changes
- Reduced default guild membership display limit from 12 to 3 on player cards
- Updated dynamic resize calculation to cap at 3 memberships (was minimum 8)
- Excess memberships now show "+X" indicator as designed

This prevents guild memberships from breaking into multiple rows and keeps player cards visually consistent.


Ai-chan-0411:
Closing due to no review activity. Thank you for the opportunity!

